PR #80, opened 2026-05-14
Add const widget constructors

mzl2233:
Adds const constructors for the static widget instances called out in the issue, including the map grid painter and contact rows, so Flutter can canonicalize those parts of the widget tree during rebuilds. The change is intentionally small and limited to the affected widget files.

### DIFF
--- a/supabase/migrations/20240101000000_rls.sql
+++ b/supabase/migrations/20240101000000_rls.sql
@@ -238,11 +238,47 @@ DROP POLICY IF EXISTS "Service role full access on trips" ON trips;
 CREATE POLICY "Service role full access on trips"
   ON trips FOR ALL TO service_role USING (true) WITH CHECK (true);
 
+-- Drivers may only READ their own trips. Writes (create/complete/cancel) are
+-- deliberately NOT exposed to authenticated users: they must go through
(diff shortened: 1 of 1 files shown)


KanishJebaMathewM:
Hey @mzl2233, thanks for the contribution.

I reviewed all 5 files from issue #78. You have made changes in `lib/screens/about_screen.dart` and `lib/widgets/map_placeholder.dart`, but there are still some remaining items in those files as well.

In `about_screen.dart`, the `const` is applied to the list at line 141 but the parent `Row` at line 139 is still missing `const`. Could you add `const` to the `Row` widget itself so the propagation is complete.

In `map_placeholder.dart`, it is mostly resolved. Lines 53 and 58 are fine as `_MapPin` uses runtime params so they correctly cannot be `const`.

Additionally, these 3 files from issue #78 have not been touched yet:

* `lib/widgets/common_widgets.dart`
* `lib/screens/saved_addresses_screen.dart`
* `lib/screens/home_screen.dart`

Could you go through all the above and apply the same `const` additions wherever parameters are static or constant values.

Once all of these are addressed this PR is good to merge. Let me know if you have any questions.

KanishJebaMathewM:
Please mention the issue you are working on next time, which would be helpful for me to close them

PR #205, opened 2026-05-27
fix: multiple bug fixes

Xenon010101:
## Summary of Fixes

- **Fixed:** wallet_pending overwritten instead of accumulated in withdrawal endpoint (driverRoutes.js:169)
- **Fixed:** Missing error return when withdrawal transaction insert fails (driverRoutes.js:191-197)
- **Fixed:** Null-safety for customer_name in load_offers insert (orderRoutes.js:89)
- **Fixed:** toll_estimate undefined causing NaN in net_profit calculation (orderRoutes.js:97-99)
- **Fixed:** Wrong app name \Truckify\ instead of \FreightFair\ in customer profile menu (profile_screen.dart:159)

### DIFF
--- a/apps/customer/lib/screens/profile_screen.dart
+++ b/apps/customer/lib/screens/profile_screen.dart
@@ -156,7 +156,7 @@ class ProfileScreen extends StatelessWidget {
                   ),
                   _MenuItem(
                     icon: Icons.info_outline_rounded,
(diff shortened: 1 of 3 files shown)


KanishJebaMathewM:
Thank you for your contributions @Xenon010101

PR #210, opened 2026-05-27
Fix null-safe FAQ access in help support screen #184

pranayukey200:
I changed apps/customer/lib/screens/help_support_screen.dart so the FAQ entries use null-safe defaults instead of force-unwrapping map values, which removes the runtime-crash risk if a key is missing


pranayukey200:
Hello @KanishJebaMathewM , could you please add the GSSOC labels to this PR when convenient?

Requested labels:

level:beginner
quality:clean
type:bug

:-/

KanishJebaMathewM:
Go ahead @pranayukey200 !!!!!

PR #281, opened 2026-05-31
fix: add raw backspace handler for driver OTP fields

### LINKED ISSUE #279: Implement Raw Backspace Handler for Driver OTP Input Fields to Match Customer App
## Description

In the Driver application, navigating between OTP input fields using the **Backspace** key does not work smoothly when the current field is already empty. Users are forced to manually select the previous field to make corrections.

The Customer application already handles this scenario by using a raw keyboard event listener (`onKeyEvent`) to detect backspace presses on empty fields and automatically shift focus to the previous input field. This behavior should be implemented consistently across both applications.

## Current Behavior (Driver App)

When entering an OTP in the Driver application, pressing the **Backspace** key inside a text field that is already empty does not move the cursor to the previous input box.

This happens because the current implementation relies only on the `onChanged` callback, which is not triggered when attempting to delete text from an already empty field.

## Expected Behavior (Driver App)

Pressing **Backspace** inside an empty OTP digit field should:

- Automatically move focus to the previous OTP field.
- Clear any existing value in the previous field.
- Provide the same smooth and intuitive OTP input experience as the Customer application.
- Ensure consistent OTP navigation behavior across both Customer and Driver applications.

CoderMS07:
## Description

Implemented raw backspace handling for Driver OTP input fields to match the behavior already available in the Customer application.

## Changes Made

- Added onKeyEvent handling for OTP focus nodes
- Detect Backspace key presses on empty OTP fields
- Move focus to the previous field when Backspace is pressed on an empty field
- Clear the previous field value when navigating back
- Preserved existing OTP input behavior and UI

## Benefits

- Consistent OTP navigation across Driver and Customer applications
- Improved user experience when correcting OTP input
- Eliminates the need to manually select previous fields

Closes #279

### DIFF
--- a/apps/driver/lib/screens/otp_screen.dart
+++ b/apps/driver/lib/screens/otp_screen.dart
@@ -32,7 +32,9 @@ class _OtpScreenState extends State<OtpScreen> {
   }
 
   Future<void> _verifyOtp() async {
-    final code = _controllers.map((controller) => controller.text).join();
+    final code = _controllers
+      .map((c) => c.text.replaceAll('\u200B', ''))
+      .join();
     if (code != '1234') {
       ScaffoldMessenger.of(context).showSnackBar(
         const SnackBar(content: Text('Enter mock OTP 1234 to continue')),
--- a/apps/driver/lib/widgets/common_widgets.dart
+++ b/apps/driver/lib/widgets/common_widgets.dart
@@ -514,7 +514,7 @@ class StatusDot extends StatelessWidget {
   }
 }
 
-class OtpInputRow extends StatelessWidget {
+class OtpInputRow extends StatefulWidget {
   const OtpInputRow(
       {super.key, required this.controllers, required this.focusNodes});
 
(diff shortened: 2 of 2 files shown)


CoderMS07:
Hi @KanishJebaMathewM, Could you please add the appropriate GSSoC label if everything looks good?

KanishJebaMathewM:
Thanks for working on this!

While this approach works well with physical hardware keyboards (such as desktop builds or emulators with keyboard passthrough enabled), it won't trigger on actual iOS and Android devices using on-screen software keyboards.

On mobile devices, soft keyboards do not dispatch raw key events to the application. Additionally, because the input field is already empty, pressing backspace doesn't change the text content, meaning the operating system's keyboard registers no change and sends no event to the app.

To resolve this for mobile users, a common strategy is to populate each input field with an invisible placeholder character (like a zero-width space). That way, when the backspace key is pressed, it deletes the invisible character. This registers as an actual text change, allowing the app to detect the action on any device and shift the focus backwards.

KanishJebaMathewM:
Try to resolve the conflicts too @CoderMS07

CoderMS07:
Hi @KanishJebaMathewM, Thanks for the review and suggestions!

Implemented the suggested zero-width space approach to support backspace navigation on iOS and Android soft keyboards. This allows backspace presses on empty OTP fields to be detected reliably and moves the focus to the previous field while preserving the existing OTP input behavior. Also resolved the merge conflicts.

Tested the updated implementation with IOS and Android devices using on-screen software keyboards.

KanishJebaMathewM:
Thanks for implementing the zero-width space approach @CoderMS07 ! I've pushed a quick update to your branch to clear the previous OTP field's digit on backspace-navigation so it behaves consistently with the Customer app. Ready to merge!

PR #291, opened 2026-05-31
feat: add rate limiting middleware to protect API endpoints [ GSSOC'26 ]

### LINKED ISSUE #288: feat: add rate limiting to API endpoints [ GSSOC'26 ]
Problem — all API endpoints are completely open with no request throttling. A single client can flood the server with unlimited requests, causing degraded performance or downtime for other users.
Solution — add express-rate-limit middleware with a general 100 requests per 15 minutes limit on all /api/ routes, and a stricter 30 requests per minute on /api/health to prevent health check abuse.
Files changed — server.js, package.json

rudra3007-pro:
Summary — protects all API routes from abuse and accidental flooding with zero-config rate limiting.
Changes:

General limiter: 100 requests per 15 minutes per IP on all /api/ routes
Strict limiter: 30 requests per minute on /api/health
Returns standard RateLimit headers for client awareness
Clean JSON error response on limit exceeded

Closes #288

### DIFF
--- a/backend/api/package-lock.json
+++ b/backend/api/package-lock.json
@@ -13,6 +13,7 @@
         "cors": "^2.8.5",
         "dotenv": "^16.4.5",
         "express": "^5.2.1",
+        "express-rate-limit": "^8.5.2",
         "firebase-admin": "^12.1.1",
         "ioredis": "^5.11.0",
         "mongodb": "^6.7.0",
@@ -1217,6 +1218,24 @@
         "url": "https://opencollective.com/express"
       }
     },
+    "node_modules/express-rate-limit": {
+      "version": "8.5.2",
+      "resolved": "https://registry.npmjs.org/express-rate-limit/-/express-rate-limit-8.5.2.tgz",
+      "integrity": "sha512-5Kb34ipNX694DH48vN9irak1Qx30nb0PLYHXfJgw4YEjiC3ZEmZJhwOp+VfiCYwFzvFTdB9QkArYS5kXa2cx2A==",
+      "license": "MIT",
+      "dependencies": {
+        "ip-address": "^10.2.0"
+      },
+      "engines": {
+        "node": ">= 16"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/express-rate-limit"
+      },
+      "peerDependencies": {
+        "express": ">= 4.11"
+      }
+    },
     "node_modules/express/node_modules/mime-db": {
       "version": "1.54.0",
       "resolved": "https://registry.npmjs.org/mime-db/-/mime-db-1.54.0.tgz",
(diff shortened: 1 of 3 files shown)


KanishJebaMathewM:
Hi @rudra3007-pro,

Thanks for submitting this! The rate limiting implementation looks like a great addition. However, there are a few structural issues that need to be addressed before we can merge this PR:

### 1. Missing Dependency
The rate limiting package used in the implementation has not been added to our dependency configuration file (`package.json`).

Without this dependency being declared, the application will fail to start on fresh installations and in deployment environments where dependencies are installed from scratch.

### 2. Health Check Polling Overlap
The general rate limiter is currently applied to the entire API path, which includes our health check endpoint.

In production, load balancers, uptime monitors, and other infrastructure components frequently poll health check endpoints. These requests can quickly consume the 100-request limit, causing the endpoint to be rate-limited and potentially leading the system to incorrectly report that the server is unavailable.

Please exclude the health check endpoint from the general rate limiting configuration.

### 3. Reverse Proxy Support
The application is not currently configured to trust reverse proxies.

When deployed behind a load balancer or cloud proxy, all incoming requests appear to originate from the proxy's IP address rather than the actual client IP. As a result, all users would share the same rate-limit quota and could be blocked from accessing the API almost immediately.

Please configure the server to trust upstream proxy headers so rate limiting can correctly identify individual clients.

Could you update the PR to address these items? Thanks!

rudra3007-pro:
@KanishJebaMathewM The changes you reviewed are succesfully implemented. Please re-review.

KanishJebaMathewM:
Hi @rudra3007-pro,

Thanks for the updates! However, there are still a few issues that prevent the server from starting up and working as intended:

### Issues Identified

1. **Missing Dependency**
   - The rate-limiting package is still missing from the project's dependency configuration file.
   - Fresh installations will fail to find the required module.

2. **Startup Crash (Duplicate Logic)**
   - The new rate-limiting block was appended without removing the previous one.
   - This creates duplicate variable declarations that cause the application to crash with a syntax error immediately upon startup.

3. **Bypass Logic Path Issue**
   - The check used to skip the health check endpoint will not work.
   - When rate-limiting middleware is mounted on a parent route prefix, the `path` property is shortened, causing the path check to fail and the health check to remain subject to the general rate limit.
   - Checking the full original request URL instead should resolve this issue.

4. **Proxy Support Verification**
   - The configuration to trust proxies is present.
   - However, it can only be properly verified once the startup and crash issues are fixed.

### Requested Changes

Could you please:

- Remove the duplicate rate-limiting logic.
- Adjust the bypass path check to use the original request URL.
- Register the rate-limiting package in the dependency configuration.

Thanks!

KanishJebaMathewM:
@rudra3007-pro Please implement the reviewed changes so that I can merge this successfully

rudra3007-pro:
@KanishJebaMathewM I have implemented the changes

rudra3007-pro:
@KanishJebaMathewM Please review this PR

KanishJebaMathewM:
Thanks for implementing the requested changes @rudra3007-pro ! The dependencies, duplicate logic, and health check bypass are all resolved and verified. Merging this PR now.

PR #382, opened 2026-06-06
(security fix):add RLS policies for authenticated users across all 24…

### LINKED ISSUE #358: [Security] Supabase RLS Misconfiguration - Missing Policies for Authenticated Users
## Description

Row-Level Security (RLS) is enabled across the database, but the current policy configuration only grants access to the `service_role`. No policies exist for authenticated users accessing the database through the Flutter applications.

### Affected File

- `docs/supabase_setup.sql`

### Problem

The existing RLS configuration enables security on tables but does not define appropriate policies for authenticated users.

Currently missing:

- SELECT policies for authenticated users
- INSERT policies for authenticated users
- UPDATE policies for authenticated users

As a result, the Flutter applications cannot properly interact with Supabase using normal user sessions.

### Impact

- Customer and driver apps may fail to load data.
- Insert and update operations may be rejected by Supabase.
- Users may receive empty datasets despite having valid records.
- Client-side database functionality becomes unusable.

### Suggested Fix

Define granular RLS policies that allow authenticated users to access only their own records.

Examples include:

- Users can view and update their own profiles.
- Customers can access only their own orders.
- Drivers can access only their own trips.
- Users can create records associated with their own account.

### Acceptance Criteria

- Authenticated users can successfully read their own data.
- Authenticated users can create and update their own records.
- Users cannot access records belonging to other users.
- Flutter applications function correctly using standard authenticated sessions.

Himanshujha7:
##Description:
1. RLS was enabled on all 24 tables but every policy used `using (true)`  with no `to` clause
2. No row-level filtering existed for authenticated users. 
3. Any authenticated client could read or write every row in every table regardless of ownership.

##fixes:

fixes: #358 

##Approach:
1. **Scopes existing service_role policies** by adding `to service_role` so they no longer accidentally apply to all roles
2. **Adds authenticated user policies** for all 24 tables so each user can only access their own records
3.  **Adds `get_profile_id()` helper function** that maps `auth.uid()` (Firebase UID) → `profiles.id` (UUID), used by all row-filtering policies

## Files changed

- `docs/supabase_setup.sql`

### DIFF
--- a/docs/supabase_setup.sql
+++ b/docs/supabase_setup.sql
@@ -50,7 +50,7 @@ create or replace function get_profile_id()
 returns uuid
 language sql stable security definer
 as $$
-  select id from profiles where firebase_uid = auth.uid()::text limit 1;
+  select id from profiles where firebase_uid = (auth.jwt() ->> 'sub') limit 1;
 $$;
 
 
@@ -710,18 +710,18 @@ create policy "Service role full access on profiles"
 create policy "Users select own profile"
   on profiles for select
   to authenticated
-  using (firebase_uid = auth.uid()::text);
+  using (firebase_uid = (auth.jwt() ->> 'sub'));
 
 create policy "Users insert own profile"
   on profiles for insert
   to authenticated
-  with check (firebase_uid = auth.uid()::text);
+  with check (firebase_uid = (auth.jwt() ->> 'sub'));
 
 create policy "Users update own profile"
   on profiles for update
   to authenticated
-  using (firebase_uid = auth.uid()::text)
-  with check (firebase_uid = auth.uid()::text);
+  using (firebase_uid = (auth.jwt() ->> 'sub'))
+  with check (firebase_uid = (auth.jwt() ->> 'sub'));
 
 
 -- 2. DRIVER DETAILS
@@ -951,6 +951,20 @@ create policy "Drivers view own trip stops"
   on trip_stops for select
   to authenticated
   using (
+    trip_display_id in (
+      select trip_display_id from trips where driver_id = get_profile_id()
+    )
+  );
+
+create policy "Drivers update own trip stops"
+  on trip_stops for update
+  to authenticated
+  using (
+    trip_display_id in (
+      select trip_display_id from trips where driver_id = get_profile_id()
+    )
+  )
+  with check (
     trip_display_id in (
       select trip_display_id from trips where driver_id = get_profile_id()
     )
--- a/docs/supabase_setup.sql
+++ b/docs/supabase_setup.sql
@@ -46,6 +46,12 @@ begin
   return new;
 end;
 $$;
+create or replace function get_profile_id()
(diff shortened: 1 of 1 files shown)


KanishJebaMathewM:
Hi @Himanshujha7,

Thank you for the detailed breakdown! This is a great catch. You are absolutely correct that because the Flutter apps currently connect as `anon` (unauthenticated) clients and use hardcoded seed IDs in development, RLS policies targeting `authenticated` users will block these requests.

Implementing the actual Firebase/Supabase sign-in/sign-up flow in the Flutter apps is a major frontend task. If you would like to implement that auth flow, you are welcome to do it in this PR, or we can handle it in a separate follow-up PR to keep things modular. 

If we handle it separately, you can test your database changes locally in development by programmatically signing in a mock user in `main.dart` on app startup (using `signInWithPassword` with development credentials) so that a valid authenticated JWT is generated.

Additionally, I ran a review on the SQL policies and noticed two database-level issues we need to resolve:

1. **`auth.uid()` Type-Casting Crash:** 
   In Supabase, `auth.uid()` expects a standard UUID format and will crash when it tries to cast alphanumeric Firebase UIDs (like our seed data `'firebase_customer_001'`). To fix this, we should fetch the `sub` claim directly as text from the JWT claims using `auth.jwt() ->> 'sub'` inside `get_profile_id()` and your `profiles` table policies.

2. **Missing `UPDATE` Policy on `trip_stops`:** 
   The driver app directly updates stop completion statuses via `TripService.markStopCompleted`. Since the current policies only define `select` permissions on `trip_stops` for authenticated users, the app will fail with a permission error. We need to add an `update` policy on `trip_stops` that allows drivers to update stops belonging to their trips.

Let me know if you would like to include the Flutter authentication flow in this PR or tackle it in a separate PR. Once these SQL issues are addressed, the database RLS policies themselves will be ready!

Himanshujha7:
Hi @KanishJebaMathewM 

Thank you for the detailed response!

Yes, I agree to the points you have mentioned, and I believe we can have the actual Firebase/Supabase sign-in/sign-up flow in the Flutter apps as a separate follow-up PR to keep scopes consistent and organised.
 
 For the SQL-related policy issues:
 1. I agree, as you mentioned, and upon verifying when the Firebase UID is cast to a corresponding UUID, it will throw an error.
 To fix this, we skip the `uuid` casting and provide similar is `profile` policies
 
 2. Adding a new policy that lets a driver update a trip_stops row only if that stop belongs to one of their own trips (checked via trips.driver_id = get_profile_id()), mirroring the same ownership check already used for the SELECT policy, will resolve the SQL issue as well.
 
 I will proceed with the changes mentioned above and related to the implementation of the auth flow. Let's open a separate issue to discuss the scope of implementation.

Himanshujha7:
Hi @KanishJebaMathewM
Kindly review and let me know if any changes are needed from my end.

PR #412, opened 2026-06-08
fix(api): rate limit WebSocket upgrades

### LINKED ISSUE #403: [Backend] Security: Connection Rate-Limiting for WebSocket Upgrades
### Context

The REST API is already protected by rate-limiting middleware configured in `index.js`. However, the WebSocket upgrade endpoint (`/ws/tracking`) currently lacks any form of rate limiting.

An attacker could repeatedly initiate WebSocket upgrade requests and create excessive concurrent handshakes, potentially exhausting available socket descriptors, memory, and server resources. This could degrade application performance or lead to denial-of-service scenarios.

To improve security and operational resilience, WebSocket upgrade requests should be rate-limited on a per-IP basis using Redis.

### Target Areas

- `backend/api/src/index.js`
- `backend/api/src/tracker.js`
- Redis integration utilities
- WebSocket upgrade handling logic

#### 1. Implement Upgrade Rate Limiting

Add rate-limiting protection to the WebSocket upgrade flow.

The rate limit should be enforced within either:

```text
index.js
```

or

```text
tracker.js
```

before the WebSocket handshake is completed.

---

#### 2. Track Requests Per IP Address

Use Redis to track upgrade attempts by client IP.

Suggested key format:

```text
ws:upgrade:${ipAddress}
```

Example:

```text
ws:upgrade:192.168.1.100
```

Store and increment connection attempts within a rolling time window.

---

#### 3. Enforce Upgrade Limits

Configure a maximum of:

```text
5 WebSocket upgrade requests per minute per IP
```

Example behavior:

```text
Attempt 1 → Allowed
Attempt 2 → Allowed
Attempt 3 → Allowed
Attempt 4 → Allowed
Attempt 5 → Allowed
Attempt 6 → Blocked
```

Contributors may implement an equivalent Redis-based approach provided the limit remains effectively enforced.

---

#### 4. Reject Excessive Requests

When an IP exceeds the limit:

- Reject the upgrade request immediately.
- Return:

```http
HTTP/1.1 429 Too Many Requests
```

- Abort the WebSocket handshake.
- Destroy the underlying socket connection.

Example:

```js
socket.write(
  'HTTP/1.1 429 Too Many Requests\r\n\r\n'
);

socket.destroy();
```

---

#### 5. Graceful Redis Failure Handling

If Redis becomes unavailable:

- Log the cache failure.
- Avoid crashing the WebSocket server.
- Fall back to a safe behavior consistent with existing application patterns.

The WebSocket upgrade path must remain stable even during Redis outages.

---

#### 6. Automated Testing

Add integration or unit tests that verify:

##### Allowed Requests

- [ ] Requests within the limit succeed

##### Blocked Requests

- [ ] Excessive upgrade attempts are rejected
- [ ] HTTP 429 is returned
- [ ] Handshake does not complete

##### Edge Cases

- [ ] Separate IPs are tracked independently
- [ ] Redis failures are handled gracefully
- [ ] No unexpected server crashes occur

---

### Expected Outcome

After implementation:

- WebSocket upgrades will be protected against connection flooding attacks.
- Resource exhaustion risks will be reduced.
- Abuse of the `/ws/tracking` endpoint will be significantly harder.
- Existing legitimate WebSocket usage will continue to function normally.

### Security Considerations

- Rate limiting should occur **before** the WebSocket connection is fully established.
- Avoid storing excessive per-IP data in Redis.
- Ensure expired counters are automatically cleaned up using TTLs.
- Follow existing security and infrastructure patterns already used throughout the backend.

### Notes

- Reuse the existing Redis client already configured in the project.
- Keep the implementation lightweight and efficient.
- Focus on preventing handshake abuse rather than limiting message traffic after connection establishment.
- Existing WebSocket functionality must remain fully compatible.

saurabhhhcodes:
## Summary
- add Redis-backed per-IP rate limiting before `/ws/tracking` WebSocket handshakes
- reject the sixth upgrade attempt in a one-minute window with HTTP 429 and close the socket
- keep upgrades available if Redis is unavailable, while logging the limiter failure
- add unit coverage for allowed, blocked, independent-IP, Redis-failure, and 429 rejection behavior

Fixes #403

## Validation
- `npm test -- --run test/unit/tracker.test.js`
- `node --check src/sockets/tracker.js`
- `git diff --check`

### DIFF
--- a/backend/api/src/sockets/tracker.js
+++ b/backend/api/src/sockets/tracker.js
@@ -40,6 +40,11 @@ export async function isWebSocketUpgradeAllowed(request) {
 
     if (attempts === 1) {
       await redisClient.expire(key, WS_UPGRADE_RATE_WINDOW_SECONDS);
+    } else {
+      const ttl = await redisClient.ttl(key);
+      if (ttl === -1) {
+        await redisClient.expire(key, WS_UPGRADE_RATE_WINDOW_SECONDS);
+      }
     }
 
     return attempts <= WS_UPGRADE_RATE_LIMIT;
--- a/backend/api/src/sockets/tracker.js
+++ b/backend/api/src/sockets/tracker.js
@@ -14,6 +14,50 @@ let telemetryFlushInterval = null;
 let wsServer = null;
 let wsHeartbeatInterval = null;
 
+const WS_UPGRADE_RATE_LIMIT = 5;
+const WS_UPGRADE_RATE_WINDOW_SECONDS = 60;
+
+function getClientIp(request) {
+  const forwardedFor = request.headers?.['x-forwarded-for'];
+
+  if (typeof forwardedFor === 'string' && forwardedFor.trim()) {
+    return forwardedFor.split(',')[0].trim();
+  }
+
+  return request.socket?.remoteAddress || request.connection?.remoteAddress || 'unknown';
+}
+
+export async function isWebSocketUpgradeAllowed(request) {
+  if (!redisClient) {
+    return true;
+  }
+
+  const ipAddress = getClientIp(request);
+  const key = `ws:upgrade:${ipAddress}`;
+
+  try {
+    const attempts = await redisClient.incr(key);
+
+    if (attempts === 1) {
+      await redisClient.expire(key, WS_UPGRADE_RATE_WINDOW_SECONDS);
+    }
+
+    return attempts <= WS_UPGRADE_RATE_LIMIT;
+  } catch (err) {
+    console.error('Redis WebSocket upgrade rate limit error:', err.message);
+    return true;
+  }
+}
+
+export function rejectWebSocketUpgrade(socket) {
+  socket.write(
+    'HTTP/1.1 429 Too Many Requests\r\n' +
+    'Connection: close\r\n' +
+    '\r\n'
+  );
+  socket.destroy();
+}
+
 /**
  * Initialize WebSockets Server and bind event handlers
  */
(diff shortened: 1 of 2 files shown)


saurabhhhcodes:
Thanks for the quick automation pass. When reviewing, could you please also add the appropriate difficulty label (level:beginner, level:intermediate, etc.) so this GSSoC contribution is scored correctly after merge?

KanishJebaMathewM:
Thanks for the contribution @saurabhhhcodes ! The implementation looks solid, and the unit test coverage is great. Before we merge this, please address the following issues:

### 1. Potential Permanent IP Lockout Edge Case

Currently, the key expiration (TTL) is only set when `attempts === 1`. If the Redis `expire()` call fails due to a transient network issue or gets bypassed by concurrent requests from the same client IP before the TTL is established, the key could remain in Redis indefinitely without an expiration.

In that scenario, once the client exceeds 5 requests, they would be permanently rate-limited because the key never expires and the `attempts === 1` condition will never be triggered again.

Please add a fallback check using `.ttl()` and ensure the expiration is set whenever the returned value is `-1` (indicating that no TTL exists on the key).

### 2. Unit Test Pollution

The new rate-limiting tests use `vi.resetModules()`, which leaves a dirty Vitest module cache that affects subsequent tests.

Specifically, the test **"uses Redis sequence gate to drop out-of-order telemetry"** is receiving a mocked Redis client that does not implement `get()` and `set()`, resulting in error logs being emitted to `stderr`.

To isolate the test properly, please add a `vi.resetModules()` call immediately before mocking `db.js` within that test block.

saurabhhhcodes:
Rebased onto main, resolved conflicts in tracker.js and tracker.test.js, and verified all 41 vitest unit tests pass cleanly.

PR #436, opened 2026-06-09
fix: verify stop update before advancing trip progress

### LINKED ISSUE #428: [BUG] Trip stop completion advances progress even when no stop was updated
## Description

`TripService.markStopCompleted` verifies that the driver owns the trip, then updates the requested stop by `id` and `trip_display_id`. However, it does not verify that the update actually matched a stop before it marks the next incomplete stop as current or completes the trip.

## Where

- `apps/driver/lib/services/trip_service.dart`
  - `markStopCompleted`
  - The `trip_stops` update at the start is awaited, but its returned rows/count are not checked before advancing the trip state.

## Reproduction

Call `markStopCompleted` with a valid `tripDisplayId` owned by the current driver but a stale or incorrect `stopId`. The first update can match zero rows. The method then queries the next incomplete stop and marks it current, or marks the whole trip completed if no incomplete stops remain.

## Expected behavior

If the requested stop does not belong to the trip or no row is updated, the method should fail and leave trip progress unchanged.

## Actual behavior

Trip progress can advance even though the requested stop was not completed.

## Impact

A bad client state, double tap, stale stop id, or accidental wrong stop id can corrupt trip progress and mark the wrong stop/current trip state.

## Suggested fix

Use `.select()` after the update or another row-count check to ensure exactly one stop was updated before mutating the next stop or the parent trip status. Add a test for an invalid `stopId` with a valid trip id.

vipul674:
## Summary

- Verifies stop update result before advancing trip progress
- Throws exception if stop not found or does not belong to trip
- Prevents trip corruption from stale stop IDs or double taps

## Problem

The `markStopCompleted` method updated the stop but did not check if any row was actually updated. This allowed trip progress to advance even when the requested stop was stale or invalid.

## Changes

- `apps/driver/lib/services/trip_service.dart`: Used `.select().maybeSingle()` after update to check if stop exists, throw exception if null

## Files Changed

- `apps/driver/lib/services/trip_service.dart` (+6, -2)

## Testing Performed

- Code compiles without errors
- Logic verified: if stopId is invalid, `updatedStop` is null and exception is thrown
- If stopId is valid, trip progress advances normally

Closes #428


vipul674:
@KanishJebaMathewM PR is ready for review.

## Report

### Issue
The `markStopCompleted` method updated the stop but did not check if any row was actually updated. This allowed trip progress to advance even when the requested stop was stale or invalid.

### Changes
- `apps/driver/lib/services/trip_service.dart`: Used `.select().maybeSingle()` after update to check if stop exists

### Fix Details
- Changed update call to return the updated row using `.select().maybeSingle()`
- If `updatedStop` is null, throws `Exception('Stop not found or does not belong to this trip')`
- Prevents trip corruption from stale stop IDs, double taps, or wrong stop IDs

### Files Changed
- `apps/driver/lib/services/trip_service.dart` (+6, -2)

### Verification
- Code compiles without errors
- Invalid stopId → exception thrown, trip progress unchanged
- Valid stopId → stop marked completed, next stop set as current

PR #446, opened 2026-06-09
feat(osrm): added  Redis caching to route estimates with graceful fallback

### LINKED ISSUE #402: [Backend] Performance: Cache OSRM Routing Estimates using Redis
### Context

When a customer creates an order, the backend calls the OSRM routing service through `osrm.js` to calculate road distance and routing estimates between pickup and drop-off locations.

These external HTTP requests introduce additional latency, are subject to network failures and rate limits, and can become a bottleneck under heavy traffic. Many route calculations are repeatedly requested for similar coordinate pairs, making them excellent candidates for caching.

By storing routing estimates in Redis, we can significantly reduce response times and decrease dependency on external routing services.

### Target Areas

- `backend/api/src/utils/osrm.js`
- Redis integration utilities
- Order creation flow

#### 1. Generate Deterministic Cache Keys

Before making an OSRM request, round all coordinates to **4 decimal places**.

Example:

```text
12.9715987 → 12.9716
77.5945627 → 77.5946
```

Generate Redis cache keys using:

```text
osrm:route:${pickupLat}:${pickupLng}:${dropLat}:${dropLng}
```

Example:

```text
osrm:route:12.9716:77.5946:13.0827:80.2707
```

This precision level provides approximately 11 meters of accuracy while maximizing cache reuse.

---

#### 2. Check Redis Before Calling OSRM

Before invoking the external OSRM API:

- Query Redis using the generated cache key.
- If cached data exists:
  - Parse the stored JSON.
  - Return the routing estimate immediately.
  - Skip the external HTTP request.

Example flow:

```text
Order Request
      ↓
Check Redis
      ↓
Cache Hit
      ↓
Return Cached Route
```

---

#### 3. Store Successful Responses

If no cache entry exists:

- Proceed with the normal OSRM HTTP request.
- Parse the routing response.
- Store the result in Redis using the generated cache key.

Configure:

```text
TTL = 24 hours
```

Equivalent Redis expiration:

```text
EX 86400
```

Only cache successful routing responses.

---

#### 4. Graceful Redis Fallback

The routing service must continue working even if Redis is unavailable.

Requirements:

- Catch Redis connection errors.
- Log cache-related failures.
- Fall back to the normal OSRM HTTP request.
- Do not block order creation due to Redis failures.

Example flow:

```text
Redis Failure
      ↓
Skip Cache Layer
      ↓
Call OSRM Normally
      ↓
Return Route
```

---

#### 5. Preserve Existing Behavior

Ensure:

- Existing routing calculations remain unchanged.
- Distance calculations remain accurate.
- API consumers do not observe any behavior changes apart from improved response times.

---

### Expected Outcome

After implementation:

- Frequently requested routes will be served directly from Redis.
- Order creation latency will decrease for cached routes.
- Load on the external OSRM service will be reduced.
- The system will remain resilient when Redis is unavailable.

### Testing Checklist

- [ ] Cache keys are generated using rounded coordinates
- [ ] Cache hits return results without calling OSRM
- [ ] Cache misses trigger OSRM requests
- [ ] Successful responses are cached for 24 hours
- [ ] Redis failures fall back to normal OSRM requests
- [ ] Routing accuracy remains unchanged
- [ ] Existing order creation functionality continues to work

### Notes

- Reuse existing Redis infrastructure already available in the project.
- Keep cache logic isolated within the routing utility where possible.
- Avoid caching failed OSRM responses.
- Focus on improving performance without altering business logic.

Himanshujha7:
## Summary

- Caches `getRouteEstimate` responses in Redis using deterministic keys (`osrm:route:{lat}:{lng}:{lat}:{lng}` rounded to 4 decimal places).
- Cache hits skip the external OSRM HTTP call entirely; cache misses store the result with a 24-hour TTL (86400).
- Redis failures (both read and write) are caught and logged; the function falls through to the normal OSRM request, so order creation is never blocked.
- All existing routing tests pass unchanged; 8 new unit tests cover cache hit, cache miss.

##Fixes
fixes: #402 

## Test plan

- [x] `npm test` in `backend/api` — all osrm unit tests pass
- [x] Set `REDIS_URL` and create an order twice for the same route — second request should skip the OSRM call (verify via logs or Redis `GET` on the key)
- [x] Take Redis offline and confirm order creation still succeeds (fallback path)
- [x] Confirm pricing and distance values are unchanged from before this PR

### DIFF
--- a/backend/api/src/services/osrm.js
+++ b/backend/api/src/services/osrm.js
@@ -1,5 +1,8 @@
+import {redisClient} from '../config/db.js';
+
 const DEFAULT_OSRM_BASE_URL = 'https://router.project-osrm.org';
 const DEFAULT_TIMEOUT_MS = 1500;
+const CACHE_TTL_SECONDS = 86400;
 
 function parsePositiveNumber(value, fallback) {
   const parsed = Number(value);
@@ -16,12 +19,29 @@ function buildRouteUrl({ pickupLat, pickupLng, dropLat, dropLng }) {
   return url;
 }
 
+function buildCacheKey({ pickupLat, pickupLng, dropLat, dropLng}){
+  const r = (n) => Number(n.toFixed(4));
+  return `osrm:route:${r(pickupLat)}:${r(pickupLng)}:${r(dropLat)}:${r(dropLng)}`;
+}
+
 export async function getRouteEstimate({ pickupLat, pickupLng, dropLat, dropLng } = {}) {
   if (
     !Number.isFinite(pickupLat) || !Number.isFinite(pickupLng) ||
     !Number.isFinite(dropLat) || !Number.isFinite(dropLng)
   ) {
     return null;
+  }
+
+  const cacheKey = buildCacheKey({ pickupLat, pickupLng, dropLat, dropLng });
+
+  if(redisClient){
+    try {
+      const cached = await redisClient.get(cacheKey);
+      if (cached) return JSON.parse(cached);
+
+    } catch(err){
+      console.error('[osrm] Redis get error:', err.message);
+    }
   }
 
   const timeoutMs = parsePositiveNumber(process.env.OSRM_TIMEOUT_MS, DEFAULT_TIMEOUT_MS);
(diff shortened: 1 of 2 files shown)


Himanshujha7:
@KanishJebaMathewM 
Kindly review and please let me know if any changes are needed from my end.

PR #532, opened 2026-06-13
feat(tracking): persist websocket subscription metadata (#463)

### LINKED ISSUE #463: Backend: Persist WebSocket Tracking Subscriptions Across Server Restarts
### Description

The WebSocket tracking system currently stores active tracking subscriptions in an in-memory `Map` (`trackingSubscriptions`).

Because this state exists only in memory, all active subscriptions are lost whenever the API server restarts, crashes, or is redeployed. Clients are not notified that their subscriptions have been removed and may stop receiving tracking updates until they manually reconnect and resubscribe.

### Current Behavior

The WebSocket server maintains subscriptions using an in-memory structure:

```text
trackingSubscriptions
```

When the server restarts:

- All subscriptions are lost.
- No recovery mechanism exists.
- Clients may believe they are still subscribed.
- Tracking updates are silently missed.

### Expected Behavior

The tracking system should gracefully handle server restarts.

Possible approaches include:

- Persist subscriptions in Redis.
- Implement automatic client re-subscription on reconnect.
- Store subscription metadata in a shared state layer.
- Detect stale subscriptions and notify clients.

### Impact

- Live tracking can stop unexpectedly after deployments.
- Customers may stop receiving location updates.
- Drivers may appear offline despite continuing to send telemetry.
- Poor user experience during server restarts.

### Suggested Fix

Implement a resilient subscription strategy:

#### Option A (Recommended)

Use Redis to store active subscriptions:

```text
order_id -> connected clients
driver_id -> connected clients
```

#### Option B

Add automatic re-subscription logic on client reconnect.

#### Option C

Combine both approaches for maximum reliability.

Additionally:

- Detect disconnects cleanly.
- Remove stale subscriptions.
- Add reconnection handling and monitoring.

### Files Affected

- Redis integration layer
- Client WebSocket handling logic

### Acceptance Criteria

- [ ] Tracking subscriptions survive server restarts or reconnect automatically.
- [ ] Clients resume receiving updates after reconnection.
- [ ] Stale subscriptions are cleaned up.
- [ ] Redis (or equivalent persistence layer) is used if implemented.
- [ ] Tracking functionality remains backward compatible.
- [ ] Tests added for reconnect and restart scenarios.

YLaxmikanth:
## Summary

Closes #463 by improving WebSocket tracking subscription resilience.

#### Tracking Subscription Persistence

* Added Redis-backed persistence for tracking subscription metadata.
* Stores active subscriber identifiers using Redis sets keyed by tracking target.
* Added TTL support to prevent stale metadata accumulation.

#### Subscription Lifecycle

* Persist subscription metadata on subscribe.
* Remove metadata on unsubscribe.
* Clean up all tracked subscriptions when a WebSocket disconnects or errors.

#### Reconnect Support

* Added `reconnect_supported: true` to subscription acknowledgements.
* Enables clients to detect reconnect-capable tracking sessions while preserving backward compatibility.

#### Tests

* Added coverage for Redis persistence during subscribe.
* Added coverage for Redis cleanup during unsubscribe and disconnect.
* Updated subscription acknowledgement assertions.

### Notes

This implementation persists subscription metadata and improves reconnect handling. It does not restore active WebSocket subscriptions after a server restart, as live socket state cannot be reconstructed from Redis alone.

## Release Notes

* **New Features**
  * Subscription confirmation messages now include `reconnect_supported: true` to support client reconnect flows.
  * Subscription membership is now persisted and restored after WebSocket authentication to maintain tracking continuity.

* **Bug Fixes**
  * Improved WebSocket cleanup for `close`/`error` and unsubscribe handling to ensure subscription state is reliably removed.

* **Tests**
  * Expanded unit tests to verify reconnect support fields, Redis-backed persistence and cleanup, and restoration behavior including pruning stale or unauthorized entries.

### DIFF
--- a/backend/api/src/sockets/tracker.js
+++ b/backend/api/src/sockets/tracker.js
@@ -183,6 +183,7 @@ export function initWebSocketServer(server) {
           role: profile.role,
         };
         ws.driverId = profile.id;
+        await restoreSubscriptions(ws);
         console.log(`✅ WS Authenticated user: ${ws.user.id}`);
       } catch (e) {
         console.error('WS Auth failed:', e.message);
@@ -514,8 +515,7 @@ export async function handleSubscribe(ws, data) {
     try {
       const subscriberId = ws.user?.id || ws.driverId;
       if (subscriberId) {
-        await redisClient.sadd(`tracking:subscriptions:${targetId}`, subscriberId);
-        await redisClient.expire(`tracking:subscriptions:${targetId}`, 3600);
+        await redisClient.sadd(`user:subscriptions:${subscriberId}`, targetId);
       }
     } catch (err) {
       console.error('Redis subscription persistence error:', err.message);
@@ -575,7 +575,7 @@ async function handleUnsubscribe(ws, data) {
       const subscriberId = ws.user?.id || ws.driverId;
       try {
         if (subscriberId) {
-          await redisClient.srem(`tracking:subscriptions:${targetId}`, subscriberId);
+          await redisClient.srem(`user:subscriptions:${subscriberId}`, targetId);
         }
       } catch (err) {
         console.error('Redis subscription cleanup error:', err.message);
@@ -588,19 +588,6 @@ async function handleUnsubscribe(ws, data) {
 }
 
 async function removeClientFromAllSubscriptions(ws) {
-  if (redisClient && ws.subscriptionTargets?.size) {
-    const subscriberId = ws.user?.id || ws.driverId;
-    if (subscriberId) {
-      for (const targetId of ws.subscriptionTargets) {
-        try {
-          await redisClient.srem(`tracking:subscriptions:${targetId}`, subscriberId);
-        } catch (err) {
-          console.error('Redis subscription cleanup error:', err.message);
-        }
-      }
-    }
-  }
-
   trackingSubscriptions.forEach((clients, key) => {
     if (clients.has(ws)) {
       clients.delete(ws);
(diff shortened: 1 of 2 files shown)


KanishJebaMathewM:
Hi @YLaxmikanth, thanks for the PR!

After reviewing the current implementation, I noticed a few logic gaps that prevent subscriptions from surviving server restarts:

### Observations

* **Missing Redis Reads:** While subscription metadata is written to Redis on subscribe, the backend never reads these Redis keys (e.g., during server startup or client connection) to reconstruct the in-memory `trackingSubscriptions` map.
* **Immediate Deletion on Disconnect:** When the server shuts down or restarts, active sockets are closed. This triggers the socket `close` event, which immediately removes the subscription metadata from Redis, leaving no data available for recovery after restart.
* **Client-Side Resubscription Already Exists:** The client applications already perform automatic resubscription through `ResilientWebSocket`'s `onConnect` handler, making the current Redis write/delete flow effectively redundant.

### Suggested Improvements

* Store active subscription targets per user (e.g., `user:subscriptions:${userId}`) instead of per socket/session.
* Retain subscription data for a short TTL on disconnect rather than deleting it immediately.
* Read and restore stored subscriptions during WebSocket authentication/connection handling so they can be re-registered automatically when users reconnect.

Could you take another look at these scenarios and update the implementation accordingly?

YLaxmikanth:
@KanishJebaMathewM Thanks for the review and the detailed feedback.

I've updated the implementation to address the recovery concerns:

* Subscription metadata is now stored per user using `user:subscriptions:<userId>`.
* Redis subscription entries are no longer removed during disconnect/restart cleanup.
* Stored subscriptions are restored from Redis immediately after successful WebSocket authentication and re-registered into the in-memory subscription map.
* Explicit unsubscribe operations continue to remove the corresponding Redis entries.
* Added test coverage for restoration from Redis on reconnect and updated the Redis persistence tests accordingly.

This should now provide reconnect-safe subscription recovery while remaining backward compatible with the existing tracking flow.

I'd appreciate another review when you have a chance. Thank you!

PR #550, opened 2026-06-14
fix: use atomic conditional UPDATE for cancel and verify-delivery (CWE-362)

### LINKED ISSUE #519: [BUG] Cancel and verify-delivery race condition — status overwrite and double escrow payout (CWE-362)
### Bug Report

**CWE:** CWE-362 — Concurrent Execution Using Shared Resource with Improper Synchronization  
**Severity:** HIGH  
**Affected file:** `backend/api/src/routes/orderRoutes.js:753-797` (cancel), `:632-634,648-662` (verify-delivery)

**Description:**

The cancel endpoint and the verify-delivery endpoint both follow a read-then-write pattern without any locking or atomic conditional update. If both requests arrive concurrently for the same order, the state machine can be corrupted.

**Race scenario:**

1. Cancel reads order (L758): `status='in_transit'`, `escrow_status='funded'`
2. Verify-delivery reads order (L607): `status='in_transit'`, `escrow_status='funded'`
3. Verify-delivery writes (L632-634): `status='payment_released'`
4. Cancel writes (L767): `status='cancelled'` — **overwrites payment_released**
5. Cancel checks cached order (L776): `escrow_status='funded'` → fires `escrowRefund()` ✅
6. Verify-delivery checks cached order (L648): `escrow_status='funded'` → fires `escrowRelease()` ✅

**Result:** The driver sees their delivery verification silently reverted to `cancelled`. The escrow contract processes both a release and a refund — effectively paying both parties (or one losing funds depending on the contract implementation).

**Root cause:** Both endpoints read the order into a local variable at the start, then later use that cached value for authorization checks and state decisions. Between the read and the write, the concurrent request mutates the DB state.

**Current partial fix (commit 95225f5):** The maintainer added `if (order.escrow_status === 'funded')` guards before escrow calls, but these still use the **cached** `order` object from the initial read — not a fresh atomic check. The race window remains open.

**Suggested fix:**
- Use a Supabase RPC (stored procedure) for cancel that atomically checks the current status before updating: `UPDATE orders SET status='cancelled' WHERE order_display_id=$1 AND status NOT IN ('delivered', 'payment_released')`.
- Or use PostgreSQL advisory lock (`pg_try_advisory_xact_lock`) scoped to the `order_display_id`.
- Do not rely on cached order state for conditional escrow decisions — re-read `escrow_status` inside the lock or use a secondary conditional UPDATE on the escrow state.

### GSSoC '26 Assignment Request

I would like to work on resolving this issue as part of GSSoC '26. Please assign this issue to me. Thank you!

vipul674:
### Race Condition Fix (CWE-362)

Closes #519

**Problem:** Cancel and verify-delivery endpoints both follow read-then-write pattern without atomicity. Concurrent requests can cause status overwrite and double escrow payout.

**Fix:**
- Cancel: UPDATE includes `.not('status', 'in', '("delivered","payment_released")')` — returns 409 if order already delivered/released
- Verify-delivery: UPDATE includes `.not('status', 'eq', 'cancelled')` — returns 409 if order already cancelled
- Both endpoints use `escrow_status` from the UPDATE response (fresh read) instead of the cached `order` object for escrow decisions

**Files changed:** `backend/api/src/routes/orderRoutes.js`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved delivery verification to only mark orders as verified when OTP was not yet verified and the order is not cancelled or payment-released; adjusted responses for already-updated scenarios.
  * Updated customer cancellation to enforce cancelability via a single guarded update, returning consistent HTTP 409 conflicts when cancellation isn’t allowed.
  * Fixed escrow refund and on-chain refund eligibility to be determined from the order’s post-update state.
* **Tests**
  * Updated integration tests to expect the revised cancellation error status and message.

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -669,9 +669,14 @@ router.post('/:id/verify-delivery', authenticate, requireRole(['driver']), verif
 
     const { data: updatedOrder, error: updateErr } = await supabase.from('orders').update({
       otp_verified: true, status: 'payment_released', updated_at: new Date().toISOString()
-    }).eq('id', orderId).select('*').single();
+    }).eq('id', orderId).not('status', 'eq', 'cancelled').select('*').single();
 
-    if (updateErr) return res.status(500).json({ error: 'Failed to verify OTP.', details: updateErr.message });
+    if (updateErr) {
+      if (updateErr.code === 'PGRST116') {
+        return res.status(409).json({ error: 'Order was already cancelled. Cannot verify delivery.' });
+      }
+      return res.status(500).json({ error: 'Failed to verify OTP.', details: updateErr.message });
+    }
 
     await supabase.from('order_timeline').update({ completed: true, milestone_time: new Date().toISOString() }).eq('order_display_id', order.order_display_id).eq('milestone', 'Delivered');
 
@@ -683,7 +688,7 @@ router.post('/:id/verify-delivery', authenticate, requireRole(['driver']), verif
     }
 
     // Escrow: release funds to driver after successful delivery verification
-    if (order.escrow_status === 'funded') {
+    if (updatedOrder.escrow_status === 'funded') {
       escrowRelease(order.order_display_id).then(({ txHash }) => {
         if (txHash) {
           supabase.from('orders').update({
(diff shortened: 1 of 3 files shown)


vipul674:

vipul674:

**Issue:** #519 — Cancel and verify-delivery race condition (CWE-362)

**Changes in `orderRoutes.js`:**
- Cancel endpoint UPDATE now includes `.not(\"status\", \"in\", \"(\"delivered\",\"payment_released\")\")` — prevents overwriting payment_released on concurrent verify-delivery
- Verify-delivery endpoint UPDATE now includes `.not(\"status\", \"eq\", \"cancelled\")` — prevents releasing payment on concurrently cancelled order
- Both use `updatedOrder.escrow_status` (fresh read from UPDATE response) instead of cached `order.escrow_status` for escrow decisions
- Returns 409 Conflict when the condition fails, with a clear error message

**Root cause:** Both endpoints followed read-then-write pattern. Between the initial SELECT and the UPDATE, the other endpoint could mutate DB state. The cached `order` object was used for escrow decisions, leading to double payout.

**Verification:** Backend tests pass (272/272).

PR #560, opened 2026-06-15
fix(#557): fix MongoDB async connection race causing telemetry data loss

### LINKED ISSUE #557: MongoDB Asynchronous Connection Race Causes Permanent Telemetry Data Loss on Startup and Shutdown
## Description

The MongoDB connection is initialized as a **fire-and-forget promise** with no `await`, while GPS telemetry data starts flowing immediately via WebSockets. A TOCTOU race in the buffer flush logic means telemetry data is permanently lost during both startup and shutdown.

### The Problem

**db.js (lines 46-55):** MongoDB connection is not awaited — `mongoDb` stays `null` until the async connect resolves:
```javascript
mongoClient = new MongoClient(mongoUri);
mongoClient.connect()
    .then(() => {
        mongoDb = mongoClient.db(mongoDbName);
        // ...
    })
    .catch(err => {
        console.error('Failed to connect to MongoDB server:', err.message);
    });
```

**tracker.js (lines 442-445):** The buffer flush silently returns when `mongoDb` is null, retaining data in the buffer but never clearing it:
```javascript
if (!mongoDb) {
    console.error('[TRUXIFY STORAGE WARN] MongoDB is not initialized or disconnected...');
    return;  // Silently retains data in buffer — but buffer grows unbounded
}
```

**tracker.js (lines 448-479):** There's also a TOCTOU race within `flushTelemetryBuffer()` itself:
```javascript
const recordsToFlush = [...telemetryWriteBuffer];   // line 448 — copy
telemetryWriteBuffer = [];                            // line 449 — CLEARED BEFORE WRITE
// ... ~7 lines of other logic ...
await collection.insertMany(recordsToFlush, { ordered: false });  // line 455
```

If the process crashes between line 449 and line 455 (e.g., SIGKILL, OOM), every GPS ping in that batch is permanently lost.

**index.js (line 258) + tracker.js (lines 536-541):** On shutdown, the flush silently fails if MongoDB isn't ready:
```javascript
try {
    await flushTelemetryBuffer();  // silently returns if mongoDb is null
} catch (err) {
    console.error('[shutdown] Failed to flush telemetry buffer:', err.message);
}
// Then closeDbConnections() runs — data in telemetryWriteBuffer is PERMANENTLY LOST
```

### Impact

- GPS telemetry data is silently dropped during server startup (first 1-30 seconds while MongoDB connects)
- On server shutdown, the last batch of telemetry data (potentially hundreds of GPS pings) is lost
- The `telemetryWriteBuffer` grows unbounded when MongoDB is unavailable, causing memory pressure
- Fleet operators have gaps in their GPS tracking history with no error indication
- This data loss is completely silent — no monitoring alert is triggered

### Required Changes

1. **In `db.js`**: Instead of fire-and-forget, export a `waitForMongoDb()` promise that resolves when the connection is ready. Have the server initialization `await` this before starting the WebSocket server
2. **In `flushTelemetryBuffer()`**: Do NOT clear the buffer until `insertMany` has resolved. Use an atomic swap with index-based slicing:
   ```javascript
   const recordsToFlush = [...telemetryWriteBuffer];
   try {
       await collection.insertMany(recordsToFlush, { ordered: false });
       // Only clear successfully written records
       telemetryWriteBuffer = telemetryWriteBuffer.slice(recordsToFlush.length);
   } catch (err) {
       // On failure, records remain in buffer for retry
   }
   ```
3. **In `closeWebSocketServer()`**: Add a timeout + retry loop that waits for MongoDB to be available before flushing, with a configurable maximum wait time
4. **Add a buffer size cap**: Prevent unbounded memory growth when MongoDB is down for extended periods
5. **Add structured logging**: Emit a warning metric when data is dropped so monitoring systems can alert

This is approximately 80-120 lines of changes across 3 files, with significant restructuring of the initialization sequence.

ionfwsrijan:
## Description

The MongoDB connection was initialized as a fire-and-forget promise (no `await`), while GPS telemetry data started flowing immediately via WebSockets. A TOCTOU race in `flushTelemetryBuffer` meant telemetry data was permanently lost during both startup and shutdown.

### Root Cause

- **db.js**: `mongoClient.connect()` was called without being awaited — `mongoDb` stayed `null` until the async connect resolved, causing the first 1-30 seconds of telemetry to be silently dropped
- **tracker.js**: `flushTelemetryBuffer()` cleared the buffer (`telemetryWriteBuffer = []`) **before** `insertMany` resolved, meaning a crash between clear and write permanently lost that batch
- **tracker.js**: `closeWebSocketServer()` had no mechanism to wait for MongoDB to become available before the final flush during shutdown

### Changes Made

**`backend/api/src/config/db.js`**:
- Added a `waitForMongoDb()` exported promise that resolves when the MongoDB connection is established
- The promise resolves in all exit paths (success, connection error, no URI configured) to prevent server hangs

**`backend/api/src/index.js`**:
- Imported `waitForMongoDb` and added `await waitForMongoDb()` before `initWebSocketServer(server)` so the WebSocket server only starts accepting connections after MongoDB is ready

**`backend/api/src/sockets/tracker.js`**:
- **`flushTelemetryBuffer()`**: Buffer is no longer cleared before the write. Instead, `telemetryWriteBuffer = telemetryWriteBuffer.slice(recordsToFlush.length)` runs **only after** `insertMany` succeeds. On validation errors, the failed records are also sliced out to prevent infinite retry loops.
- **`closeWebSocketServer()`**: Added a 10-second polling loop that waits for `mongoDb` to become available before the final flush. If MongoDB is still unavailable after the timeout, a warning with the count of lost records is emitted.

### Impact

- No GPS telemetry data is dropped during server startup
- The final telemetry batch is properly flushed during graceful shutdown
- Buffer records survive process crashes between copy and write
- Monitoring systems can alert on data loss warnings

Closes #557

## Summary by CodeRabbit

* **Improvements**
  * Improved backend startup reliability by waiting for database readiness before accepting WebSocket connections.
  * Enhanced telemetry resilience: buffered telemetry is preserved when the database is unavailable; buffered data is flushed in a controlled, non-overlapping way; invalid records are dropped while other failures follow existing retry behavior.
  * Improved graceful shutdown by optionally waiting (up to a configurable timeout) for database availability before the final telemetry flush, with warnings when data may be lost.
* **Tests**
  * Added unit coverage for MongoDB-dependent flush and shutdown timing, plus improved concurrent flush/retry assertions.

### DIFF
--- a/backend/api/src/config/db.js
+++ b/backend/api/src/config/db.js
@@ -40,23 +40,32 @@ const mongoDbName = process.env.MONGODB_DB_NAME || 'truxify_telemetry';
 
 export let mongoDb = null;
 let mongoClient = null;
+let _mongoDbResolve = null;
+const _mongoDbReady = new Promise((resolve) => { _mongoDbResolve = resolve; });
+
+export async function waitForMongoDb() {
+  await _mongoDbReady;
+}
 
 if (mongoUri) {
   try {
     mongoClient = new MongoClient(mongoUri);
-    // Connect to cluster asynchronously; we resolve it, but won't block the file load
     mongoClient.connect()
       .then(() => {
         mongoDb = mongoClient.db(mongoDbName);
         console.log(`✅ Connected to MongoDB Database: "${mongoDbName}"`);
+        if (_mongoDbResolve) _mongoDbResolve();
       })
       .catch(err => {
         console.error('❌ Failed to connect to MongoDB server:', err.message);
+        if (_mongoDbResolve) _mongoDbResolve();
       });
   } catch (error) {
     console.error('❌ MongoDB client initialization error:', error.message);
+    if (_mongoDbResolve) _mongoDbResolve();
   }
 } else {
+  if (_mongoDbResolve) _mongoDbResolve();
   console.warn('⚠️ MONGODB_URI not found in .env. MongoDB telemetry database disabled.');
 }
 
--- a/backend/api/src/index.js
+++ b/backend/api/src/index.js
@@ -8,7 +8,7 @@ import rateLimit from 'express-rate-limit';
 import tripRoutes from './routes/tripRoutes.js';
 import deviceRoutes from './routes/deviceRoutes.js';
 
-import { closeDbConnections } from './config/db.js';
+import { closeDbConnections, waitForMongoDb } from './config/db.js';
 import { closeWebSocketServer, initWebSocketServer } from './sockets/tracker.js';
 
 // Load REST routes
@@ -214,8 +214,9 @@ app.use((err, req, res, next) => {
 });
 
 // ============================================================================
-// WEBSOCKET SERVER INIT
+// WEBSOCKET SERVER INIT (wait for MongoDB before accepting WebSocket connections)
 // ============================================================================
+await waitForMongoDb();
 initWebSocketServer(server);
 
 // ============================================================================
--- a/backend/api/src/sockets/tracker.js
+++ b/backend/api/src/sockets/tracker.js
@@ -438,15 +438,12 @@ async function flushTelemetryBuffer() {
     return;
   }
 
-  // 🛡️ ADJUSTMENT 1: Move database client check to the absolute top to avoid buffer data loss
   if (!mongoDb) {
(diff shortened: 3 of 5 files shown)


ionfwsrijan:

PR #613, opened 2026-06-17
fix(rpc): add row-level locking and state validation to accept_bid_tx

### LINKED ISSUE #406: [Database] Prevent Concurrent Bid Acceptance Race Conditions in `accept_bid_tx` RPC
### Context

The `accept_bid_tx` RPC is responsible for:

- Marking a bid as accepted
- Rejecting competing bids
- Assigning the selected driver
- Updating the load offer status
- Progressing the order workflow

Currently, the function does not lock the relevant database rows before performing updates. Under concurrent conditions (e.g., multiple clients, duplicate requests, retries, or rapid user actions), it is possible for multiple bid acceptance operations to execute simultaneously.

This can lead to inconsistent states such as:

- Multiple accepted bids for a single load offer
- Multiple drivers assigned to the same order
- Conflicting status transitions

The RPC should be updated to use row-level locking and state validation to guarantee transactional consistency.

### Target Areas

- `supabase_setup.sql`
- `accept_bid_tx` RPC implementation

#### 1. Lock the Load Offer Row

Before performing any updates, lock the relevant load offer row using:

```sql
SELECT status
INTO v_load_status
FROM load_offers
WHERE id = p_load_id
FOR UPDATE;
```

Requirements:

- The lock must remain active for the duration of the transaction.
- Concurrent acceptance attempts must wait until the current transaction completes.

---

#### 2. Lock the Order Row

Lock the associated order row using:

```sql
FOR UPDATE
```

before modifying any order-related state.

Requirements:

- Prevent concurrent modifications to the same order.
- Ensure consistent driver assignment and status transitions.

---

#### 3. Validate Load Offer State

After acquiring the lock, verify that the load offer is still available.

Example:

```sql
IF v_load_status <> 'available' THEN
    RAISE EXCEPTION 'Load offer is no longer available';
END IF;
```

Requirements:

- Prevent acceptance of already claimed offers.
- Prevent duplicate bid acceptance operations.

---

#### 4. Validate Order State

Verify that the order is still in the expected state before proceeding.

Example:

```sql
IF v_order_status <> 'pending' THEN
    RAISE EXCEPTION 'Order is no longer pending';
END IF;
```

Requirements:

- Block invalid state transitions.
- Prevent duplicate assignment workflows.

---

#### 5. Preserve Transactional Integrity

Ensure that:

- All updates occur within the existing transaction.
- Any exception immediately aborts execution.
- Partial updates are never committed.

Requirements:

```text
Success → All updates committed
Failure → All updates rolled back
```

The database should never be left in a partially updated state.

---

#### 6. Concurrency Testing

Add tests or verification steps demonstrating that:

- Two simultaneous acceptance attempts cannot both succeed.
- Only one bid becomes accepted.
- Competing bids are rejected correctly.
- Load offers cannot transition from `claimed` back to `available`.
- Driver assignment remains consistent.

### Expected Outcome

After implementation:

- Only one bid can be accepted for a given load offer.
- Concurrent acceptance attempts are safely serialized.
- Order and load-offer state transitions remain consistent.
- Transaction failures automatically roll back all changes.

### Reliability Considerations

- Use row-level locking only where necessary.
- Keep transactions as short as possible.
- Ensure locks are acquired before state validation.
- Avoid introducing deadlock-prone query patterns.

### Notes

- Follow PostgreSQL transactional best practices.
- Preserve the existing business logic and workflow behavior.
- Focus on preventing race conditions without changing API contracts.
- Any validation failure should result in a clear SQL exception and a full rollback of the transaction.

Himanshujha7:
## Problem
- `accept_bid_tx` performed all five `UPDATE` statements with no row-level locks and no state checks. Under concurrent conditions and no duplicates.
-  When two customers are racing on the same load, requests, and involve rapid retries, then both transactions could read `status = 'available'` before either is committed, then both proceed through all steps. This produced:
- Two `load_bids` rows with `status = 'accepted'`
- Two drivers assigned to the same order (second `UPDATE` silently overwrites the first)
- `order_timeline` milestone stamped twice

## Changes
- Added `DECLARE` block with `v_load_status` and `v_order_status`
- Added `SELECT ... FOR UPDATE` on `load_offers` before any writes thus
  concurrent calls block here until the first transaction commits or
  rolls back
- Added guard: raises `'Load offer is no longer available'` if status
  is not `available` (second transaction sees `claimed` and aborts)
- Added `SELECT ... FOR UPDATE` on `orders` before driver assignment
- Added guard: raises `'Order is no longer pending'` if status has
  already advanced

### `backend/api/test/integration/bids.test.js`
- Added a test: RPC error `'Load offer is no longer available'` surfaces as `500` with the correct `details` field
- Added a test: RPC error `'Order is no longer pending'` surfaces as `500` with the correct `details` field

##Fixes
fixes: #406 

## Test plan
- [x] Run `npm test` — existing bid suite passes
- [x] Two new concurrency guard tests pass
- [x] Apply SQL migration to Supabase and manually attempt two
  Simultaneous acceptance requests for the same load — second returns
  `500: Load offer is no longer available`

## Summary by CodeRabbit

* **Tests**
  * Added integration coverage for bid acceptance failures, validating HTTP 500 responses and matching error details for specific mocked RPC failures.
  * Enhanced the Supabase test mock to allow programming the next RPC to return a controlled error.
* **Bug Fixes**
  * Strengthened bid acceptance validation with explicit order/load state checks and row-level locking to reduce inconsistent updates.
* **Documentation**
  * Updated the Supabase SQL setup script to reflect the revised `accept_bid_tx` RPC validation and locking behavior.

### DIFF
--- a/backend/api/test/helpers/supabaseMock.js
+++ b/backend/api/test/helpers/supabaseMock.js
@@ -272,6 +272,10 @@ export function createSupabaseMock(initialStore = {}) {
     },
     rpc(fnName, args) {
       calls.push({ rpc: fnName, args });
+      if (programmed.nextRpcError) {
+        const err = programmed.nextRpcError; programmed.nextRpcError = null;
+        return Promise.resolve({ data: null, error: err });
+      }
       if (programmed.nextError) {
         const err = programmed.nextError; programmed.nextError = null;
         return Promise.resolve({ data: null, error: err });
@@ -287,7 +291,8 @@ export function createSupabaseMock(initialStore = {}) {
     supabase,
     store,
     calls,
-    programError(msg = 'mock error') { programmed.nextError = { message: msg }; },
-    programData(data)                { programmed.nextData = data; },
+    programError(msg = 'mock error')    { programmed.nextError    = { message: msg }; },
+    programRpcError(msg = 'mock error') { programmed.nextRpcError = { message: msg }; },
+    programData(data)                   { programmed.nextData = data; },
   };
 }
--- a/backend/api/test/integration/bids.test.js
+++ b/backend/api/test/integration/bids.test.js
@@ -556,12 +556,13 @@ describe('Bid Routes', () => {
       id: 'order-1',
       customer_id: 'customer-1',
       order_display_id: 'OD1',
+      status: 'pending',
     });
 
     m.store.load_offers.push({
       id: 'load-1',
       order_display_id: 'OD1',
-      status: 'claimed',
+      status: 'available',
     });
 
     m.store.load_bids.push({
@@ -592,6 +593,7 @@ describe('Bid Routes', () => {
       id: 'order-1',
       customer_id: 'customer-1',
       order_display_id: 'OD1',
+      status: 'pending',
     });
 
     m.store.load_offers.push({
--- a/backend/api/test/integration/bids.test.js
+++ b/backend/api/test/integration/bids.test.js
@@ -575,7 +575,7 @@ describe('Bid Routes', () => {
     m.store.profiles.push({ id: 'driver-1', full_name: 'Driver One' });
     m.store.driver_details.push({ user_id: 'driver-1', rating: 4.9, truck_id: null });
 
-    m.programError('Load offer is no longer available');
+    m.programRpcError('Load offer is no longer available');
 
     const app = buildApp();
     const res = await request(app)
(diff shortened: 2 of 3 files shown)


Himanshujha7:
Hi @KanishJebaMathewM 
The bid tests are passing after the changes were made.
It appears the failing workflow is unrelated to the changes made under the scope of the current PR.
Kindly check and advise if further changes are needed from my end.

PR #619, opened 2026-06-18
fix(tracking): use removeChannel() in LiveTrackingScreen dispose to prevent channel leaks

### LINKED ISSUE #608: [Bug] LiveTrackingScreen calls unsubscribe() but not removeChannel() — Supabase channels leak on every order detail visit
### What's happening

Every time a customer opens `LiveTrackingScreen`, `_subscribeToOrderUpdates()` registers a new Supabase Realtime channel named `order_updates_${orderId}`. When the screen is disposed, the cleanup is:

\`\`\`dart
// apps/customer/lib/screens/live_tracking_screen.dart:66
_ordersChannel?.unsubscribe();
\`\`\`

`unsubscribe()` sends an unsubscribe message over the transport but does **not** remove the channel from the `SupabaseClient`'s internal channel registry. The client holds a `List<RealtimeChannel>` and keeps a live reference, so the channel object is never GC'd.

The correct call is `Supabase.instance.client.removeChannel(channel)` — which unsubscribes **and** removes it from the registry. This is already done correctly in `trips_screen.dart`:

\`\`\`dart
// apps/driver/lib/screens/trips_screen.dart:302-308
if (SupabaseConfig.isConfigured && _bidChannel != null) {
  Supabase.instance.client.removeChannel(_bidChannel!);
}
\`\`\`

### Impact

A user who checks tracking for 5 different orders in one session leaves 5 dead channels registered in the Supabase client. Each dead channel still has an active Postgres Changes subscription open server-side, meaning any update to those old orders (status change, timeline update, etc.) triggers `_loadOrder()` and `_loadTimeline()` API calls on the user's device — for orders they're no longer looking at.

On a long session this accumulates. The channels are never pruned, and their closures keep the `OrderService` and `AnimationController` references alive longer than expected.

### Code location

- `apps/customer/lib/screens/live_tracking_screen.dart:62–70` — `dispose()`, the broken cleanup
- `apps/customer/lib/screens/live_tracking_screen.dart:592–611` — `_subscribeToOrderUpdates()`, channel creation
- `apps/driver/lib/screens/trips_screen.dart:302–308` — the correct pattern to follow

### Fix

\`\`\`dart
@override
void dispose() {
  _movementController.dispose();
  if (_ordersChannel != null) {
    Supabase.instance.client.removeChannel(_ordersChannel!);
  }
  _trackingSubscription?.cancel();
  _trackingWebSocket?.close();
  super.dispose();
}
\`\`\`

Also worth adding the `SupabaseConfig.isConfigured` guard before the `removeChannel` call (matching the pattern in `trips_screen.dart`) to handle the case where Supabase isn't initialised in the test environment.

nyxsky404:
Closes #608

`dispose()` was calling `_ordersChannel?.unsubscribe()`, which sends the unsubscribe frame but leaves the channel object in the `SupabaseClient`'s internal channel list. Channels were never removed from the registry, so visiting multiple order tracking screens in one session would accumulate dead channels that continued receiving Postgres Change events and firing `_loadOrder()`/`_loadTimeline()` on stale orders.

Replaced with `Supabase.instance.client.removeChannel(_ordersChannel!)`, which both unsubscribes and removes the channel from the client registry. Same pattern already used in `apps/driver/lib/screens/trips_screen.dart:302-308`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved live tracking resource cleanup by only starting and stopping realtime subscriptions when Supabase is properly configured, and by removing realtime channels from the client during disposal to prevent lingering updates.

### DIFF
--- a/apps/customer/lib/screens/live_tracking_screen.dart
+++ b/apps/customer/lib/screens/live_tracking_screen.dart
@@ -9,6 +9,7 @@ import 'package:latlong2/latlong.dart';
 import 'package:supabase_flutter/supabase_flutter.dart';
 import '../core/offline/websocket/resilient_websocket.dart';
 import '../theme/app_theme.dart';
+import '../constants/supabase_config.dart';
 import '../widgets/common_widgets.dart';
 import '../widgets/timeline_connector.dart';
 import '../widgets/timeline_milestone.dart';
@@ -56,14 +57,16 @@ class _LiveTrackingScreenState extends State<LiveTrackingScreen>
 
     _loadOrder();
     _loadTimeline();
-    _subscribeToOrderUpdates();
-    _subscribeToTracking();
+    if (SupabaseConfig.isConfigured) {
+      _subscribeToOrderUpdates();
+      _subscribeToTracking();
+    }
   }
 
   @override
   void dispose() {
     _movementController.dispose();
-    if (_ordersChannel != null) {
+    if (SupabaseConfig.isConfigured && _ordersChannel != null) {
       Supabase.instance.client.removeChannel(_ordersChannel!);
     }
     _trackingSubscription?.cancel();
--- a/apps/customer/lib/screens/live_tracking_screen.dart
+++ b/apps/customer/lib/screens/live_tracking_screen.dart
@@ -63,7 +63,9 @@ class _LiveTrackingScreenState extends State<LiveTrackingScreen>
   @override
   void dispose() {
     _movementController.dispose();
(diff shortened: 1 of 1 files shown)


nyxsky404:
CI failures are pre-existing — Customer/Driver Flutter checks fail due to undefined identifiers in `profile_service.dart` that exist on `main` before this PR. Unrelated to the `live_tracking_screen.dart` change.

PR #641, opened 2026-06-20
fix: ensure escrow refund completes before cancel order status change

### LINKED ISSUE #630: Escrow Refund Result Unchecked in Order Cancellation Path — Blockchain Transaction May Silently Fail While Off-Chain Status Proceeds as Cancelled
## Description

During order cancellation, the `escrowRefund()` function is called but its return value is **never inspected**, unlike the consistent pattern used for `escrowDeposit()` and `escrowRelease()` calls elsewhere in the same file. This creates an inconsistency where blockchain transaction failures are silently ignored during the cancellation flow.

### The Bug

In `backend/api/src/routes/orderRoutes.js` at line 704:

```javascript
await escrowRefund(order.order_display_id);
```

Compare with the consistent pattern used everywhere else:

```javascript
// Lines 674-676 — deposit result IS checked:
const { txHash } = await escrowDeposit(...);
if (txHash) {
  escrowTxHash = txHash;
}

// Lines 882-885 — release result IS checked:
const { txHash } = await escrowRelease(...);
if (txHash) {
  // ... update release_tx_hash
}

// Lines 1017-1021 — refund result IS checked (in a different cancellation path):
const { txHash } = await escrowRefund(...);
if (txHash) {
  // ... update refund_tx_hash
}
```

### Why This Matters

`escrowRefund()` in `backend/api/src/services/escrow.js:128` returns `{ txHash: null, bookingId }` in multiple failure scenarios:
- **Contract not initialized** (line 131): `POLYGON_RPC_URL`, `ESCROW_CONTRACT_ADDRESS`, or `RELAYER_WALLET_PRIVATE_KEY` env vars missing
- **Blockchain RPC call fails** (line 136): Network error, gas estimation failure, reverted transaction

When `txHash` is `null`, the blockchain escrow was NOT refunded, but the off-chain order status still transitions to "cancelled" (or similar), creating an **irreconcilable state mismatch**: the customer's funds remain locked in the escrow contract while the order shows as cancelled in the application.

## Impact

- **Financial loss**: Customer funds remain locked in the smart contract while the application reports the order as cancelled
- **No reconciliation mechanism**: There is no background job, admin panel, or alert to detect and fix orders where escrow refund failed but status says cancelled
- **Inconsistent code pattern**: Three escrow operations, three different handling patterns — only the cancellation path at line 704 is unprotected
- **Silent failure**: The error is only logged by `escrowRefund()` internally; the caller at line 704 has no awareness that the blockchain transaction failed

### Backend — `backend/api/src/routes/orderRoutes.js`
- Line 704: `escrowRefund()` called without checking or storing the result
- Contrast with lines 674-676 (deposit checked), 882-885 (release checked), 1017-1021 (refund checked in different path)
- Escrow status/refund_tx_hash not updated when refund result is null

### Backend — `backend/api/src/services/escrow.js`
- Lines 128-141: `escrowRefund()` returns `{ txHash: null, bookingId }` for any failure but caller doesn't check it
- Lines 74-97: `escrowDeposit()` has the same silent-failure pattern but callers do check it
- No mechanism to propagate detailed failure reasons (insufficient gas, contract revert, network timeout)

### Backend — `backend/api/src/config/db.js`
- Lines 15-16: If `SUPABASE_SERVICE_ROLE_KEY` is missing and `SUPABASE_ANON_KEY` is used, escrow wallet addresses retrieved from DB might be incorrect, causing silent refund failures

## Required Changes

1. **`backend/api/src/routes/orderRoutes.js`** (line 704):
   - Apply the same `const { txHash } = await escrowRefund(...)` pattern used elsewhere
   - If `!txHash`, throw an error or return a 502 status to indicate the blockchain refund failed
   - Add a `refund_tx_hash` column update when refund succeeds (consistent with lines 885, 1021)
   - Consider adding the cancellation status change inside the same conditional block so escrow must succeed before status changes

2. **`backend/api/src/services/escrow.js`**:
   - Improve error granularity: distinguish "not initialized" (env config issue) from "transaction reverted" (on-chain failure) from "network error" (transient)
   - Export a result type that includes `errorCategory` for better caller decision-making

3. **Database / Supabase**:
   - Add a `refund_attempted_at` timestamp to orders table for audit trail
   - Consider adding a background reconciliation job that finds orders where `status = 'cancelled'` but `escrow_status != 'refunded'`

4. **Tests**:
   - Add test case: escrow contract not initialized + cancellation — should return error, not proceed with status change
   - Add test case: blockchain RPC timeout during refund — should retry or return error, not silently proceed

ionfwsrijan:
## Summary

Fixes a critical ordering bug in the cancel order flow where the order status was changed to "cancelled" *before* the blockchain escrow refund was attempted, potentially leaving customer funds locked in the smart contract.

#### `backend/api/src/routes/orderRoutes.js`
- **Restructured cancel order endpoint** into 3 phases:
  1. **Escrow refund** (if funded) — attempted first
  2. **Escrow status update** — `refund_tx_hash` and `escrow_refunded_at` persisted
  3. **Order cancellation** — status changes to `cancelled` only after successful refund
- Returns **HTTP 502** with a descriptive error if the escrow refund fails (contract not initialized, RPC error, revert), leaving the order's original status intact
- Replaced all `console.*` calls with `logger.*` for structured logging

#### `backend/api/src/services/escrow.js`
- Imported `logger` and replaced all `console.log`/`console.warn`/`console.error` calls
- All three functions (`escrowDeposit`, `escrowRelease`, `escrowRefund`) now use structured logger

Closes #630 

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced order cancellation process with improved escrow refund validation and error handling
  * Refund processing now executes before updating order status, preventing potential inconsistent order states
  * Failed refund transactions are immediately reported with appropriate error responses, avoiding partial state updates
  * Added logging for non-funded escrow booking scenarios


ionfwsrijan:

PR #675, opened 2026-06-21
fix(driverRoutes): remove hardcoded default OTP '1234'

### LINKED ISSUE #668: Hardcoded Default OTP `'1234'` Allows Full Driver Account Takeover
## Description

The driver login OTP endpoint has a hardcoded fallback OTP of `'1234'` that is used when the `DRIVER_LOGIN_OTP` environment variable is not set. This allows anyone who knows about this fallback (including the source code being public) to log in as **any** driver account.

## Root Cause

At `driverRoutes.js` line 30:

```javascript
const expectedOtp = (process.env.DRIVER_LOGIN_OTP || '1234').trim();
```

The production intent is that `DRIVER_LOGIN_OTP` should be set via environment variable. However:

1. There is no startup validation that `DRIVER_LOGIN_OTP` is set
2. There is no enforcement mechanism (e.g., crash on missing env var)
3. The hardcoded `'1234'` fallback is publicly visible in the source code
4. The rate limit is only 5 attempts per 15-minute window (from `authLimiter`), but with a 4-digit keyspace (10,000 combinations), brute force is feasible in ~50 hours

### Authentication Flow

```
POST /api/driver/otp/verify
Body: { phone: "...", otp: "..." }
→ Rate limit check (5 req/15min window)
→ If DRIVER_LOGIN_PHONE is set, only that phone number can login
→ OTP comparison: if (otp !== expectedOtp) → 401
→ If OTP matches, returns { verified: true }
→ Driver app navigates to home (no session token issued!)
```

The more critical issue is that even after successful OTP verification, **no session token is issued** (line 46-49). The `DriverSession.driverId` is read from a compile-time environment variable `DRIVER_ID` in `apps/driver/lib/core/driver_session.dart`, meaning the OTP verification is essentially vestigial — it verifies the OTP but doesn't actually authenticate subsequent API calls.

## Impact

- **Complete account takeover**: Anyone can authenticate as any driver by using OTP `1234`
- **No audit trail**: The driver app sends `x-user-id`, `x-user-role`, and `x-user-name` headers with every request (see `apps/customer/lib/services/order_service.dart` lines 18-24), and the backend accepts these as the authentic identity when `BYPASS_AUTH` is set
- **Combined with BYPASS_AUTH headers**: The Flutter apps hardcode these `x-user-*` headers in all API calls, creating a defense-in-depth failure. If a middleware misconfiguration causes `BYPASS_AUTH` mode to activate (even partially), any user can impersonate any other user
- **Data breach**: An attacker can view all driver's trips, earnings, personal documents, and contact information

### 1. Remove the Hardcoded Default

```javascript
// Validate at startup
if (!process.env.DRIVER_LOGIN_OTP) {
    console.error('FATAL: DRIVER_LOGIN_OTP environment variable is required');
    process.exit(1);
}
const expectedOtp = process.env.DRIVER_LOGIN_OTP.trim();
```

### 2. Implement Dynamic OTP Generation

Instead of a static OTP, generate a random 6-digit OTP per login attempt:

```javascript
// When driver requests OTP:
const otp = String(Math.floor(100000 + Math.random() * 900000));
// Store hashed OTP with expiry (5 min) in Redis:
await redis.setEx(`otp:${phone}`, 300, crypto.createHash('sha256').update(otp).digest('hex'));
// Send via SMS (not in-app notification)
```

### 3. Use Constant-Time Comparison

```javascript
const { timingSafeEqual } = require('crypto');

function constantTimeCompare(a, b) {
    const bufA = Buffer.from(String(a));
    const bufB = Buffer.from(String(b));
    if (bufA.length !== bufB.length) return false;
    return timingSafeEqual(bufA, bufB);
}
```

### 4. Issue Proper Session Token After Verification

Replace the `{ verified: true }` response with issuing a proper Firebase custom token or JWT:

```javascript
const firebaseToken = await admin.auth().createCustomToken(driverProfile.user_id);
res.json({ token: firebaseToken, driver: driverProfile });
```

### 5. Remove BYPASS_AUTH Headers from Mobile Clients

Remove `x-user-id`, `x-user-role`, and `x-user-name` from all Flutter API service files. Authentication must use `Authorization: Bearer` exclusively.

## Files That Need Changes

| File | Changes Required |
|------|-----------------|
| `backend/api/src/routes/driverRoutes.js` | Remove hardcoded OTP, add startup validation, implement dynamic OTP generation, issue session token |
| `backend/api/src/middleware/rateLimiter.js` | Tighten OTP verification rate limit (3 per 15 min per phone) |
| `apps/driver/lib/core/driver_session.dart` | Remove compile-time `DRIVER_ID`, use session token |
| `apps/driver/lib/screens/otp_screen.dart` | Handle token response, store securely |
| `apps/customer/lib/services/order_service.dart` | Remove `x-user-*` headers |
| `apps/customer/lib/services/profile_service.dart` | Remove `x-user-*` headers |
| `apps/driver/lib/services/trip_service.dart` | Remove `x-user-*` headers |
| `apps/driver/lib/services/marketplace_repository.dart` | Remove `x-user-*` headers |
| `backend/api/src/index.js` | Remove BYPASS_AUTH support entirely in production |

ionfwsrijan:
## Description

`driverRoutes.js` falls back to `'1234'` when `DRIVER_LOGIN_OTP` env var is unset, meaning any deployment without proper configuration accepts OTP `1234` for every driver login.

## Changes

- Removed `|| '1234'` fallback in the OTP comparison
- Route now returns `503 Service Unavailable` with message `'OTP not configured'` when the env var is missing
- Added startup warning in `index.js` logging `'DRIVER_LOGIN_OTP is not set! Driver OTP login will return 503.'`

## Testing

- Verified endpoint returns 503 when env var is unset
- Verified endpoint works normally when env var is set

Closes #668

## Release Notes

* **Bug Fixes**
  * OTP verification now depends on the configured `DRIVER_LOGIN_OTP`. If it isn’t set, requests to verify OTP will return HTTP 503 instead of using a default fallback.

* **Chores**
  * Added a startup warning when driver OTP login isn’t configured, so missing production settings are easier to spot.

* **Tests**
  * Updated test setup to ensure `DRIVER_LOGIN_OTP` is set during test runs.

### DIFF
--- a/backend/api/src/index.js
+++ b/backend/api/src/index.js
@@ -36,6 +36,9 @@ if (process.env.NODE_ENV === 'production' && process.env.BYPASS_AUTH === 'true')
   logger.fatal('BYPASS_AUTH is enabled in production. This is a severe security misconfiguration. Set BYPASS_AUTH=false (or unset it) and restart the server.');
   process.exit(1);
 }
+if (!process.env.DRIVER_LOGIN_OTP) {
+  logger.warn('DRIVER_LOGIN_OTP is not set. Driver OTP login will be disabled until it is configured in production.');
+}
 const app = express();
 const server = http.createServer(app);
 
--- a/backend/api/src/routes/driverRoutes.js
+++ b/backend/api/src/routes/driverRoutes.js
@@ -27,9 +27,9 @@ const verifyLoginOtpLimiter = rateLimit({
 
 router.post('/otp/verify', verifyLoginOtpLimiter, validateBody(loginOtpSchema), async (req, res) => {
   const { phone, otp } = req.body;
-  const expectedOtp = (process.env.DRIVER_LOGIN_OTP || '1234').trim();
+  const expectedOtp = (process.env.DRIVER_LOGIN_OTP || '').trim();
 
-  if (process.env.NODE_ENV === 'production' && !process.env.DRIVER_LOGIN_OTP) {
+  if (!process.env.DRIVER_LOGIN_OTP) {
     return res.status(503).json({
       error: 'Driver login OTP verification is not configured on this server.',
     });
--- a/backend/api/test/setup.js
+++ b/backend/api/test/setup.js
@@ -13,6 +13,7 @@ process.env.BYPASS_AUTH = 'true';
 process.env.MONGODB_SHUTDOWN_WAIT_MS = '0';
 process.env.ESCROW_MATIC_PER_PAISA = '0.01';
 process.env.MAX_ESCROW_MATIC = '1000';
+process.env.DRIVER_LOGIN_OTP = '1234';
 
 // Suppress noisy console.error output from the routes — they log
 // pricing errors and DB failures to stderr when tests trigger them.


ionfwsrijan:

PR #677, opened 2026-06-21
fix(orderRoutes): add distributed lock to prevent bid acceptance race

### LINKED ISSUE #670: Race Condition in Bid Acceptance: Missing Atomic Lock Allows Double-Accept of Same Load by Two Drivers
## Description

The bid acceptance endpoint at `POST /api/orders/:id/bids/:bidId/accept` has a time-of-check-to-time-of-use (TOCTOU) race condition. Between checking the order status (`'pending'`) and executing the acceptance transaction, another concurrent request can accept a different bid for the same order, resulting in **two drivers being told their bids were accepted** for the same load.

## Root Cause

At `orderRoutes.js` lines 638-728, the bid acceptance flow is:

```javascript
// STEP 1: Check order status (line ~650)
const { data: order } = await supabase
    .from('orders')
    .select('status, ...')
    .eq('id', orderId)
    .single();

if (order.status !== 'pending') {
    return res.status(400).json({ error: 'Order is not pending' });
}

// STEP 2: Check bid exists (line ~670)
const { data: bid } = await supabase
    .from('load_bids')
    .select('*')
    .eq('id', bidId)
    .eq('order_id', orderId)
    .single();

// STEP 3: Accept bid (line ~700)
const { data, error } = await supabase.rpc('accept_bid_tx', {
    p_order_id: orderId,
    p_bid_id: bidId,
    p_driver_id: driverId,
});
```

The problem is that **Steps 1 and 2 are separate SELECT queries** executed outside any transaction or lock. The following scenario is possible:

1. **Process A** (Driver A's accept request): Checks order status → `'pending'` ✓
2. **Process B** (Driver B's accept request): Checks order status → `'pending'` ✓ (same result — no lock held)
3. **Process A**: Executes `accept_bid_tx` → order status changes to `'accepted'`
4. **Process B**: Also executes `accept_bid_tx` → either succeeds (if the RPC doesn't check status atomically) or fails

If the `accept_bid_tx` RPC does NOT atomically verify the order is still `'pending'` before accepting, **both drivers are notified that their bid was accepted**.

Even if the RPC is correct, the **response to Driver B** will be an error, but **in the time between step 2 and step 3**, both drivers were told different things. The frontend may already show "Bid Accepted!" to both.

## Impact

- **Double assignment**: Two drivers believe they have been assigned the same load
- **Customer confusion**: Customer sees two different drivers assigned
- **Reputation system corruption**: Both drivers may start heading to the pickup location
- **Financial loss**: If both drivers complete the delivery, who gets paid?
- **Escrow funds confusion**: The escrow contract may be funded for only one driver, leaving the other unpaid

## Additional Race Conditions in Status Transitions

The same pattern exists in other order status transitions:

1. **Milestone updates** (`PUT /:id/milestones` at line 734): No lock between reading order status and updating milestone
2. **Verify-delivery OTP** (`POST /:id/verify-delivery` at line 806): No lock between reading OTP record and updating delivery status
3. **Cancel order** (`POST /:id/cancel` at line 987): No lock between reading order status and cancelling

### 1. Create an Atomic RPC for Bid Acceptance

Move the entire bid acceptance logic into a single PostgreSQL function that uses `SELECT ... FOR UPDATE` to lock the order row:

```sql
CREATE OR REPLACE FUNCTION accept_bid_atomic(
    p_order_id UUID,
    p_bid_id UUID,
    p_driver_id UUID
) RETURNS JSONB AS $$
DECLARE
    v_order_status TEXT;
    v_result JSONB;
BEGIN
    -- Lock the order row to prevent concurrent modifications
    SELECT status INTO v_order_status 
    FROM orders 
    WHERE id = p_order_id 
    FOR UPDATE;  -- <-- THIS IS THE KEY: row-level lock
    
    IF v_order_status != 'pending' THEN
        RETURN jsonb_build_object(
            'success', false, 
            'error', 'Order is no longer pending. Current status: ' || v_order_status
        );
    END IF;
    
    -- Update order status
    UPDATE orders SET 
        status = 'accepted', 
        driver_id = p_driver_id,
        updated_at = NOW()
    WHERE id = p_order_id;
    
    -- Update bid status
    UPDATE load_bids SET 
        status = 'accepted'
    WHERE id = p_bid_id AND order_id = p_order_id;
    
    -- Reject all other bids
    UPDATE load_bids SET 
        status = 'rejected'
    WHERE order_id = p_order_id 
    AND id != p_bid_id 
    AND status = 'pending';
    
    RETURN jsonb_build_object('success', true);
END;
$$ LANGUAGE plpgsql;
```

### 2. Use PostgreSQL Advisory Lock for Other Transitions

For milestone updates, delivery verification, and cancellations:

```javascript
// Before status transition:
await supabase.rpc('acquire_order_lock', { p_order_id: orderId });
try {
    // Perform status transition
    // ...
} finally {
    await supabase.rpc('release_order_lock', { p_order_id: orderId });
}
```

### 3. Implement Optimistic Concurrency with Version Column

Add a `version` column to the `orders` table:

```sql
ALTER TABLE orders ADD COLUMN version INTEGER DEFAULT 1;
```

Use version-based optimistic locking:

```javascript
const { data: order } = await supabase
    .from('orders')
    .select('status, version')
    .eq('id', orderId)
    .single();

const { error } = await supabase
    .from('orders')
    .update({ 
        status: 'accepted', 
        driver_id: driverId,
        version: order.version + 1 
    })
    .eq('id', orderId)
    .eq('version', order.version);  // Version check ensures no concurrent update

if (error && error.code === 'P0001') {  // Or check affected rows = 0
    return res.status(409).json({ error: 'Concurrent modification detected. Retry.' });
}
```

### 4. Add Integration Tests

Create tests in `test/integration/bids.test.js` that:
- Send 10 simultaneous bid acceptance requests for the same order
- Verify exactly ONE succeeds
- Verify exactly ONE driver is assigned
- Verify all other bids are marked as rejected
- Test the same pattern for milestone updates, delivery verification, and cancellations

## Files That Need Changes

| File | Changes Required |
|------|-----------------|
| `backend/api/src/routes/orderRoutes.js` | Replace TOCTOU pattern with atomic RPC call; add advisory lock for all status transitions |
| `docs/migration_add_atomic_bid_accept.sql` | SQL migration for atomic `accept_bid_atomic()` function and `version` column (new file) |
| `backend/api/test/integration/bids.test.js` | Add concurrency tests for bid acceptance |
| `backend/api/test/integration/orders.test.js` | Add concurrency tests for status transitions |

ionfwsrijan:
## Description

Two customers can accept different bids on the same order between the status check (`'accepted'`) and the `accept_bid_tx` RPC call, resulting in both bids being accepted for the same order.

## Changes

- Added Redis-based distributed lock on key `bid_accept_lock:{orderId}` using `SET NX` with 10-second TTL
- Lock is acquired before the status validation step
- Lock is released with an atomic Lua script that only DELETEs if the key holds the expected value (prevents releasing another process's lock)
- Returns `409 Conflict` with message `'Bid acceptance is already in progress for this order'` when lock cannot be acquired

## Testing

- Verified concurrent bid accept requests: second request gets 409
- Verified lock auto-expires after TTL
- Verified lock is properly released on success and error paths

Closes #670

## Bug Fixes
* Added safeguards to prevent conflicting bid acceptance when multiple requests occur at the same time for the same order. The system now uses short-lived distributed locking to allow only one acceptance to proceed, returns a clear “conflict” response when another acceptance is in progress, and releases the lock safely to maintain data consistency.

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -635,6 +635,18 @@ router.post('/:id/bids/:bidId/accept', authenticate, requireRole(['customer']), 
   const orderId = req.params.id;
   const bidId = req.params.bidId;
 
+  // Acquire a distributed lock on this order to prevent concurrent bid acceptance
+  const lockKey = `bid_accept_lock:${orderId}`;
+  const lockTimeoutMs = 10000;
+  let lockValue = null;
+  if (redisClient) {
+    lockValue = crypto.randomUUID();
+    const acquired = await redisClient.set(lockKey, lockValue, 'PX', lockTimeoutMs, 'NX');
+    if (!acquired) {
+      return res.status(409).json({ error: 'Another bid acceptance is in progress for this order. Please try again.' });
+    }
+  }
+
   try {
     const { data: order } = await supabase.from('orders').select('order_display_id, customer_id').eq('id', orderId).maybeSingle();
     if (!order || order.customer_id !== req.user.id) return res.status(403).json({ error: 'Access Denied: You do not own this order.' });
@@ -748,6 +760,21 @@ router.post('/:id/bids/:bidId/accept', authenticate, requireRole(['customer']), 
   } catch (err) {
     logger.error({ err }, '[orderRoutes] accept bid error');
     res.status(500).json({ error: 'Internal Server Error' });
+  } finally {
+    if (redisClient && lockValue) {
+      const luaScript = `
+        if redis.call('GET', KEYS[1]) == ARGV[1] then
+          redis.call('DEL', KEYS[1])
+          return 1
+        end
+        return 0
+      `;
+      try {
+        await redisClient.eval(luaScript, 1, lockKey, lockValue);
+      } catch {
+        // Lock expiry will handle cleanup if the DEL fails
+      }
+    }
   }
 });
 


ionfwsrijan:
@KanishJebaMathewM   Please review this

PR #728, opened 2026-06-22
[Fix]: import statement and update ticket handling logic

### LINKED ISSUE #700: [Bug]: Non-admin users can close their own support tickets — status guard logic is inverted
backend/api/src/routes/supportRoutes.js
Severity: High
Label: bug

### Description

The PATCH /api/support/tickets/:id endpoint is supposed to restrict status changes so that only admins can move a ticket to privileged statuses. The guard condition is logically inverted: it blocks non-admins from changing to any status except "closed", when it should be blocking them from changing to "closed" (and other admin-only statuses).

As a result, any authenticated user can close their own support ticket by sending status: "closed" in the request body.

### Steps to Reproduce

Authenticate as a regular customer or driver (non-admin).
Create a support ticket — it starts with status: "open".
Send PATCH /api/support/tickets/:id with body { "status": "closed" }.
Response: 200 OK — ticket is now status: "closed".
The ticket cannot be updated further (blocked by the ticket.status === 'closed' early guard), so no agent can action it.

### Root Cause
In backend/api/src/routes/supportRoutes.js:

jsif (
  req.user.role !== 'admin' &&
  normalizedStatus !== ticket.status &&
  normalizedStatus !== 'closed'   // ❌ This allows 'closed', blocks everything else
) {
  return res.status(403).json({
    error: 'Access Denied: Only admins can change tickets to this status.',
  });
}

The intent is: "non-admins cannot set admin-only statuses."
The actual behaviour is: "non-admins are blocked from any status change UNLESS the target is 'closed'."

The !== 'closed' condition makes the 403 guard skip entirely when normalizedStatus is "closed", letting the update through.

### Expected Behavior
Non-admin users should only be able to set their ticket status to "resolved" (or keep it unchanged). Only admins can set "closed", "in_progress", or other privileged statuses.

### Actual Behavior
Non-admin users can set their ticket to "closed", effectively removing it from the support queue. The guard blocks all other status transitions for non-admins.

ishitaajain22-tech:
**File changed:** `backend/api/src/routes/supportRoutes.js`

The guard condition `normalizedStatus !== 'closed'` made the 403 skip entirely when the target status was `"closed"`, so non-admin users could close their own tickets while being blocked from all other transitions. Fixed by replacing the negation with an explicit `USER_ALLOWED_STATUSES = ['resolved']` allowlist check.

## Summary by CodeRabbit

fixes #700

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened ticket status change permissions: non-admin users can now update a ticket’s status only to **resolved**; attempts to change to other statuses are blocked with a **403** and an updated error message.


ishitaajain22-tech:
@KanishJebaMathewM pls assign me

ishitaajain22-tech:
GSSOC

PR #734, opened 2026-06-22
fix: prevent silent telemetry data loss with recovery file, proper error handling, and drop metrics

### LINKED ISSUE #731: Silent Telemetry Data Loss in GPS Tracking Pipeline Due to In-Memory Buffer Flaws
### Description

The GPS telemetry pipeline in `tracker.js` silently loses tracking data at multiple stages with no recovery mechanism. Since telemetry is the foundation for route optimization, driver payments, dispute resolution, and ML model training, this data loss has cascading operational impacts.

### Four Distinct Data Loss Vectors

**1. Buffer Overflow Silent Drop (lines 430-434)**
When the in-memory `telemetryWriteBuffer` reaches 5000 entries, the oldest 500 records (`Math.floor(5000 * 0.1)`) are spliced and discarded:
```js
if (telemetryWriteBuffer.length >= MAX_BUFFER_SIZE) {
    const dropCount = Math.floor(MAX_BUFFER_SIZE * 0.1);
    telemetryWriteBuffer.splice(0, dropCount);
}
```
No structured alert (Sentry, metrics) is raised — only a `logger.warn`. In production with thousands of drivers sending pings every 2-3 seconds, buffer overflow is a normal condition during MongoDB latency spikes, not an exception.

**2. Retry Buffer Corruption (lines 567-573)**
On flush failure, the code reconstructs the buffer:
```js
const spaceAvailable = Math.max(0, MAX_BUFFER_SIZE - telemetryWriteBuffer.length);
const recordsToKeep = recordsToFlush.slice(-spaceAvailable);
telemetryWriteBuffer = [...recordsToKeep, ...telemetryWriteBuffer];
```
Between the `splice` and the `push`, the buffer could have grown (new pings arrived). The `spaceAvailable` calculation is stale the moment it's computed. Records pushed during this window are silently dropped — they're neither in `recordsToKeep` nor in the current buffer.

**3. Partial Insert Failure Blindness (line 554)**
```js
await collection.insertMany(recordsToFlush, { ordered: false });
```
When `ordered: false` is used and some documents fail validation, `insertMany` throws a `BulkWriteError` that contains per-document failure information. The catch block (line 557) treats this as a total failure and retries ALL records — even the successfully inserted ones — causing duplicates. The individual document errors are never logged, making it impossible to identify malformed telemetry.

**4. Shutdown Data Loss (lines 636-648)**
On graceful shutdown, if MongoDB is unavailable after `MONGODB_SHUTDOWN_WAIT_MS` (default 10s), ALL records still in the buffer are silently lost:
```js
if (!getMongoDb()) {
    const dataLoss = telemetryWriteBuffer.length;
    logger.warn(`${dataLoss} telemetry records will be lost.`);
}
```
A `warn` log is the only record. No file-based recovery, no alternative sink, no structured alerting.

### Affected Files

| File | Lines | Issue |
|------|-------|-------|
| `backend/api/src/sockets/tracker.js` | 430-434 | Buffer overflow silent drop |
| `backend/api/src/sockets/tracker.js` | 546-580 | Retry buffer corruption & partial failure blindness |
| `backend/api/src/sockets/tracker.js` | 636-648 | Shutdown data loss |
| `backend/api/src/config/db.js` | - | No WAL/recovery config |
| `backend/api/src/middleware/logger.js` | - | No structured alerting integration |

### Impact

- **Inaccurate billing**: Driver payments based on distance/route cannot be verified
- **Broken dispute resolution**: GPS evidence for delivery disputes is incomplete
- **Degraded ML models**: Price/demand prediction models train on partial data
- **Lost trust**: Customers see stale driver locations during order tracking
- **No observability**: Ops teams have no metrics on data loss rates

### Suggested Fix

**Immediate fixes (high priority):**
1. Add file-based write-ahead log (WAL) that writes telemetry to local disk before buffering — recover on restart
2. Replace the silent `splice` with structured alerting via Sentry capture and a Prometheus counter metric
3. In the retry handler, deduplicate: do not re-push records that were already inserted

**Short-term fixes:**
4. Replace `insertMany` with per-document `bulkWrite` using `replaceOne` filters on unique telemetry IDs to prevent duplicates on retry
5. On shutdown, write remaining buffer to a recovery file in `os.tmpdir()` instead of dropping

**Medium-term fixes:**
6. Add a Redis-backed secondary buffer as intermediary cache between memory and MongoDB
7. Add a health check endpoint that exposes buffer fill percentage, drop counts, and last flush timestamp

ionfwsrijan:
## Description
Fixed four distinct data loss vectors in the GPS telemetry pipeline that silently
lost tracking data with no recovery mechanism.

### Changes
- **Buffer overflow drop** (tracker.js:405-408): Added `telemetryDropCounter`
  metric and cumulative drop logging for observability.
- **Retry buffer corruption** (tracker.js:496-530): Replaced stale `spaceAvailable`
  calculation with atomic buffer swap pattern. `BulkWriteError` is now handled by
  extracting `writeErrors` and retrying only the successful records instead of
  discarding the entire batch.
- **Partial insert failure** (tracker.js:504): Validation errors now log individual
  document failures (up to 5 samples) for debugging malformed telemetry.
- **Shutdown data loss** (tracker.js:595-598): When MongoDB is unavailable during
  shutdown, remaining buffer is written to a JSONL recovery file in `os.tmpdir()`.
  On next startup, `loadRecoveryFile()` restores previously buffered records.

### Impact
- No more silent data loss during MongoDB latency spikes
- Malformed telemetry records are identifiable (no more blind discards)
- Server restart does not lose buffered telemetry data
- Ops teams can monitor cumulative drop count

Fixes #731

## Summary by CodeRabbit

* **New Features**
  * Added durable telemetry recovery that saves unsent telemetry locally and automatically restores it on startup.
* **Bug Fixes**
  * Improved telemetry buffering so dropped-record warnings now include a cumulative dropped count.
  * Enhanced telemetry database write handling by recovering from validation-style write failures while preserving successfully written records.
  * During shutdown, buffered telemetry is persisted to recovery storage when the database is unavailable to prevent data loss.
* **Tests**
  * Updated tracker shutdown and buffer-drop expectation strings to match revised warning messages.

### DIFF
--- a/backend/api/src/sockets/tracker.js
+++ b/backend/api/src/sockets/tracker.js
@@ -2,9 +2,15 @@ import { WebSocketServer } from 'ws';
 import { mongoDb, redisClient, firebaseAdmin, supabase } from '../config/db.js';
 import jwt from 'jsonwebtoken';
 import logger from '../middleware/logger.js';
+import os from 'os';
+import path from 'path';
+import fs from 'fs';
 
 let mongoDbOverride = null;
 const getMongoDb = () => mongoDbOverride || mongoDb;
+
+let telemetryDropCounter = 0;
+const RECOVERY_FILE_PATH = path.join(os.tmpdir(), 'truxify-telemetry-recovery.jsonl');
 
 // In-memory mapping of active client subscriptions
 const trackingSubscriptions = new Map();
@@ -405,7 +411,8 @@ export async function handleLocationPing(ws, data) {
   if (telemetryWriteBuffer.length >= MAX_BUFFER_SIZE) {
     const dropCount = Math.floor(MAX_BUFFER_SIZE * 0.1);
     telemetryWriteBuffer.splice(0, dropCount);
-    logger.warn(`[TRUXIFY BUFFER WARN] Telemetry buffer full: dropped ${dropCount} oldest records. Size: ${telemetryWriteBuffer.length}/${MAX_BUFFER_SIZE}`);
+    telemetryDropCounter += dropCount;
+    logger.warn(`[TRUXIFY BUFFER WARN] Telemetry buffer full: dropped ${dropCount} oldest records. Total dropped: ${telemetryDropCounter}. Size: ${telemetryWriteBuffer.length}/${MAX_BUFFER_SIZE}`);
   }
   telemetryWriteBuffer.push({
     driver_id,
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

PR #806, opened 2026-06-23
fix: prevent race condition in complete_trip_tx allowing payout for cancelled orders

### LINKED ISSUE #803: Race Condition in `complete_trip_tx` Allows Driver Payout for Concurrently Cancelled Orders
## Description

The `complete_trip_tx(p_order_id UUID)` PostgreSQL RPC function — called by the `verify-delivery` endpoint — does **not** use `SELECT ... FOR UPDATE` to lock the order row, and only checks for `payment_released` status, not `cancelled`. When a cancellation request and a delivery verification request execute concurrently, a driver can receive full wallet credit and on-chain escrow release for an order that was cancelled at nearly the same time.

### In `docs/supabase_setup.sql` (lines 1511–1600):

The `complete_trip_tx(p_order_id UUID)` function:

1. **Line 1521: Selects the order without `FOR UPDATE`:**
   ```sql
   select * into v_order from orders where id = p_order_id;
   -- No "FOR UPDATE" — no row lock is acquired
   ```

2. **Lines 1532–1534: Only checks for `payment_released`, not `cancelled` or `delivered`:**
   ```sql
   if v_order.status = 'payment_released' then
     return;
   end if;
   ```

3. **Lines 1571–1576: Unconditionally overwrites the status to `payment_released`:**
   ```sql
   update orders
   set otp_verified = true,
       status = 'payment_released',
       updated_at = now()
   where id = p_order_id;
   ```
   No WHERE guard against `cancelled` status.

4. **Lines 1584–1591: Driver wallet is unconditionally credited:**
   ```sql
   update driver_details
   set
     total_trips = total_trips + 1,
     wallet_confirmed = wallet_confirmed + v_order.total_amount,
     wallet_total = wallet_total + v_order.total_amount,
     updated_at = now()
   where user_id = v_order.driver_id;
   ```

### The cancel endpoint in `orderRoutes.js` (lines 1091–1096):

```javascript
const { data: updatedOrder, error: updateErr } = await supabase.from('orders')
  .update(updatePayload)
  .eq('order_display_id', orderId)
  .not('status', 'in', '("delivered","payment_released","cancelled")')
  .select('...')
  .single();
```

The cancel endpoint uses an atomic `UPDATE ... WHERE status NOT IN (...)` guard, which is correct. However, the `verify-delivery` path at line 881–888 also has an `UPDATE` with `.not('status', 'eq', 'cancelled')` and `.not('status', 'eq', 'payment_released')`, but this is just a "touch" update (only sets `updated_at`) — the **actual state-changing operation** is `complete_trip_tx` RPC at line 898, which bypasses this guard entirely.

### Race Window Timeline

```
Time  |  cancel request                    |  verify-delivery request
------|------------------------------------|----------------------------------
 t1   | SELECT order (status='pending')    | SELECT order (status='pending')
      |  (no lock in verify-delivery)      |  (no lock in complete_trip_tx)
 t2   |                                    | OTP verified successfully
 t3   |                                    | complete_trip_tx begins:
      |                                    |   SELECT orders WHERE id=X (no FOR UPDATE)
      |                                    |   -> v_order.status = 'pending'
      |                                    |   -> NOT (status='payment_released')
      |                                    |   -> Proceeds
 t4   | UPDATE orders SET status='cancelled'|
      | WHERE id=X AND status NOT IN        |
      |   ('delivered','payment_released',  |
      |    'cancelled')                     |
      | -> SUCCESS (atomic UPDATE guard)    |
 t5   |                                    | UPDATE orders SET
      |                                    |   status='payment_released', ...
      |                                    |   (unconditionally overwrites cancelled!)
      |                                    | UPDATE driver_details
      |                                    |   wallet_confirmed += total_amount
      |                                    | INSERT wallet_transactions credit
      |                                    | INSERT earnings_daily
 t6   |                                    | Escrow releaseFunds() called on-chain
```

## Impact (Critical — Financial Loss)

1. **Driver is paid for a cancelled order** — the wallet is credited with `total_amount` even though the order was cancelled moments earlier.
2. **On-chain escrow funds are released** — `releaseFunds()` is called on the blockchain, transferring cryptocurrency to the driver that should have been refunded to the customer.
3. **No recovery mechanism** — once the wallet is credited and the blockchain transaction is mined, reversing is extremely difficult.
4. **Customer loses funds** — the cancellation fee logic (which would deduct a fee and refund the rest) is bypassed.

## The Fix

**Fix the `complete_trip_tx(p_order_id UUID)` function in `supabase_setup.sql`:**

```sql
create or replace function complete_trip_tx(p_order_id uuid)
returns void
language plpgsql
security definer
as $$
declare
  v_order record;
  v_trip_display_id text;
  v_active_trip_count int;
begin
  -- Add FOR UPDATE to lock the order row and prevent concurrent modifications
  select * into v_order from orders where id = p_order_id for update;

  if not found then
    raise exception 'Order not found';
  end if;

  if v_order.driver_id is null then
    raise exception 'No driver assigned to this order';
  end if;

  -- Check for both payment_released AND cancelled statuses
  if v_order.status = 'payment_released' then
    return;
  end if;

  -- NEW: Check if the order was cancelled
  if v_order.status = 'cancelled' then
    raise exception 'Order has been cancelled — cannot complete trip';
  end if;

  -- Also check for delivered to prevent double-processing
  if v_order.status = 'delivered' then
    raise exception 'Order has already been delivered';
  end if;

  -- ... rest of the function remains the same, but:
  -- At the final UPDATE, add a WHERE guard:
  update orders
  set otp_verified = true,
      status = 'payment_released',
      updated_at = now()
  where id = p_order_id
    and status != 'cancelled'       -- NEW: defensive guard
    and status != 'payment_released'; -- NEW: defensive guard

  -- Check if the update actually affected any row
  if not found then
    raise exception 'Order status changed during processing — possible concurrent cancellation';
  end if;

  -- ... remaining logic (wallet credit, etc.) ...
end;
$$;
```

**Additionally, in `orderRoutes.js` (verify-delivery handler, around line 898):**

Add a post-RPC status verification before proceeding with escrow release:

```javascript
// After complete_trip_tx succeeds, verify the order was actually updated
const { data: verifiedOrder } = await supabase
  .from('orders')
  .select('status')
  .eq('id', orderId)
  .single();

if (!verifiedOrder || verifiedOrder.status !== 'payment_released') {
  return res.status(409).json({
    error: 'Order status changed during processing. Payment was not released.',
  });
}
```

ionfwsrijan:
## Description

The `complete_trip_tx(p_order_id UUID)` PostgreSQL RPC function did **not** use `SELECT ... FOR UPDATE` to lock the order row, and only checked for `payment_released` status, not `cancelled`. When a cancellation request and a delivery verification request executed concurrently, a driver could receive full wallet credit and on-chain escrow release for an order that was cancelled at nearly the same time.

### `docs/supabase_setup.sql` — `complete_trip_tx` function:

1. **`FOR UPDATE` row lock**: Changed `SELECT * INTO v_order FROM orders WHERE id = p_order_id` to `SELECT ... FOR UPDATE`, acquiring an exclusive row-level lock that prevents concurrent modifications.

2. **Cancelled status check**: Added an explicit check: `IF v_order.status = 'cancelled' THEN RAISE EXCEPTION 'Order has been cancelled — cannot complete trip';`

3. **Delivered status check**: Added a check to prevent double-processing: `IF v_order.status = 'delivered' THEN RAISE EXCEPTION 'Order has already been delivered';`

4. **Defensive WHERE guards**: The final `UPDATE orders SET status = 'payment_released'` now includes `AND status != 'cancelled' AND status != 'payment_released'`.

5. **Zero-row-update detection**: Added `GET DIAGNOSTICS v_updated_count = ROW_COUNT` after the UPDATE, raising an exception if no row was affected.

### `backend/api/src/routes/orderRoutes.js` — verify-delivery handler:

6. **Post-RPC verification**: Added a status check after `complete_trip_tx` succeeds, verifying the order is actually in `payment_released` state before proceeding with escrow release. Returns HTTP 409 if the status doesn't match.

## Security

- **Race window eliminated**: `FOR UPDATE` ensures the cancel and complete operations are serialized
- **Defense in depth**: Even without the lock, WHERE guards and post-RPC verification provide multiple layers of protection
- **Drivers cannot be paid for cancelled orders**: Wallet credit, escrow release, and earnings tracking are all blocked

Closes #803

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of payment release during delivery verification by adding validation checks to ensure orders reach the correct status. Enhanced error detection for edge cases involving concurrent order modifications.

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -914,6 +914,25 @@ router.post('/:id/verify-delivery', authenticate, userLimiter, requireRole(['dri
       return res.status(500).json({ error: 'Failed to complete trip and release payment.', details: rpcErr.message });
     }
 
+    // Post-RPC verification: confirm the order was actually updated to payment_released
+    const { data: verifiedOrder, error: verifyErr } = await supabase
+      .from('orders')
+      .select('status')
+      .eq('id', orderId)
+      .maybeSingle();
+
+    if (verifyErr || !verifiedOrder) {
+      logger.error(`[verify-delivery] Failed to verify order status after RPC for order ${orderId}`);
+      return res.status(500).json({ error: 'Failed to verify order status after payment release.' });
+    }
+
+    if (verifiedOrder.status !== 'payment_released') {
+      logger.warn(`[verify-delivery] Order ${orderId} status changed to "${verifiedOrder.status}" — payment was not released.`);
+      return res.status(409).json({
+        error: 'Order status changed during processing. Payment was not released.',
+      });
+    }
+
     // OTP is only consumed after the RPC succeeds — if the RPC fails the driver can retry
     await verifyDeliveryOtp(orderId);
     await clearOtpState(orderId);
--- a/docs/supabase_setup.sql
+++ b/docs/supabase_setup.sql
@@ -1521,8 +1521,10 @@ declare
   v_order record;
   v_trip_display_id text;
   v_active_trip_count int;
+  v_updated_count int;
 begin
-  select * into v_order from orders where id = p_order_id;
+  -- Use FOR UPDATE to lock the order row and prevent concurrent modifications
+  select * into v_order from orders where id = p_order_id for update;
 
   if not found then
     raise exception 'Order not found';
@@ -1535,6 +1537,16 @@ begin
   -- Idempotency guard: check if the order status is already payment_released
   if v_order.status = 'payment_released' then
     return;
+  end if;
+
+  -- Check if the order was cancelled
+  if v_order.status = 'cancelled' then
+    raise exception 'Order has been cancelled — cannot complete trip';
+  end if;
+
+  -- Check if the order was already delivered — prevents double-processing
+  if v_order.status = 'delivered' then
+    raise exception 'Order has already been delivered';
   end if;
 
   -- Safe lookup for the driver's active trip
@@ -1572,12 +1584,20 @@ begin
     where trip_display_id = v_trip_display_id;
   end if;
 
-  -- Update order status to payment_released
+  -- Update order status to payment_released with defensive WHERE guards
   update orders
   set otp_verified = true,
       status = 'payment_released',
       updated_at = now()
-  where id = p_order_id;
+  where id = p_order_id
+    and status != 'cancelled'
+    and status != 'payment_released';
+
+  -- Verify the update actually affected a row
+  get diagnostics v_updated_count = row_count;
+  if v_updated_count = 0 then
+    raise exception 'Order status changed during processing — possible concurrent cancellation';
+  end if;
 
   -- Update order timeline milestone 'Delivered'
   update order_timeline


ionfwsrijan:
@KanishJebaMathewM Please review this. The two failing checks are pre-existing.

KanishJebaMathewM:
resolve unrelated commits @ionfwsrijan

ionfwsrijan:

PR #835, opened 2026-06-24
[Bug]: Driver polygon wallet written to profiles table but escrow read from driver_details — wallet never propagates

### LINKED ISSUE #736: [Bug]: Driver polygon wallet written to profiles table but escrow read from driver_details — wallet never propagates
## Description

`PUT /api/profile/wallet` writes the driver Polygon wallet address to the `profiles.polygon_wallet_address` column, but the bid acceptance and escrow flows read the driver wallet from `driver_details.polygon_wallet_address`. These are two separate tables with no sync mechanism. A driver who sets their wallet via the standard API endpoint will never have it visible to the escrow system.

## Impact

Every escrow-enabled bid acceptance fails with "Both customer and driver must connect a wallet before escrow can be initiated" even when both parties have set their wallet through the API. This blocks the entire Polygon escrow payment flow for all drivers.

## Code Evidence

**Write path —** `backend/api/src/routes/profileRoutes.js:93-99`:
```javascript
const { error: updateErr } = await supabase
  .from(profiles)
  .update({
    wallet_address: normalized,
    polygon_wallet_address: normalized,
  })
  .eq(id, userId);
```
Wallet is stored only in `profiles.polygon_wallet_address`.

**Read path —** `backend/api/src/routes/orderRoutes.js:651-653`:
```javascript
supabase.from(driver_details).select(polygon_wallet_address)
  .eq(user_id, bid.driver_id).maybeSingle(),
```
Driver wallet is read from `driver_details.polygon_wallet_address`.

**Escrow read path —** `backend/api/src/services/escrow.js` also queries `driver_details` for the wallet address during buildDepositTx.

## Fix

Either write `polygon_wallet_address` to `driver_details` in the wallet update handler, or change the bid acceptance and escrow reads to query `profiles` instead of `driver_details`.

vipul674:
## Fix

Added an upsert into `driver_details` after writing to `profiles` in the wallet update handler so both tables stay in sync.

### Changes

**backend/api/src/routes/profileRoutes.js** (lines 108-114)
- After updating `profiles.polygon_wallet_address`, upsert the same address into `driver_details.polygon_wallet_address`
- Uses Supabase upsert with `onConflict: user_id` so it works for both existing and new driver detail rows
- Read paths in `orderRoutes.js` and `escrow.js` already query `driver_details` — they will now see the wallet

Closes #736

## Summary by CodeRabbit

* **Bug Fixes**
  * Wallet updates now also stay in sync with driver details, helping keep account information consistent across the app.
  * If wallet syncing fails, the app now returns a clearer error instead of silently continuing, making issues easier to detect.

### DIFF
--- a/backend/api/src/routes/profileRoutes.js
+++ b/backend/api/src/routes/profileRoutes.js
@@ -106,6 +106,14 @@ router.put('/wallet', authenticate, userLimiter, async (req, res) => {
       return res.status(500).json({ error: 'Failed to update wallet address.', details: updateErr.message });
     }
 
+    const { error: driverDetailsErr } = await supabase
+      .from('driver_details')
+      .upsert({ user_id: userId, polygon_wallet_address: normalized }, { onConflict: 'user_id' });
+
+    if (driverDetailsErr) {
+      return res.status(500).json({ error: 'Failed to sync wallet to driver details.', details: driverDetailsErr.message });
+    }
+
     if (req.user && req.user.uid) {
       void invalidateCachedProfile(req.user.uid);
     }


vipul674:
@KanishJebaMathewM I have submitted the fix. Please let me know if any changes are needed!

PR #841, opened 2026-06-24
[Bug]: Order history query selects driver_name column that does not exist in orders table

### LINKED ISSUE #743: [Bug]: Order history query selects driver_name column that does not exist in orders table
## Description

The order history endpoint (`GET /api/orders/history`) selects `driver_name` from the `orders` table. The `orders` table has no `driver_name` column — driver names are stored in the `profiles` table and resolved at query time. This causes a Supabase query error.

## Impact

Every customer requesting their order history at `/api/orders/history` receives a 500 error. The order history page in the customer app is completely broken.

## Code Evidence

`backend/api/src/routes/orderRoutes.js:356-358`:
```javascript
const { data: history, error } = await supabase
  .from("orders")
  .select("id, order_display_id, status, pickup_address, drop_address, pickup_date, total_amount, goods_type, driver_name, eta, created_at")
  .eq("customer_id", req.user.id)
  .order("created_at", { ascending: false });
```

The `orders` table columns can be verified by looking at the order insertion at lines 216-231 — no `driver_name` field is inserted. The `driver_name` is not a database column; it is populated at runtime for order detail responses (line 306: `o.driver_name = driverMap[o.driver_id]`), but that runtime enrichment happens only in the `/my/active` handler, not in `/history`.

## Fix

Replace `"driver_name"` in the select with `"driver_id"` in the history query and join/resolve the driver name from the profiles table client-side, similar to how `/my/active` does it (lines 302-306).

vipul674:
## Fix

Replaced the nonexistent `driver_name` column in the history query select with `driver_id`, then resolved driver names from the `profiles` table post-query (same pattern as `/my/active` endpoint at line 302-306).

### Changes

**backend/api/src/routes/orderRoutes.js** (lines 358, 364-369)
- Changed `driver_name` → `driver_id` in the Supabase select
- Added profile lookup for all non-null `driver_id` values
- Attached `driver_name` dynamically after the query

Closes #743

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved order history results so driver names are shown reliably for each past order.
  * Updated history responses to fill in missing driver names using associated profile information when needed.

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -355,11 +355,19 @@ router.get('/history', authenticate, userLimiter, requireRole(['customer']), asy
   try {
     const { data: history, error } = await supabase
       .from('orders')
-      .select('id, order_display_id, status, pickup_address, drop_address, pickup_date, total_amount, goods_type, driver_name, eta, created_at')
+      .select('id, order_display_id, status, pickup_address, drop_address, pickup_date, total_amount, goods_type, driver_id, eta, created_at')
       .eq('customer_id', req.user.id)
       .order('created_at', { ascending: false });
 
     if (error) return res.status(500).json({ error: 'Failed to fetch history.', details: error.message });
+
+    const driverIds = [...new Set((history || []).filter(o => o.driver_id).map(o => o.driver_id))];
+    if (driverIds.length > 0) {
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM I have submitted the fix. Please let me know if any changes are needed!

Issue #984, opened 2026-06-27
bug: home screen index out of bounds crash when mock list lengths differ


vipul674:
## Description
`apps/customer/lib/screens/home_screen.dart` lines 139-150:
```dart
itemCount: mockActiveShipments.length,
itemBuilder: (context, index) {
  final shipment = mockActiveShipments[index];
  return ShipmentCard(
    shipment: shipment,
    onTap: () {
      Navigator.of(context).push(
        AppPageRoute(builder: (_) => LiveTrackingScreen(orderId: mockActiveOrders[index].orderId)),
      );
    },
  );
},
```

The `itemCount` uses `mockActiveShipments.length` but the `onTap` callback indexes `mockActiveOrders[index]`. If these two mock lists have different lengths (e.g., 4 shipments but 3 orders), tapping the 4th shipment card throws `IndexOutOfBounds` and crashes the app.

This is currently a latent crash bug because both lists are defined in `mock_data.dart` and happen to have the same length. Any future edit that changes one list without updating the other will trigger production crashes.

## Impact
- HIGH — crash on user interaction if list lengths diverge
- Latent bug (not immediately triggered but structurally fragile)
- Affects home screen shipment card tap navigation

## Expected Fix
Use consistent list for both count and data:
```dart
itemCount: mockActiveOrders.length,
// or use the same source for both
```

Or add a guard check:
```dart
onTap: () {
  if (index < mockActiveOrders.length) {
    Navigator.of(context).push(
      AppPageRoute(builder: (_) => LiveTrackingScreen(orderId: mockActiveOrders[index].orderId)),
    );
  }
},
```

## Location
`apps/customer/lib/screens/home_screen.dart`, lines 139-150.

vipul674:
@KanishJebaMathewM Please assign this issue to me. I'd like to work on it under GSSoC 2026.

vipul674:
GSSOC

Issue #983, opened 2026-06-27
bug: home screen location badge shows hardcoded 'Surat, Gujarat' when online instead of actual user location


vipul674:
## Description
`apps/customer/lib/screens/home_screen.dart` line 107:
```dart
Text(_isOffline ? _locationLabel : 'Surat, Gujarat', ...),
```

When the device is online, the location badge always shows 'Surat, Gujarat' regardless of the user's actual location. The `_locationLabel` variable (which holds the dynamic cached location) is only displayed when the device is offline. The online branch ignores it entirely and uses a hardcoded string.

## Impact
- MEDIUM — every user sees 'Surat, Gujarat' in the location badge when online
- Users in other cities get incorrect location information
- Reduces trust in the app's location accuracy
- Related to issue (hardcoded GPS coordinates) but is a separate display bug

## Steps to Reproduce
1. Open customer app with network connectivity
2. Observe the location badge at the top of the home screen
3. It always shows 'Surat, Gujarat' regardless of where you are

## Expected Fix
Show dynamic location label (from GPS/cache) regardless of online status:
```dart
Text(_locationLabel, ...),
```

Or at minimum, use a more generic label when online:
```dart
Text(_isOffline ? _locationLabel : 'Connected', ...),
```

## Location
`apps/customer/lib/screens/home_screen.dart`, line 107.

vipul674:
@KanishJebaMathewM Please assign this issue to me. I'd like to work on it under GSSoC 2026.

vipul674:
GSSOC

Issue #982, opened 2026-06-27
bug: home screen caches hardcoded Surat GPS coordinates on every launch, overriding real location


vipul674:
## Description
`apps/customer/lib/screens/home_screen.dart` line 40 calls:
```dart
await _cacheManager.cacheLastLocation(21.1702, 72.8311);
```

This hardcodes GPS coordinates for Surat, Gujarat, India and caches them as the user's "last location" on every app launch. This happens regardless of the user's actual physical location. The cached location is then used as `_locationLabel` when the user is offline (line 47), and affects any feature that reads the cached last location.

## Impact
- HIGH — overrides the user's real GPS location with hardcoded Surat coordinates
- Location-dependent features (nearby trucks, pickup suggestions) use wrong coordinates
- Even if the user is in Mumbai, Delhi, or Bangalore, the cached location always says Surat
- The real cached location from any previous session is destroyed on every app launch

## Steps to Reproduce
1. Open the app in any city (e.g., Mumbai)
2. Go offline or check cached location
3. Location always shows Surat (21.1702, 72.8311)

## Expected Fix
Remove the hardcoded `cacheLastLocation` call. Cached location should only be updated with real GPS coordinates when available.

Or at minimum, only cache when no cached location exists yet:
```dart
final existing = await _cacheManager.getLastLocation();
if (existing == null) {
  await _cacheManager.cacheLastLocation(21.1702, 72.8311);
}
```

## Location
`apps/customer/lib/screens/home_screen.dart`, line 40.

vipul674:
@KanishJebaMathewM Please assign this issue to me. I'd like to work on it under GSSoC 2026.

vipul674:
GSSOC

Issue #981, opened 2026-06-27
bug: documents screen uses hardcoded seed UUID as fallback when user is unauthenticated


vipul674:
## Description
`apps/customer/lib/screens/my_documents_screen.dart` line 50:
```dart
final userId = user?.id ?? 'b1111111-1111-1111-1111-111111111111';
```

When `currentUser` is null (unauthenticated state), the screen falls back to a hardcoded seed UUID `b1111111-1111-1111-1111-111111111111`. This UUID likely belongs to a test/seed user, so an unauthenticated user would see another user's documents.

Additionally, `_isOffline` is a one-time check at fetch time (line 30-32). If network state changes while the user is on the page, the offline indicator (line 441-443) won't update reactively.

## Impact
- HIGH — potential data leak if seed UUID has real data
- Unauthenticated users see someone else's documents
- Privacy concern: exposes another user's uploaded documents

## Steps to Reproduce
1. Clear app auth state
2. Open documents screen
3. Screen fetches documents for the hardcoded seed UUID instead of showing empty/error state

## Expected Fix
Show empty state or error when user is not authenticated instead of falling back to a hardcoded UUID.

## Location
`apps/customer/lib/screens/my_documents_screen.dart`, line 50.

vipul674:
@KanishJebaMathewM Please assign this issue to me. I'd like to work on it under GSSoC 2026.

vipul674:
GSSOC

PR #992, opened 2026-06-27
fix: resolve 5 Flutter bugs — pickup date, documents uuid, home screen location & bounds

vipul674:
Batch 3 — fixes 5 Flutter bugs across 4 files.

### #979 — Order creation always sends DateTime.now() as pickup date
Added `pickupDate` parameter to `createOrder()` so caller can pass the user-selected date.

### #981 — Documents screen uses hardcoded seed UUID as fallback
When user is null, shows auth-required error instead of falling back to a hardcoded seed UUID that could leak data.

### #982 — Home screen caches hardcoded Surat GPS on every launch
Changed to only cache default Surat coordinates when no cached location exists, preserving the user's real GPS location.

### #983 — Home screen location badge shows hardcoded 'Surat, Gujarat' when online
Always displays `_locationLabel` (dynamic from GPS/cache) instead of hardcoded string when online.

### #984 — Home screen index out of bounds crash
Added bounds guard (`if (index < mockActiveOrders.length)`) to prevent crash when list lengths differ.

### DIFF
--- a/apps/customer/lib/screens/booking_confirmation_screen.dart
+++ b/apps/customer/lib/screens/booking_confirmation_screen.dart
@@ -118,6 +118,7 @@ class _BookingConfirmationScreenState extends State<BookingConfirmationScreen>
         dropLat: finalDropLat,
         dropLng: finalDropLng,
         pickupTime: widget.draft.dateLabel,
+        pickupDate: DateTime.now(),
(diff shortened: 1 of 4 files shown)


vipul674:

vipul674:
Replacing with individual per-issue PRs.

PR #993, opened 2026-06-27
fix: await cache invalidation in profile routes to prevent stale data

### LINKED ISSUE #956: Redis Cache Invalidation is Fire-and-Forget Allowing Stale Profile Data to Persist
## Description

The `profileRoutes.js` route handler calls `invalidateCachedSupabaseProfile(userId)` to clear stale Redis cache entries after profile updates, but **multiple calls to this function are not awaited** — they are fired and immediately forgotten. This means:

1. The cache invalidation may **never complete** if it encounters an error (Redis timeout, connection error)
2. The HTTP response is sent **before** the cache is actually invalidated
3. A subsequent read request (from another route or another server instance) may still receive the **stale cached profile**
4. Unhandled promise rejections from the async invalidation calls are silently swallowed in older Node.js versions

## Locations

`backend/api/src/routes/profileRoutes.js` — the `invalidateCachedSupabaseProfile` calls at:

1. **Line ~125** — After profile name/email update
2. **Line ~170** — After phone number update
3. **Line ~219** — After avatar update

All three call sites follow the same pattern:

```js
invalidateCachedSupabaseProfile(userId);  // no await
res.json({ ... });
```

## Root Cause

The `invalidateCachedSupabaseProfile` function (imported from a cache service) is an `async` function that performs a Redis `DEL` operation. Without `await`, the caller does not wait for the Redis operation to complete before:

1. Sending the HTTP response
2. Returning from the route handler
3. Potentially ending the request-response cycle

The function signature likely looks like:

```js
async function invalidateCachedSupabaseProfile(userId) {
  const cacheKey = `profile:${userId}`;
  await redis.del(cacheKey);
}
```

Without awaiting, any error thrown by `redis.del()` will result in an **unhandled promise rejection**. In Node.js 14 and earlier, these are silently swallowed. In Node.js 16+, they trigger a `process.on('unhandledRejection')` event but still don't block the response.

## Impact

- **Stale profile data in cache**: Users see old profile information for up to the TTL duration (potentially hours) after updating
- **Inconsistent user experience**: A user updates their name but the mobile app still shows the old name because the cache wasn't cleared before the read happened
- **Race conditions**: If a request to read the profile arrives immediately after the update, it reads the stale cache before the invalidation completes (if it ever does)
- **Silent failures**: Redis connection errors during invalidation are silently swallowed, masking infrastructure problems
- **Eventual inconsistency**: The system works "eventually" when the cache TTL expires, but this defeats the purpose of explicit invalidation

## Expected Behavior

All cache invalidation calls should be awaited, and errors should be handled gracefully:

```js
try {
  await invalidateCachedSupabaseProfile(userId);
} catch (err) {
  // Log but don't fail the response — the profile update itself succeeded
  console.error(`Failed to invalidate cache for user ${userId}:`, err.message);
  // Optionally, report to Sentry or metrics
}

return res.json({ ... });
```

Or, if the response should not wait for cache invalidation (to reduce latency), use a fire-and-forget pattern with proper error handling:

```js
setImmediate(async () => {
  try {
    await invalidateCachedSupabaseProfile(userId);
  } catch (err) {
    console.error(`Cache invalidation failed for user ${userId}:`, err.message);
  }
});
```

This ensures:
1. Errors are logged and observable
2. The cache is eventually invalidated
3. The response is not delayed by the cache operation
4. Unhandled promise rejections are prevented

## Files to Modify

1. **`backend/api/src/routes/profileRoutes.js`** — Add `await` to all three `invalidateCachedSupabaseProfile()` calls, with try/catch error handling
2. **`backend/api/src/services/profileCache.js`** (or the cache service) — Review the function signature and ensure it properly handles errors internally

ionfwsrijan:
### Problem
Three `invalidateCachedSupabaseProfile(...)` calls on the profile update and wallet-update routes used the `void` operator, which silently swallows rejections. If the Supabase cache invalidation failed, the error was lost and the cache could serve stale data on the next request.

### Changes

**`backend/api/src/routes/profileRoutes.js`**
- Lines 125, 170, 219 — Replaced `void invalidateCachedSupabaseProfile(req.user.id)` with `await` inside try/catch, matching the identical pattern already used by the sibling `invalidateCachedProfile` call on line 122.

```js
// Before
void invalidateCachedSupabaseProfile(req.user.id);
```

```js
// After
try { await invalidateCachedSupabaseProfile(req.user.id); } catch (_) { /* logged internally */ }
```

### Impact

- Failures in Supabase-side cache invalidation are now caught (and logged) instead of silently lost.
- Cache consistency is preserved — a failed invalidation will not silently return stale data.

Closes #956

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved profile cache refresh reliability after wallet address, profile, and notification token updates.
  * Profile changes now wait for cache invalidation to complete, helping keep account data more consistent across the app.
  * Cache refresh errors are handled more safely, reducing the chance of silent failures.

### DIFF
--- a/backend/api/src/routes/profileRoutes.js
+++ b/backend/api/src/routes/profileRoutes.js
@@ -122,7 +122,7 @@ router.put('/wallet', authenticate, userLimiter, validateBody(updateWalletSchema
       try { await invalidateCachedProfile(req.user.uid); } catch (_) { logger.error('Cache invalidation failed', _); }
     }
     if (req.user && req.user.id) {
-      void invalidateCachedSupabaseProfile(req.user.id);
+      try { await invalidateCachedSupabaseProfile(req.user.id); } catch (_) { /* logged internally */ }
     }
 
     res.json({ success: true, walletAddress: normalized });
@@ -167,7 +167,7 @@ router.put('/', authenticate, userLimiter, validateBody(updateProfileSchema), as
       try { await invalidateCachedProfile(req.user.uid); } catch (_) { /* logged internally */ }
     }
     if (req.user && req.user.id) {
-      void invalidateCachedSupabaseProfile(req.user.id);
+      try { await invalidateCachedSupabaseProfile(req.user.id); } catch (_) { /* logged internally */ }
     }
 
     res.json({
@@ -216,7 +216,7 @@ router.put('/fcm-token', authenticate, userLimiter, async (req, res) => {
       try { await invalidateCachedProfile(req.user.uid); } catch (_) { /* logged internally */ }
     }
     if (req.user.id) {
-      void invalidateCachedSupabaseProfile(req.user.id);
+      try { await invalidateCachedSupabaseProfile(req.user.id); } catch (_) { /* logged internally */ }
     }
 
     return res.json({ success: true, message: 'FCM token updated successfully.' });


ionfwsrijan:

PR #995, opened 2026-06-27
fix: add missing logger import in profileRoutes.js (#971)

### LINKED ISSUE #971: bug: `logger` not imported in profileRoutes.js — `ReferenceError` on cache invalidation failure
## Description
`backend/api/src/routes/profileRoutes.js` calls `logger.error()` at line ~89 inside the wallet-update handler's `catch` block but never imports `logger` from `../middleware/logger.js`. If `invalidateCachedProfile()` throws, the server crashes with `ReferenceError: logger is not defined`.

## Impact
- Server crash (uncaught ReferenceError) when cache invalidation fails during a wallet update
- No error logging — the actual failure is silently lost
- Affects: POST `/profile/wallet-update` endpoint

## Steps to Reproduce
1. Trigger a wallet-update request that causes `invalidateCachedProfile()` to reject
2. Server throws `ReferenceError: logger is not defined`
3. The requesting client gets a 500 with no useful error information

## Expected Fix
Add at the top of `profileRoutes.js`:
```js
const logger = require('../middleware/logger');
```

## Location
`backend/api/src/routes/profileRoutes.js` — `catch` block around `invalidateCachedProfile()` call

## Note
Issue #959 adds logger to the `void` pattern calls in this file, but does **not** address the existing `logger.error()` call that was already present. Both changes are needed for full coverage.

## Labels
`bug`, `backend`, `high-priority`

vipul674:
Fixes #971 — server crashes with ReferenceError when cache invalidation fails because logger is never imported. Added `import logger from '../middleware/logger.js'`.

### DIFF
--- a/backend/api/src/routes/profileRoutes.js
+++ b/backend/api/src/routes/profileRoutes.js
@@ -12,6 +12,7 @@ import { ProfileModel } from '../models/ProfileModel.js';
 import { invalidateCachedProfile, invalidateCachedSupabaseProfile } from '../lib/profileCache.js';
 import { validateParams } from '../middleware/validate.js';
 import { paramIdSchema } from '../validation/requestSchemas.js';
+import logger from '../middleware/logger.js';
 import { updateProfileSchema, updateWalletSchema } from '../validation/requestSchemas.js';
 
 const router = express.Router();


vipul674:

PR #1005, opened 2026-06-27
fix: add pickupDate parameter to createOrder instead of hardcoding DateTime.now() (#979)

### LINKED ISSUE #979: bug: order creation always sends `DateTime.now()` as pickup date instead of user-selected date
## Description
`apps/customer/lib/services/order_service.dart` — `createOrder()` method accepts parameters `pickupAddress`, `dropAddress`, `pickupLat`, `pickupLng`, `dropLat`, `dropLng`, `pickupTime`, `goodsType`, `weightTonnes`, etc., but does NOT accept a `pickupDate` parameter. Instead, line 57 hardcodes:

```dart
'pickup_date': DateTime.now().toIso8601String(),
```

The user selects a pickup date in the UI (`find_trucks_screen.dart` stores it as `_selectedDate`), but this value is never forwarded to the API. Every order is created with the current device time as the pickup date, regardless of what the user selected.

## Impact
- CRITICAL — every order has the wrong pickup date
- If the user selected a future date, the order ships immediately (today)
- If the user is in a different timezone, the date may be off by a day
- Warehouse/scheduling relies on the pickup date — wrong date means missed pickups

## Steps to Reproduce
1. Open customer app
2. Search for trucks and set pickup date to tomorrow
3. Complete booking
4. Check API — saved pickup_date is today, not tomorrow

## Expected Fix
Add `pickupDate` parameter to `createOrder()` and pass it through:
```dart
'pickup_date': pickupDate, // from parameter instead of DateTime.now()
```

## Location
`apps/customer/lib/services/order_service.dart`, line 57.

vipul674:
Fixes #979 — order creation always sends DateTime.now() as pickup_date. Added optional pickupDate parameter so caller can pass user-selected date.

### DIFF
--- a/apps/customer/lib/services/order_service.dart
+++ b/apps/customer/lib/services/order_service.dart
@@ -35,6 +35,7 @@ class OrderService {
     required double weightTonnes,
     String? paymentMethodId,
     String? upiId,
+    DateTime? pickupDate,
   }) async {
     final user = SupabaseService.currentUser;
     final fullName = user?.userMetadata?['full_name']?.toString();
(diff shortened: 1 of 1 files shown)


vipul674:

PR #1006, opened 2026-06-27
fix: replace hardcoded seed UUID fallback with auth check in documents screen (#981)

### LINKED ISSUE #981: bug: documents screen uses hardcoded seed UUID as fallback when user is unauthenticated

vipul674:
Fixes #981 — documents screen uses hardcoded seed UUID when unauthenticated, potentially leaking data. Shows sign-in required error instead.

### DIFF
--- a/apps/customer/lib/screens/my_documents_screen.dart
+++ b/apps/customer/lib/screens/my_documents_screen.dart
@@ -47,7 +47,15 @@ class _MyDocumentsScreenState extends State<MyDocumentsScreen> {
 
     try {
       final user = _supabase.auth.currentUser;
-      final userId = user?.id ?? 'b1111111-1111-1111-1111-111111111111'; // Fallback to seed customer
+      if (user == null) {
+        if (!mounted) return;
+        setState(() {
+          _error = 'Please sign in to view your documents.';
+          _isLoading = false;
+        });
+        return;
+      }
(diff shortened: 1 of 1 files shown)


vipul674:

PR #1007, opened 2026-06-27
fix: only cache default Surat location when no cached location exists (#982)

### LINKED ISSUE #982: bug: home screen caches hardcoded Surat GPS coordinates on every launch, overriding real location

vipul674:
Fixes #982 — home screen caches hardcoded Surat GPS on every launch, overriding real location. Changed to only cache default when no location exists yet.

### DIFF
--- a/apps/customer/lib/screens/home_screen.dart
+++ b/apps/customer/lib/screens/home_screen.dart
@@ -37,7 +37,10 @@ class _HomeScreenState extends State<HomeScreen> {
     final connectivity = await Connectivity().checkConnectivity();
     final hasNetwork = connectivity.isNotEmpty && !connectivity.contains(ConnectivityResult.none);
     await _cacheManager.open();
-    await _cacheManager.cacheLastLocation(21.1702, 72.8311);
+    final existingLocation = await _cacheManager.getLastLocation();
+    if (existingLocation == null) {
+      await _cacheManager.cacheLastLocation(21.1702, 72.8311);
+    }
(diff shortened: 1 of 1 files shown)


vipul674:

PR #1015, opened 2026-06-27
fix(profile): add dedicated updateProfileSchema export in schemas/profile.js

### LINKED ISSUE #955: Profile Update Endpoint Crashes with Runtime Import Error Due to Missing Schema Export
## Description

The `profileRoutes.js` at line 15 imports `updateProfileSchema` from `requestSchemas.js`:

```js
const { updateProfileSchema } = require('../../validation/requestSchemas');
```

However, `requestSchemas.js` **never exports** a function or object called `updateProfileSchema`. This schema is supposed to validate the `PUT /api/profile` request body (name, email, phone, avatar updates), but since the import resolves to `undefined`, any call to this endpoint triggers a **runtime crash**:

```
TypeError: updateProfileSchema.validate is not a function
```

This makes the **entire profile update feature completely non-functional**.

## Location

- **`backend/api/src/routes/profileRoutes.js`** — line 15 (the import) and lines ~100-130 (the `PUT /` handler that uses it)
- **`backend/api/src/validation/requestSchemas.js`** — the entire file (the schema was likely planned but never written, or was accidentally omitted during a refactor)

## Root Cause

The `requestSchemas.js` file contains schemas for auth, load, driver, truck, and order operations, but does not contain a `profileSchema` or `updateProfileSchema` export. The `profileRoutes.js` middleware references it at:

```js
const { error: validationError } = updateProfileSchema.validate(req.body);
```

When `updateProfileSchema` is `undefined`, calling `.validate()` throws a `TypeError`. Depending on the error handling middleware configuration, this may:
1. Crash the request with a 500 Internal Server Error (if no global error handler exists)
2. Return an unhandled promise rejection (if the route handler is async)
3. Leak stack traces to the client (depending on `NODE_ENV`)

## Impact

- **Profile updates completely broken**: Users cannot update their name, email, phone number, or avatar URL
- **500 error on every profile update attempt**: The endpoint always returns an error, making it seem like the server is broken
- **User-facing errors**: In the Flutter apps, profile update attempts will fail with generic network errors, confusing users
- **Blocks related features**: Any feature that depends on profile updates (e.g., driver profile completion, customer account settings) is also blocked

### Option A: Create the Missing Schema (Recommended)

Add the missing `updateProfileSchema` to `requestSchemas.js`:

```js
const updateProfileSchema = Joi.object({
  name: Joi.string().min(2).max(100).trim(),
  email: Joi.string().email().lowercase().trim(),
  phone: Joi.string().pattern(/^\+?[\d\s-]{7,20}$/),
  avatar_url: Joi.string().uri().allow(''),
  bio: Joi.string().max(500).allow(''),
  // ... other profile fields
}).min(1);  // At least one field must be provided
```

Export it from the module:

```js
module.exports = {
  // ... existing exports
  updateProfileSchema,
};
```

### Option B: Add Inline Validation

If `requestSchemas.js` is intended to be the central validation hub, add the schema there. If the route should handle its own validation, remove the import from `profileRoutes.js` and validate inline.

## Files to Modify

1. **`backend/api/src/validation/requestSchemas.js`** — Add the `updateProfileSchema` Joi schema with proper field validation
2. **`backend/api/src/routes/profileRoutes.js`** — (Optionally) add a fallback check before using the schema:
   ```js
   if (!updateProfileSchema) {
     return res.status(500).json({ error: 'Profile validation schema not configured' });
   }
   ```

ionfwsrijan:
### Problem
The `updateProfileSchema` was not exported from any dedicated schema file. The profile update route (`PUT /`) accepted raw request bodies without Zod validation, allowing malformed payloads to reach the handler and potentially cause 500 errors or data corruption.

### Changes

**`backend/api/src/schemas/profile.js`** (new)
- Exports `updateProfileSchema` — a Zod object with optional fields `full_name`, `language`, `dark_mode`, `is_online` in strict mode (unknown fields are rejected).

**`backend/api/src/routes/profileRoutes.js`**
- Imports `validateBody` from `../middleware/validate.js`
- Imports `updateProfileSchema` from `../schemas/profile.js`
- Adds `validateBody(updateProfileSchema)` to the `PUT /` route so invalid requests are rejected with a 400 before executing the handler.

### Testing
- `PUT /api/profile` with an empty body → allowed (all fields optional)
- `PUT /api/profile` with an extra unknown field → `400 { error: ... }`
- `PUT /api/profile` with invalid field type (e.g. `dark_mode: "yes"`) → `400 { error: ... }`
- `PUT /api/profile` with valid fields → `200 { message: "Profile updated", profile: ... }`

Closes #955

### DIFF
--- a/backend/api/src/routes/profileRoutes.js
+++ b/backend/api/src/routes/profileRoutes.js
@@ -12,7 +12,8 @@ import { ProfileModel } from '../models/ProfileModel.js';
 import { invalidateCachedProfile, invalidateCachedSupabaseProfile } from '../lib/profileCache.js';
 import { validateParams } from '../middleware/validate.js';
 import { paramIdSchema } from '../validation/requestSchemas.js';
-import { updateProfileSchema, updateWalletSchema } from '../validation/requestSchemas.js';
+import { updateProfileSchema } from '../schemas/profile.js';
+import { updateWalletSchema } from '../validation/requestSchemas.js';
 
 const router = express.Router();
 
--- a/backend/api/src/schemas/profile.js
+++ b/backend/api/src/schemas/profile.js
@@ -1,0 +1,8 @@
+import { z } from 'zod';
+
+export const updateProfileSchema = z.object({
+  full_name: z.string().trim().min(1, 'Name cannot be empty').max(100, 'Name must be 100 characters or fewer').optional(),
+  language: z.string().min(2, 'Invalid language code').max(10, 'Invalid language code').optional(),
+  dark_mode: z.boolean().optional(),
+  is_online: z.boolean().optional(),
+}).strict();


ionfwsrijan:

PR #1115, opened 2026-06-28
fix : treat empty-string env vars as unset in readRateCard

### LINKED ISSUE #1109: fix : add null guard for empty string env vars in readRateCard
The readRateCard() helper in backend/api/src/lib/pricing.js uses Number(process.env.KEY ?? DEFAULTS.VALUE) to read pricing configuration from environment variables. When an environment variable is set to an empty string (), the nullish coalescing operator (??) does NOT fall through to the default because an empty string is not null or undefined — it is a defined value. Number() then silently returns 0, causing all pricing calculations to produce zero values in production.

Changes that Need to be Made:
- backend/api/src/lib/pricing.js: Modify the readRateCard() function to treat empty strings as unset, falling through to the defaults. A simple helper like a nullish empty-string check (e.g. (env ?? '').trim() === '' ? null : env) can be applied to each field, or each Number() call can be guarded with a non-empty-string check before conversion.

Impact that it would Provide:
- Production deployments with empty-string env vars will correctly fall back to hardcoded defaults instead of silently producing zero pricing.
- Financial calculations (baseFreight, tollEstimate, platformFee, totalAmount) will remain correct.

tmdeveloper007:
Closes #1109.

Introduced a rateEnv() helper that treats absent, null, empty-string, or non-numeric environment variable values as fall-through to the hardcoded defaults. Previously, Number('') returned 0, silently producing zero pricing for any empty-string env var.

Changes Made:
- backend/api/src/lib/pricing.js: Added rateEnv() helper and updated all 7 rate card field reads to use it

Impact it Made:
- Empty-string env vars now fall back to defaults instead of producing zero
- 302 unit tests still pass

### DIFF
--- a/backend/api/src/lib/pricing.js
+++ b/backend/api/src/lib/pricing.js
@@ -27,15 +27,28 @@ const DEFAULTS = Object.freeze({
   TOLL_PER_KM: 200,         // paisa per km, proxy for highway toll
 });
 
+/**
+ * Safely converts an environment variable to a finite number, falling back to
+ * the default if the variable is absent, empty, or contains a non-numeric value.
+ */
+function rateEnv(key, fallback) {
+  const raw = process.env[key];
+  if (raw === undefined || raw === null || raw.trim() === '') {
+    return fallback;
+  }
+  const parsed = Number(raw);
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

PR #1116, opened 2026-06-28
fix : use fire-and-forget for awardReputationPoints instead of blocking await

### LINKED ISSUE #1110: fix : use fire-and-forget for awardReputationPoints instead of blocking await
The submit_rating endpoint in backend/api/src/routes/orderRoutes.js (line ~561) calls await awardReputationPoints(polygonAddress, stars) on the HTTP critical path. The awardReputationPoints() function is documented as intentionally fire-and-forget — callers should NOT await it because blockchain confirmation can take seconds and must never block the HTTP response. The Supabase RPC is the source of truth for ratings.

Changes that Need to be Made:
- backend/api/src/routes/orderRoutes.js (submit_rating handler): Remove the await before awardReputationPoints() and replace it with void to make the fire-and-forget intent explicit: void awardReputationPoints(polygonAddress, stars). The existing try/catch wrapper is no longer needed since void Promises do not throw synchronously.

Impact that it would Provide:
- HTTP responses for the submit_rating endpoint will return immediately without waiting for blockchain confirmation.
- Rating submissions will feel instant to users even when Polygon RPC is slow or congested.
- The Supabase rating is saved regardless of blockchain outcome.

tmdeveloper007:
Closes #1110.

Replaced 'await awardReputationPoints(...)' with 'void awardReputationPoints(...).catch(...)' so blockchain confirmation does not block the HTTP response on the submit_rating endpoint. The Supabase rating is the source of truth; on-chain reputation is best-effort and should never delay the API response.

Changes Made:
- backend/api/src/routes/orderRoutes.js: Changed from 'try { await awardReputationPoints(...) } catch (repErr)' to 'void awardReputationPoints(...).catch((repErr) => ...)'

Impact it Made:
- Rating submission API returns immediately without waiting for blockchain confirmation
- Users see instant response even when Polygon RPC is slow or congested
- 302 unit tests still pass

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -557,11 +557,10 @@ router.post('/:id/ratings', authenticate, userLimiter, requireRole(['customer'])
     const polygonAddress = driverDetails?.polygon_wallet_address ?? null;
 
     if (polygonAddress) {
-      try {
-        await awardReputationPoints(polygonAddress, stars);
-      } catch (repErr) {
+      // Fire-and-forget: blockchain confirmation must never block the HTTP response.
+      void awardReputationPoints(polygonAddress, stars).catch((repErr) => {
         logger.error('[reputation] On-chain reputation update failed:', repErr.message);
-      }
+      });
     } else {
       logger.warn(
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

PR #1117, opened 2026-06-28
fix : return safe fallback string from predictDemandLimiter keyGenerator instead of throwing

### LINKED ISSUE #1111: fix : return safe fallback string from predictDemandLimiter keyGenerator instead of throwing
The predictDemandLimiter rate limiter in backend/api/src/routes/orderRoutes.js defines its keyGenerator as:
  if (!req.user || !req.user.id) { throw new Error('User is not authenticated'); }
Throwing from a keyGenerator crashes the express-rate-limit middleware with an unhandled exception, returning a 500 Internal Server Error to the client instead of applying the rate limit gracefully. The correct behaviour is to return a safe fallback key so the rate limit is applied by IP.

Changes that Need to be Made:
- backend/api/src/routes/orderRoutes.js: Change the keyGenerator's throw to return a fallback string using the IP-based keyGenerator from the existing userKeyGenerator helper (ipKeyGenerator from express-rate-limit, or use req.ip directly). For example: return req.ip ?? 'unknown-ip'.

Impact that it would Provide:
- The predict-demand endpoint will return a proper 429 rate-limit response when hit by an unauthenticated or rate-abusive client, instead of a 500 crash.
- The rate limiter will apply its cap to unauthenticated requests by IP, matching the intent of the limit.

tmdeveloper007:
Closes #1111.

Changed predictDemandLimiter's keyGenerator to return req.ip (or 'unknown-ip') when req.user is absent, instead of throwing an Error. Throwing from a keyGenerator crashes the express-rate-limit middleware and returns a 500 to the client instead of a proper 429 rate-limit response.

Changes Made:
- backend/api/src/routes/orderRoutes.js: Changed keyGenerator fallback from throw to return

Impact it Made:
- Unauthenticated requests to /predict-demand get rate-limited by IP instead of crashing
- Proper 429 responses instead of 500 crashes for unauthenticated rate abuse
- 302 unit tests still pass

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -129,7 +129,7 @@ const predictDemandLimiter = rateLimit({
   max: process.env.NODE_ENV === 'test' ? 1000 : 10,
   keyGenerator: (req) => {
     if (!req.user || !req.user.id) {
-      throw new Error('User is not authenticated');
+      return req.ip ?? 'unknown-ip';
     }
     return req.user.id;
   },


tmdeveloper007:

PR #1118, opened 2026-06-28
fix : verify on-chain tx sender matches registered customer wallet in confirm-deposit

### LINKED ISSUE #1112: fix : verify on-chain tx sender matches registered customer wallet in confirm-deposit
The recordDepositTx() function in backend/api/src/services/escrow.js verifies that a submitted transaction targets the Escrow contract and that the bookingId encoded in the deposit call matches, but it does NOT verify that tx.from (the transaction sender / msg.sender on-chain) matches the registered customer wallet address. An attacker could submit any deposit transaction for the correct booking ID from any wallet, and the backend would accept it.

The decoded deposit transaction (from escrowContract.interface.parseTransaction) includes the customer address as decoded.args[1]. The fix should verify that tx.from.toLowerCase() === decoded.args[1].toLowerCase() (case-insensitive since Ethereum addresses are checksumed).

Changes that Need to be Made:
- backend/api/src/services/escrow.js: After parsing the deposit transaction and validating the bookingId in recordDepositTx(), add a check: if (tx.from.toLowerCase() !== decoded.args[1].toLowerCase()) { return { error: 'Transaction sender does not match registered customer wallet' }; }

Impact that it would Provide:
- Prevents an attacker from confirming a deposit from a wallet that is not the registered customer wallet.
- Closes the security gap where any valid Polygon deposit call with the correct booking ID would be accepted regardless of sender.
- Aligns with the contract comment: deposit() requires msg.sender == customer.

tmdeveloper007:
Closes #1112.

Added a check in recordDepositTx() that verifies the on-chain transaction sender (tx.from) matches the customer address decoded from the deposit call arguments. Previously, any valid Polygon deposit transaction with the correct booking ID would be accepted regardless of the sender address.

Changes Made:
- backend/api/src/services/escrow.js: Added txCustomer extraction from decoded args and sender verification check

Impact it Made:
- Prevents attackers from confirming escrow deposits from wallets that are not the registered customer
- Closes a security gap in the confirm-deposit flow
- 302 unit tests still pass

### DIFF
--- a/backend/api/src/services/escrow.js
+++ b/backend/api/src/services/escrow.js
@@ -139,9 +139,15 @@ export async function recordDepositTx(bookingId, txHash) {
     return { error: 'Transaction is not a deposit call' };
   }
 
-  const [txBookingId] = decoded.args;
+  const [txBookingId, txCustomer] = decoded.args;
   if (txBookingId !== bookingId) {
     return { error: 'Transaction booking ID does not match' };
+  }
+
+  // Verify the on-chain sender (tx.from) matches the customer address in the deposit call.
+  // This prevents an attacker from submitting a deposit transaction from any wallet.
+  if (tx.from.toLowerCase() !== txCustomer.toLowerCase()) {
+    return { error: 'Transaction sender does not match registered customer wallet' };
   }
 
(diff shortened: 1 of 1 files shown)


tmdeveloper007:
Closing - this fix was already merged in a previous run.

PR #1137, opened 2026-06-28
fix: fix truck list query scoping bug in GET /api/trucks (#1126)

### LINKED ISSUE #1126: fix: truck list route skips query execution when name filter not provided
## Bug Description

In ackend/api/src/routes/truckRoutes.js, the GET / handler has a scoping bug where the query execution is inside the if (name) block.

**Result:** If 
ame is not provided, the entire database query is skipped and the endpoint never sends a response (hanging request). Additionally, min_capacity and max_capacity filters only work when 
ame is also provided.

## Proposed Fix

Move the closing brace of if (name) to only wrap the name filter line, so that capacity filtering and query execution happen regardless of whether name is provided.

sanrishi:
Resolves #1126.

The query execution was inside the if (name) block, causing the endpoint to hang when no name filter was provided. min_capacity/max_capacity filters also only applied when name was set. Corrects the brace placement so query runs unconditionally.

### DIFF
--- a/backend/api/src/routes/truckRoutes.js
+++ b/backend/api/src/routes/truckRoutes.js
@@ -79,6 +79,7 @@ router.get('/', authenticate, requireRole(['driver']), userLimiter, async (req, 
 
     if (name) {
       query = query.ilike('name', `%${name}%`);
+    }
     if (min_capacity) {
       query = query.gte('max_capacity_tons', Number(min_capacity));
     }


sanrishi:
@KanishJebaMathewM Please merge this PR. It fixes a scoping bug where the truck list endpoint hangs (never responds) when the \
ame\ query parameter is not provided. Capacity filters only worked when name was also set.

PR #1138, opened 2026-06-28
fix: handle empty string env vars in readRateCard() (#1132)

### LINKED ISSUE #1132: fix: readRateCard in pricing.js doesnt handle empty string env vars
## Bug Description

readRateCard() in backend/api/src/lib/pricing.js uses ?? for fallbacks, which only checks null/undefined. If an env var is set to empty string, Number('') returns 0 instead of the default, silently corrupting pricing calculations.

## Impact
Misconfigured env vars produce zero-value pricing, generating incorrect freight quotes.

## Proposed Fix
Use explicit truthy check instead of nullish coalescing for environment variable values.

sanrishi:
Resolves #1132.

?? only checks null/undefined, so empty string env vars produced Number('') = 0 instead of defaults. Adds parsePositiveInt/parsePositiveFloat helpers with explicit empty string checks.


sanrishi:
@KanishJebaMathewM Please merge this PR. It prevents silent pricing corruption when env vars are accidentally set to empty strings — \Number('')\ returns 0 instead of the default, producing incorrect freight quotes.

PR #1139, opened 2026-06-28
fix: add guard before calling ws.close() in spoof detection (#1127)

### LINKED ISSUE #1127: fix: tracker.js spoof detection calls ws.close() which throws on mock WebSocket
## Bug Description

In ackend/api/src/sockets/tracker.js:398, handleLocationPing calls ws.close(4010, ...) when a driver ID mismatch is detected. The mock WebSocket in tests doesn't have a close method, causing TypeError: ws.close is not a function.

## Impact
- 2 test failures in 	racker.test.js
- The spoof detection code path is not properly validated

## Proposed Fix
Add a guard before calling ws.close() and update the test to verify error messages.

sanrishi:
Resolves #1127.

handleLocationPing called ws.close() directly without checking if the method exists, causing TypeError on mock WebSocket in tests. Adds a typeof guard.


sanrishi:
@KanishJebaMathewM Please merge this PR. It adds a guard before \ws.close()\ in spoof detection, preventing a TypeError crash on mock WebSocket objects and making the code more testable.

PR #1145, opened 2026-06-28
fix: remove passthrough() from loadFilterQuerySchema to reject unknown params (fixes #1092)

### LINKED ISSUE #1092: [BUG] loadFilterQuerySchema uses passthrough() silently accepting unknown filter parameters
**Affected file:** `backend/api/src/validation/loadSchemas.js`

**Description:**  
The `loadFilterQuerySchema` Zod schema uses `.passthrough()` which allows any extra query parameters to pass through without validation or error. A typo in a query parameter (e.g., `trukc_type` instead of `truck_type`) will be silently ignored with no indication to the caller:

```js
// loadSchemas.js
export const loadFilterQuerySchema = z.object({
  page: z.coerce.number().int().positive().default(1),
  limit: z.coerce.number().int().positive().max(100).default(10),
  // ... valid filter fields ...
  material_type: enumSchema.optional(),
  weight: z.string().optional(),
  truck_type: z.string().optional(),
  // ...
}).passthrough();  // <--- allows unrecognized keys
```

**Impact:**
- Silent API misbehavior — user thinks they're filtering by a field but it's not applied
- Debugging difficulty: no warning in response or logs
- Inconsistent with the rest of the app's validation approach

**Suggested fix:**
1. Remove `.passthrough()` so unknown keys cause a validation error
2. Or use `.strip()` explicitly (which is the Zod default) to drop unknown keys without error
3. Or add logging when unknown keys are detected

vipul674:
Removes `.passthrough()` from `loadFilterQuerySchema`. Unknown query parameters will now be rejected with a validation error instead of being silently ignored.

Fixes #1092

### DIFF
--- a/backend/api/src/validation/loadSchemas.js
+++ b/backend/api/src/validation/loadSchemas.js
@@ -16,7 +16,7 @@ export const loadFilterQuerySchema = z.object({
   min_price: nonNegativeDecimalString('min_price').optional(),
   max_price: nonNegativeDecimalString('max_price').optional(),
   distance: nonNegativeDecimalString('distance').optional(),
-}).passthrough().superRefine((filters, ctx) => {
+}).superRefine((filters, ctx) => {
   if (
     filters.min_price !== undefined
     && filters.max_price !== undefined


vipul674:
@KanishJebaMathewM PR for #1092 ready for review. Removes `.passthrough()` from `loadFilterQuerySchema` so unknown query params are rejected.

PR #1153, opened 2026-06-28
fix: expire delivery OTPs on order cancellation (fixes #1085)

### LINKED ISSUE #1085: [BUG] Delivery OTPs not expired when order is cancelled — stale OTP records remain
**Affected file:** `backend/api/src/routes/orderRoutes.js` (cancel handler, lines 1243–1431)

**Description:**  
When a customer cancels an order via `POST /:id/cancel`, the handler updates the order status, manages escrow refund, and updates the timeline, but never calls `expireDeliveryOtps(orderId)`. Any delivery OTPs that were generated before cancellation remain active in the `delivery_otps` table with `verified = false`.

While a cancelled order's status prevents OTP verification from succeeding (line 962-963 guards against cancelled/payment_released), the stale OTP records are never cleaned up, leading to:
- Unnecessary rows in the database
- Potential confusion if the order is later reactivated
- Inconsistent behavior — other OTP lifecycle events (resend, successful verify) call `expireDeliveryOtps` or `clearOtpState`

**Evidence:**
```js
// orderRoutes.js:1368-1370 — cancel handler updates order_timeline
// Lines 1397-1426 — non-refund path updates order status
// Missing anywhere in the cancel handler: expireDeliveryOtps(orderId)
```

**Suggested fix:**
```js
// Add to the cancel handler after successful cancellation:
await expireDeliveryOtps(order.order_display_id);
```

vipul674:
Adds `await expireDeliveryOtps(order.order_display_id)` to the cancel handler so stale OTP records are cleaned up when an order is cancelled.

Fixes #1085

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -1436,6 +1436,8 @@ router.post('/:id/cancel', authenticate, userLimiter, requireRole(['customer']),
       .eq('order_display_id', order.order_display_id)
       .eq('milestone', 'Order Placed');
 
+    await expireDeliveryOtps(order.order_display_id);
+
     return res.json({ message: 'Order cancelled successfully.', cancellation_fee: cancellationFee, order: updatedOrder });
   } catch (err) {
     logger.error('Cancel order exception:', err.message);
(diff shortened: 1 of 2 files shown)


vipul674:
@KanishJebaMathewM PR for #1085 ready for review. Adds OTP expiration on order cancellation.

PR #1155, opened 2026-06-28
fix: add eviction limit to in-memory OTP failed attempts Map (fixes #1087)

### LINKED ISSUE #1087: [BUG] In-memory OTP failed attempts Map grows unbounded for abandoned orders when Redis unavailable
**Affected file:** `backend/api/src/routes/orderRoutes.js` (lines 131-169)

**Description:**  
The `inMemoryOtpFailedAttempts` Map stores OTP brute-force lockout state as a fallback when Redis is unavailable. However, entries are only cleaned up when `clearOtpState(orderId)` is called — which happens only on:

1. Successful OTP verification (line 985/1007)
2. New OTP generation during "In Transit" milestone (line 877)
3. OTP resend (line 1127)

If an order has OTP failures when Redis is down, and the order is then abandoned or cancelled without a successful verification, the Map entry for that `orderId` is **never removed** — it persists for the lifetime of the process.

**Code evidence:**
```js
// line 131
const inMemoryOtpFailedAttempts = new Map();

// line 135-140 — isOtpExpired only checks TTL, doesn't clean up the Map
// line 142-149 — checkOtpLockout deletes expired lock entries but only on read
// line 151-172 — recordOtpFailure creates/updates entries, no total-limit eviction
```

While each entry is small (~100 bytes), a high-volume deployment with frequent Redis outages could accumulate thousands of stale entries over time.

**Suggested fix options:**

1. Add a startup-parameter-based maximum size to `inMemoryOtpFailedAttempts` (e.g., `Map<...>` with eviction of oldest entries at 10,000 items)
2. Use a `LinkedHashMap` with access-order eviction
3. Also call `clearOtpState` in the cancel-order handler (see #1085)

vipul674:
Adds `IN_MEMORY_OTP_MAP_MAX_SIZE` configurable constant (default 10,000) with oldest-entry eviction when Map exceeds the limit, preventing unbounded memory growth.

Fixes #1087

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -42,6 +42,7 @@ const router = express.Router();
 const OTP_TTL_MINUTES = parseInt(process.env.OTP_TTL_MINUTES || '15', 10);
 const OTP_MAX_FAILED_ATTEMPTS = parseInt(process.env.OTP_MAX_FAILED_ATTEMPTS || '5', 10);
 const OTP_LOCKOUT_MINUTES = parseInt(process.env.OTP_LOCKOUT_MINUTES || '30', 10);
+const IN_MEMORY_OTP_MAP_MAX_SIZE = parseInt(process.env.IN_MEMORY_OTP_MAP_MAX_SIZE || '10000', 10);
 
 const inMemoryOtpFailedAttempts = new Map();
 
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM PR for #1087 ready for review. Adds eviction limit to in-memory OTP Map.

PR #1160, opened 2026-06-28
fix: use real backend trips instead of mock data in driver home screen (fixes #1095)

### LINKED ISSUE #1095: [BUG] Driver home screen uses mock tripHistory data instead of real backend trips
**Affected file:** `apps/driver/lib/screens/home_screen.dart` (line 833)

**Description:**  
The `_formatTimeSinceLastTrip()` method iterates over a `tripHistory` list that originates from `mock_data.dart` instead of fetching real trips from the backend:

```dart
// home_screen.dart:833
String _formatTimeSinceLastTrip() {
  // This likely references mock data rather than _tripService.fetchTrips()
  if (tripHistory.isEmpty) return 'No trips yet';
```

Meanwhile, `apps/driver/lib/screens/trips_screen.dart` already correctly fetches trips from Supabase via `_tripService`:

```dart
// trips_screen.dart — correct approach
List<TripRecord>? _trips;
// ...
_tripService.fetchTrips().then((trips) { ... });
```

**Impact:**
- The time-since-last-trip display on the home screen is unreliable
- A new driver with no real trips could see "last trip" data from mock data
- Inconsistent approach (trips_screen uses backend, home_screen uses mock data)

**Suggested fix:** Use the same `_tripService.fetchTrips()` pattern from trips_screen.dart to populate the home screen's trip history.

vipul674:
Replaces mock `tripHistory` reference with real data fetched via `_tripService.fetchTripHistory()`, consistent with the trips_screen.dart approach.

Fixes #1095

### DIFF
--- a/apps/driver/lib/screens/home_screen.dart
+++ b/apps/driver/lib/screens/home_screen.dart
@@ -13,7 +13,6 @@ import 'package:latlong2/latlong.dart' as ll;
 import 'package:truxify_driver/widgets/slide_to_confirm_button.dart';
 
 import '../core/app_routes.dart';
-import '../data/mock_data.dart';
 import '../models/app_models.dart';
 import '../models/earnings_daily_model.dart';
 import '../services/driver_earnings_service.dart';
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM PR for #1095 ready for review. Uses real backend trips instead of mock data on driver home screen.

PR #1169, opened 2026-06-28
fix: replace throw with safe fallback in predictDemandLimiter keyGenerator (fixes #1106)

### LINKED ISSUE #1106: [BUG] predictDemandLimiter keyGenerator throws uncaught Error instead of returning safe fallback string
**Affected file:** `backend/api/src/routes/orderRoutes.js` (lines 137-143)

**Description:**  
The `predictDemandLimiter` rate limiter's `keyGenerator` throws a `new Error()` when `req.user` is falsy:

```js
const predictDemandLimiter = rateLimit({
  windowMs: 60 * 1000,
  max: 10,
  keyGenerator: (req) => {
    if (!req.user || !req.user.id) {
      throw new Error('User is not authenticated');
    }
    return req.user.id;
  },
  // ...
});
```

`express-rate-limit` (v7.x) does not explicitly document catching errors from `keyGenerator`. A synchronous throw during rate-limit header calculation propagates unhandled. Currently, the `authenticate` middleware runs before this limiter (line 1479), so `req.user` is always set — making the throw dead code.

However, if middleware is reordered or refactored, this becomes an unhandled crash path. The standard pattern for defensive fallback is:

```js
keyGenerator: (req) => req.user?.id || 'unauthenticated',
```

**Impact:** Low probability but high impact if triggered — could crash the rate limiter instance and affect all requests using `predictDemandLimiter`.

**Suggested fix:** Return a safe fallback string instead of throwing:
```js
keyGenerator: (req) => req.user?.id || 'unauthenticated',
```

vipul674:
Replaces `throw new Error()` with `req.user?.id || 'unauthenticated'` fallback in the `predictDemandLimiter` keyGenerator to prevent unhandled crashes if middleware ordering changes.

Fixes #1106

### DIFF
--- a/apps/driver/lib/data/mock_data.dart
+++ b/apps/driver/lib/data/mock_data.dart
@@ -426,6 +426,11 @@ final Map<String, LoadOffer> loadOfferById = {
 
 const List<Trip> mockTrips = [
   Trip(
+    id: 'mock-trip-1',
+    route: 'Surat → Jaipur',
+    date: 'Today · 6:00 AM',
+    items: ['Textile 3t', 'Electronics 2t'],
+  Trip(
     route: 'Surat → Jaipur',
     date: 'Today · 6:00 AM',
     items: ['Textile 3t', 'Electronics 2t'],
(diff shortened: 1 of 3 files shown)


vipul674:
@KanishJebaMathewM PR for #1106 ready for review. Replaces throw with safe fallback in keyGenerator.

PR #1215, opened 2026-06-28
fix(profileRoutes): add missing 'format' query param and close CSV if block

### LINKED ISSUE #1204: fix: ReferenceError on undefined format var + broken CSV block in profileRoutes.js
## Bug Description

\ackend/api/src/routes/profileRoutes.js\ has two bugs in the driver statement CSV branch:

1. **Line 298**: \if (format === 'csv')\ references \ormat\ which was never destructured from \
eq.query\ — throws \ReferenceError\.
2. **Missing closing brace**: The CSV \if\ block is never closed, so the \sort_by\ logic below it is syntactically inside it and unreachable for JSON responses.

## Impact

- Every request to \GET /api/profile/driver/statement?format=csv\ crashes with ReferenceError
- Every request with \sort_by\ silently ignores the sort parameter

## Fix

Add \ormat\ to the destructured query params and properly close the CSV block.

sanrishi:
## Summary
Fixes two bugs in the GET /driver/statement endpoint:
1. \ormat\ was not destructured from \
eq.query\, causing a \ReferenceError\ when the CSV branch was hit.
2. The \if (format === 'csv')\ block was missing its closing brace, causing a syntax error that prevented the module from loading.

## Changes
- Added \ormat\ to the destructured query params
- Added \}\ to close the CSV if block before the sort logic

Closes #1204

### DIFF
--- a/backend/api/src/routes/profileRoutes.js
+++ b/backend/api/src/routes/profileRoutes.js
@@ -242,7 +242,7 @@ router.put('/fcm-token', authenticate, userLimiter, async (req, res) => {
 // GET DRIVER STATEMENT
 router.get('/driver/statement', authenticate, requireRole(['driver']), userLimiter, validateQuery(driverStatementSchema), async (req, res) => {
   const userId = req.user.id;
-  const { start_date, end_date, sort_by } = req.query;
+  const { start_date, end_date, sort_by, format } = req.query;
 
   try {
     let query = supabase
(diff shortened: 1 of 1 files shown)


sanrishi:
@KanishJebaMathewM Please merge this PR. The \ormat\ ReferenceError crashes the CSV driver statement endpoint, and the missing closing brace causes a syntax error that prevents the entire profileRoutes module from loading. Both are critical bugs that break driver financial reporting.

PR #1216, opened 2026-06-28
fix(tripRoutes): remove const type redeclaration

### LINKED ISSUE #1206: fix: const redeclaration of type in tripRoutes.js causes SyntaxError
## Bug Description

\ackend/api/src/routes/tripRoutes.js:207-208\ has \const type\ declared twice in the same scope:

\\\js
const { type, sort } = req.query;       // line 207
const { type, min_lat, max_lat, min_lng, max_lng } = req.query;  // line 208 - SyntaxError!
\\\

This produces \SyntaxError: Identifier 'type' has already been declared\.

## Impact

The trip routes module cannot load. All trip event fetching endpoints are down.

## Fix

Rename the first \	ype\ to \ventType\ to avoid the conflict.

sanrishi:
## Summary
The \	ype\ variable was declared twice with \const\ in the same function scope (route handler for GET /:id/events), causing a SyntaxError that prevents the module from loading.

## Changes
- Merged the two separate destructuring assignments into a single \const { type, sort, min_lat, max_lat, min_lng, max_lng } = req.query;\

Closes #1206

### DIFF
--- a/backend/api/src/routes/tripRoutes.js
+++ b/backend/api/src/routes/tripRoutes.js
@@ -203,9 +203,8 @@ router.post('/events/batch', authenticate, userLimiter, validateBatchPayload(bat
  */
 router.get('/:id/events', authenticate, userLimiter, async (req, res) => {
   const tripId = req.params.id;
-  const { type, sort } = req.query;
+  const { type, sort, min_lat, max_lat, min_lng, max_lng } = req.query;
   const isAscending = sort !== 'desc';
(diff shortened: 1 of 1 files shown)


sanrishi:
@KanishJebaMathewM Please merge this PR. The duplicate const type declaration is a SyntaxError that crashes the entire trip routes module, making all trip-related endpoints (events, tracking) unavailable — critical server-breaking bug.

PR #1219, opened 2026-06-28
fix(reputation): add decreaseReputation to ABI

### LINKED ISSUE #1208: fix: decreaseReputation missing from reputation.js ABI — asymmetric on-chain scoring
## Bug Description

\ackend/api/src/services/reputation.js:26-28\ only includes \increaseReputation\ in the ABI, but \Reputation.sol\ also has \decreaseReputation\. This means the backend can only add reputation points, never subtract them — even for fraudulent drivers.

## Impact

The on-chain reputation system is asymmetric. Negative ratings and fraudulent behavior cannot be reflected on-chain, undermining the trust system's integrity.

## Fix

Add \decreaseReputation\ to the ABI and export a corresponding \deductReputationPoints\ function.

sanrishi:
## Summary
The \REPUTATION_ABI\ was missing the \decreaseReputation\ function signature, even though the module comments reference it and the contract exposes it. Any code that tries to call \decreaseReputation\ on the contract through this client would fail.

## Changes
- Added \decreaseReputation\ function signature to \REPUTATION_ABI\

Closes #1208

### DIFF
--- a/backend/api/src/services/reputation.js
+++ b/backend/api/src/services/reputation.js
@@ -25,6 +25,7 @@ import logger from '../middleware/logger.js';
 // Minimal ABI — only the subset the backend needs to call.
 const REPUTATION_ABI = [
   'function increaseReputation(address driver, uint256 points) external',
+  'function decreaseReputation(address driver, uint256 points) external',
   'function getReputation(address driver) external view returns (uint256)',
 ];
 /** @type {ethers.Contract | null} */


sanrishi:
@KanishJebaMathewM Please merge this PR. The ABI was incomplete — decreaseReputation exists in the smart contract but was missing from the client ABI. Any future or existing code attempting to decrease reputation on-chain would silently fail (etherjs returns success with null data for missing function selectors).

PR #1221, opened 2026-06-28
fix(Escrow): implement pull-based withdrawal pattern

### LINKED ISSUE #1210: fix: add pull-based withdrawal pattern to Escrow.sol to prevent locked funds
## Bug Description

\lockchain/contracts/Escrow.sol:86,100\ uses raw \call\ to send Ether to the driver/customer. If the recipient is a smart contract that rejects ETH (e.g., a multisig wallet without a receive() function), the entire transaction reverts after the escrow status is already updated, leaving no recovery path.

## Impact

Funds can be permanently locked in the escrow contract if the recipient address is a contract that does not accept ETH. There is no admin backstop or withdrawal pattern.

## Fix

Implement a pull-based withdrawal pattern using \claimableBalances\ mapping so funds are claimable by the recipient rather than pushed.

sanrishi:
## Summary
Replaced push-based fund transfers with a pull-based withdrawal pattern for security. Previously, \
eleaseFunds()\ and \
efundFunds()\ sent ether directly via \.call()\, which could fail due to recipient contract reverts or gas issues, potentially locking funds.

## Changes
- \
eleaseFunds()\ now credits \pendingWithdrawals[driver]\ instead of sending ether directly
- \
efundFunds()\ now credits \pendingWithdrawals[customer]\ instead of sending ether directly
- Added \withdraw()\ function for recipients to claim funds themselves
- Added \pendingWithdrawals\ mapping and \Withdrawn\ event

Closes #1210

### DIFF
--- a/blockchain/contracts/Escrow.sol
+++ b/blockchain/contracts/Escrow.sol
@@ -19,12 +19,14 @@ contract Escrow {
     address public owner;
     mapping(address => bool) public authorizedRelayers;
     mapping(bytes32 => BookingEscrow) public escrows;
+    mapping(address => uint256) public pendingWithdrawals;
     bool private locked;
 
     event RelayerUpdated(address indexed relayer, bool authorized);
     event Deposited(bytes32 indexed bookingId, address indexed customer, address indexed driver, uint256 amount);
     event Released(bytes32 indexed bookingId, address indexed driver, uint256 amount);
     event Refunded(bytes32 indexed bookingId, address indexed customer, uint256 amount);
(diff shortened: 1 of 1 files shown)


sanrishi:
@KanishJebaMathewM Please merge this PR. The push-based transfer pattern is a security vulnerability — if a relayer's call fails (e.g., malicious fallback function, out-of-gas), the funds remain in the contract but the booking state is already marked Released/Refunded, effectively locking them. Pull-based withdrawal lets recipients claim their own funds safely.

PR #1222, opened 2026-06-28
fix(ml): report actual model status in /health endpoint

### LINKED ISSUE #1211: fix: ML engine /health endpoint should validate model readiness
The /health endpoint in backend/ml/app/main.py always returns healthy regardless of whether ML models actually loaded. If models fail during startup, Docker healthchecks still pass and orchestrators wont restart the failed container. Fix by checking that model files exist in the storage directory and returning 503 if missing.

sanrishi:
## Summary
The \/health\ endpoint always returned \{'status': 'healthy'}\ irrespective of whether ML models were loaded/trained. For container orchestration with liveness/readiness probes, the endpoint should reflect actual model availability.

## Changes
- \/health\ now returns per-model readiness status
- Returns \'degraded'\ overall status when models are missing, \'healthy'\ when all are ready

Closes #1211

### DIFF
--- a/backend/ml/app/main.py
+++ b/backend/ml/app/main.py
@@ -21,6 +21,9 @@ from .models.collaborative_filter import collaborative_filter
 from .models.trust_scorer import trust_scorer
 from .models.deadhead_eliminator import find_return_loads
 from .models.mid_trip_reoptimiser import find_mid_trip_loads
+from .models.base import model_exists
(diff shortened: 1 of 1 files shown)


sanrishi:
@KanishJebaMathewM Please merge this PR. The /health endpoint always reports 'healthy' even when no ML models are trained/loaded, which defeats the purpose of liveness/readiness probes in container orchestration. This change surfaces actual model readiness so Kubernetes/Docker can make correct scheduling decisions.

PR #1288, opened 2026-06-29
fix(loadRoutes): guard vehicle_type against array values Closes #1267

### LINKED ISSUE #1267: fix(loadRoutes): vehicle_type array causes .toLowerCase() crash
In ackend/api/src/routes/loadRoutes.js:51, 
eq.query.vehicle_type.toLowerCase() will throw \TypeError: req.query.vehicle_type.toLowerCase is not a function\ if the user passes multiple \ehicle_type\ values (Express parses this as an array). Unlike the \status\ filter on line 71 which handles arrays, \ehicle_type\ is unprotected.

sanrishi:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of the `vehicle_type` filter on the loads listing endpoint.
  * Non-string values for `vehicle_type` no longer trigger a validation error; they are now treated as unset.
  * Requests with a string value other than `truck` still return empty results and zero totals as expected.


sanrishi:
@KanishJebaMathewM Please merge this PR. Passing multiple vehicle_type query params crashes the server with TypeError. Unlike the status filter which handles arrays, vehicle_type was unprotected.

PR #1289, opened 2026-06-29
fix(requestSchemas): remove z.any() pricing fields from createOrderSchema Closes #1286

### LINKED ISSUE #1286: fix(requestSchemas): z.any() fields allow client price manipulation
In backend/api/src/validation/requestSchemas.js lines 55-59, fields `base_freight`, `toll_estimate`, `platform_fee`, `total_amount`, and `estimated_price` use `z.any().optional()`. Any client-supplied value passes validation, allowing clients to set arbitrary pricing values that could override server-computed pricing. These fields should be rejected or stripped from client input since the server is responsible for computing pricing.

sanrishi:


sanrishi:
@KanishJebaMathewM Please merge this PR. z.any() pricing fields allow clients to set arbitrary values that override server-computed pricing. This is a security vulnerability enabling price manipulation by the client.

PR #1292, opened 2026-06-29
fix(Reputation): prevent overflow before MAX_REPUTATION cap

### LINKED ISSUE #1273: fix(Reputation): overflow before MAX_REPUTATION cap in Solidity 0.8
In blockchain/contracts/Reputation.sol line 41, `uint256 newScore = scores[driver] + points;` -- Solidity 0.8.x has built-in overflow checking, so if `points` is large enough to overflow uint256, the transaction reverts before the cap logic on line 42 (`newScore > MAX_REPUTATION ? MAX_REPUTATION : newScore`) ever executes. An attacker with relayer access could DoS legitimate reputation increases by passing a very large `points` value.

sanrishi:
In Solidity 0.8+ built-in overflow checking causes the addition
on line 41 to revert before the cap check. Now uses overflow-safe
addition that caps at MAX_REPUTATION instead of reverting.

Closes #1273

### DIFF
--- a/blockchain/contracts/Reputation.sol
+++ b/blockchain/contracts/Reputation.sol
@@ -38,8 +38,14 @@ contract Reputation {
 
     function increaseReputation(address driver, uint256 points) external onlyRelayer {
         require(driver != address(0), "Invalid driver");
-        uint256 newScore = scores[driver] + points;
-        scores[driver] = newScore > MAX_REPUTATION ? MAX_REPUTATION : newScore;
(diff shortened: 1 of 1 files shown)


sanrishi:
@KanishJebaMathewM Please merge this PR. Solidity 0.8+ overflow checking reverts the transaction before the MAX_REPUTATION cap can be applied. An attacker with relayer access can DoS reputation increases by passing large points values.

PR #1296, opened 2026-06-29
fix(escrowRefundReconciliation): handle scalar RPC return values

### LINKED ISSUE #1275: fix(escrowRefundReconciliation): .length check on scalar RPC return
In backend/api/src/services/escrowRefundReconciliation.js line 62, `claimed.length === 0` assumes the `claim_refund_reconciliation` RPC always returns an array. If it returns a single row as an object, `claimed.length` is undefined, and `undefined === 0` is false -- so the code proceeds to process the refund anyway, potentially causing double-processing of refunds.

sanrishi:
The claimed.length check assumed the RPC always returns an array.
When the RPC returns a single row (object), claimed.length is undefined
causing the skip condition to fail silently, potentially double-processing
refunds.

Closes #1275

### DIFF
--- a/backend/api/src/services/escrowRefundReconciliation.js
+++ b/backend/api/src/services/escrowRefundReconciliation.js
@@ -59,7 +59,7 @@ export async function reconcilePendingEscrowRefunds() {
             p_instance_id: instanceId,
           });
 
-        if ((!claimed || claimed.length === 0) && !claimError) {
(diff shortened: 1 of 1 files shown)


sanrishi:
@KanishJebaMathewM Please merge this PR. The .length check on RPC results assumes array return, but if the RPC returns a scalar object the check silently fails, potentially double-processing refunds.

PR #1297, opened 2026-06-29
fix(flutter): gracefully handle Firebase init failure instead of rethrowing

### LINKED ISSUE #1271: fix(customer main): Firebase rethrow crashes app while Supabase gracefully logs
In \pps/customer/lib/main.dart:40\ (and analogously in the driver app), \
ethrow\ after Firebase initialization failure causes the app to crash immediately. However, Supabase initialization (line 44-52) silently catches errors. The inconsistency means a Firebase configuration issue is fatal while a Supabase issue is tolerated. Firebase init should be handled gracefully.

sanrishi:
Both apps used rethrow after Firebase initialization failure, causing
unrecoverable crashes. Changed to match Supabase's graceful error handling
so a Firebase misconfiguration doesn't crash the app.

Closes #1271

### DIFF
--- a/apps/customer/lib/main.dart
+++ b/apps/customer/lib/main.dart
@@ -37,7 +37,6 @@ void main() async {
     }
   } catch (e) {
     debugPrint('Firebase initialization failed: $e');
-    rethrow;
   }
 
   // Initialize Supabase if keys are provided.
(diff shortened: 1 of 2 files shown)


sanrishi:
@KanishJebaMathewM Please merge this PR. Both Flutter apps crash on Firebase init failure due to rethrow, while Supabase errors are gracefully logged. This inconsistency makes Firebase issues fatal when they should be recoverable.

PR #1298, opened 2026-06-29
fix(authRoutes): clear dangling timeouts on success

### LINKED ISSUE #1278: fix(authRoutes): Promise.race timeouts not cleared on success
In backend/api/src/routes/authRoutes.js lines 30-36 and 45-49, Promise.race() schedules setTimeout that rejects if the target operation does not complete in time. However, if the operation completes successfully before the timeout, the setTimeout is never cleared with clearTimeout(). This leaves dangling timers in the Node.js event loop, causing a minor memory leak and unnecessary event-loop work until they fire.

sanrishi:
Promise.race timers were never cleared when the target operation
completed before the timeout, causing dangling timers in the event loop.

Closes #1278


sanrishi:
@KanishJebaMathewM Please merge this PR. Promise.race timeouts were leaking as dangling timers when operations completed successfully, causing minor memory leaks.

Issue #1338, opened 2026-06-30
fix(profileCache): ttlSeconds <= 0 guard prevents caching near expiry boundary


sanrishi:
In backend/api/src/lib/profileCache.js:194, the 	tlSeconds <= 0 guard causes setCachedSupabaseProfile to skip caching when TTL is exactly 0. In auth.js line 195-197, TTL is computed as Math.min(TTL_SECONDS, decoded.exp - nowSeconds). If the token expires in exactly 0 seconds, TTL=0 and the cache is skipped, forcing a DB query on every request. Fix: use Math.max(1, ttlSeconds) to allow caching with minimum 1-second TTL.

Issue #1337, opened 2026-06-30
fix(auth): decoded variable scope leak into Supabase TTL calc


sanrishi:
In backend/api/src/middleware/auth.js:69, decoded is set by unverified jwt.decode(token) before the token type is determined. At line 195-196, the Supabase TTL calculation references decoded?.exp which is only valid for Supabase tokens. If a Firebase token happens to have an xp field, this could cause wrong TTL. Fix: move decoded into the Supabase-only code block scope.

Issue #1336, opened 2026-06-30
fix(auth): misleading 403 error for deactivated profiles


sanrishi:
In backend/api/src/middleware/auth.js:93-97 and 135-139, when a cached profile has isActive === false, the response says 'User profile not found in database. Register user in profiles table first.' This is factually wrong - the profile WAS found (in cache) but is simply deactivated. Confuses users into thinking they need to re-register. Fix: return 'User profile is inactive/deactivated.'

PR #1348, opened 2026-06-30
fix(backend): validate params DEF, auth misleading 403 + decoded scope, profileCache TTL

### LINKED ISSUE #1333: fix(validate): validateParams missing Object.defineProperty for Express 5 compat
In backend/api/src/middleware/validate.js:35, 
eq.params = result.data directly assigns without the Object.defineProperty protection used in alidateQuery (lines 54-59). In Express 5, 
eq.params can be immutable in certain configurations, causing the assignment to silently fail while 
ext() is called with stale param values. This is identical to the bug that required the Object.defineProperty wrapper for 
eq.query.

sanrishi:
Closes #1333, #1336, #1337, #1338

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of cached user profiles so short or missing cache durations are treated more reliably.
  * Updated access-denied messaging for inactive accounts to clearly explain the issue and next steps.
  * Made request parameter handling more consistent, helping reduce validation-related issues.

### DIFF
--- a/backend/api/src/lib/profileCache.js
+++ b/backend/api/src/lib/profileCache.js
@@ -191,7 +191,8 @@ export async function getCachedSupabaseProfile(userId) {
  */
 export async function setCachedSupabaseProfile(userId, profile, ttlSeconds = TTL_SECONDS) {
   const redisClient = getRedisClient();
-  if (!redisClient || !userId || !profile || ttlSeconds <= 0) return;
+  if (!redisClient || !userId || !profile) return;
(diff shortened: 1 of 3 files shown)


sanrishi:
@KanishJebaMathewM Please merge this PR. Contains 4 fixes: validateParams breaks in Express 5 without Object.defineProperty, auth.js returns wrong error for deactivated profiles, decoded variable scope leak, and profileCache rejects TTL=0 forcing DB queries on every request.

Issue #1335, opened 2026-06-30
fix(deviceController): profile sync after response causes data inconsistency


sanrishi:
In backend/api/src/controllers/deviceController.js:40-55, the cm_token profile sync runs after res.json() at line 55. If the profile update fails, the client gets success:true but profiles.fcm_token is stale. The user_devices table has the token but profiles doesn't, causing silent data inconsistency. Fix: move the profile sync before the response or use a transaction.

vipul674:
Hi @KanishJebaMathewM, I'd like to work on this issue under GSSoC'26. Please assign it to me.

aaniya22:
pls assign this to me under gssoc

Rakshak05:
Hii @KanishJebaMathewM, I would like to work on this. Can you please assign it to me under GSSoC'26 contributor!?

aaniya22:
Verified on latest main (f554580f) — in backend/api/src/controllers/deviceController.js, the profiles.fcm_token sync (line ~108) now runs and is error-checked before res.json() (line ~122). If profileSyncError occurs, the handler returns a 500 via next(new AppError(...)) instead of a false success response. This matches the fix already merged in #1349.

PR #1349, opened 2026-06-30
fix(deviceController): FCM token hijack + profile sync before response

### LINKED ISSUE #1334: fix(deviceController): FCM token conflict hi-jacks another user's device
In backend/api/src/controllers/deviceController.js:30, the upsert uses onConflict: 'fcm_token'. If two different users register the same FCM token (e.g., shared device, or app reinstall under a different account), the device record is silently reassigned from the original user to the new user. The original user stops receiving push notifications. Fix: use composite conflict target (user_id, fcm_token) or check ownership first.

sanrishi:
Two fixes:
1. Changed onConflict from 'fcm_token' to 'user_id,fcm_token' to
   prevent device token hijacking when different users share a token.
2. Moved profile sync before the response so client receives error
   if profiles.fcm_token update fails, avoiding data inconsistency.

Closes #1334, #1335

## Summary by CodeRabbit

* **Bug Fixes**
  * Device token registration now keeps user profile token data in sync more reliably.
  * If a token is reassigned to a different account, the previous account’s saved token is cleared.
  * Failed token updates now return a clear server error instead of silently logging the issue.
  * Improved handling for duplicate device token registration so token ownership is updated consistently.

### DIFF
--- a/backend/api/src/controllers/deviceController.js
+++ b/backend/api/src/controllers/deviceController.js
@@ -1,71 +1,3 @@
-// ===== PR (fix/issue-1334-device-token) changes =====
-import { supabase } from '../config/db.js';
-import logger from '../middleware/logger.js';
-
-/**
- * Register / update FCM token for a user device
- */
-export async function registerDeviceToken(req, res) {
-  try {
-    const userId = req.user?.id;
-    const { fcmToken, platform } = req.body;
-
-    if (!userId) {
-      return res.status(401).json({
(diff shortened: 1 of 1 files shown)


sanrishi:
@KanishJebaMathewM Please merge this PR. Two critical bugs: device token hijacking (upsert with onConflict:fcm_token silently reassigns devices between users) and data inconsistency (profile sync runs after success response sent).

Issue #1344, opened 2026-06-30
fix(notification_item): required but nullable userId is contradictory


sanrishi:
In packages/truxify_shared/lib/src/models/notification_item.dart:12-13, `userId` is declared `final String?` (nullable) but the constructor marks it `required`. Every call site must pass `userId: null` for unknown users, defeating the purpose of `required`. Either make it non-nullable String or remove `required`.

prasiddhi-105:
hiiii please assign this issue to me under ECSoc'26, once you assign i can work and submit a PR

Issue #1343, opened 2026-06-30
fix(support_ticket): empty string id instead of UUID


sanrishi:
In packages/truxify_shared/lib/src/models/support_ticket.dart:11, id is required non-nullable String. The SupportRepository creates tickets with id: '' (empty string). If any code reads ticket.id before persistence to Supabase (where DB generates UUID), it gets an empty string causing lookup failures. Fix: generate a UUID client-side.

prasiddhi-105:
hiiii please assign this issue to me under ECSoc'26, once you assign i can work and submit a PR

Issue #1342, opened 2026-06-30
fix(resilient_websocket): TOCTOU race condition in heartbeat timer


sanrishi:
In apps/customer/lib/core/offline/websocket/resilient_websocket.dart:83-86, the heartbeat timer checks if (_channel != null) at line 84, then uses _channel! at line 85. Between the check and the null assertion, the close() method (or a connection error) could set _channel = null, causing a NoSuchMethodError. Fix: capture _channel in a local variable before the check.

prasiddhi-105:
hiiii please assign this issue to me under ECSoc'26, once you assign i can work and submit a PR

Issue #1341, opened 2026-06-30
fix(api_client): jsonDecode cast outside try block causes unhandled TypeError


sanrishi:
In apps/customer/lib/core/api_client.dart:242-244, the jsonDecode(response.body) as Map<String, dynamic> cast at line 243 is OUTSIDE the try block on line 246. If the API returns a JSON array or primitive, the s cast throws an unhandled TypeError instead of being caught. Fix: move the cast inside the try block.

PR #1351, opened 2026-06-30
fix(flutter): api_client cast, websocket race, ticket id, notification userId

### LINKED ISSUE #1340: fix(api_client): null reasonPhrase produces literal 'null' string error messages
In apps/customer/lib/core/api_client.dart:244, 
esponse.reasonPhrase.toString() is called when reasonPhrase is null (HTTP/2). This yields the literal string 'null' displayed to users. At line 247, the catch block also accesses 
esponse.reasonPhrase where it could be null in HTTP/2. Fix: use proper null coalescing.

sanrishi:
4 fixes:
- api_client: jsonDecode cast outside try/catch, null reasonPhrase
- resilient_websocket: TOCTOU race in heartbeat and send
- support_ticket: made id optional with default empty string
- notification_item: removed required from nullable userId

Closes #1340, #1341, #1342, #1343, #1344

### DIFF
--- a/apps/customer/lib/core/api_client.dart
+++ b/apps/customer/lib/core/api_client.dart
@@ -240,9 +240,12 @@ class ApiClient {
 
     String message;
     try {
-      final json = jsonDecode(response.body) as Map<String, dynamic>;
-      message = (json['error'] ?? json['message'] ?? response.reasonPhrase)
-          .toString();
+      final decoded = jsonDecode(response.body);
+      if (decoded is Map<String, dynamic>) {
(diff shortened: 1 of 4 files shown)


sanrishi:
@KanishJebaMathewM Please merge this PR. Contains 5 Flutter fixes: unhandled TypeError from jsonDecode cast, null reasonPhrase producing 'null' strings, TOCTOU race in websocket heartbeat, support_ticket empty id, and notification_item contradictory required nullable.

PR #1452, opened 2026-07-01
Fix #1449: Change-Drop After Bid Acceptance Overwrites total_amount, Creating On-Chain/Off-Chain Financial Desync

### LINKED ISSUE #1449: Change-Drop After Bid Acceptance Overwrites `total_amount`, Creating On-Chain/Off-Chain Financial Desync
### Description

The order lifecycle has a critical window where `total_amount` becomes desynchronized from the blockchain escrow amount, allowing a financial mismatch:

1. **Order Creation** — `total_amount` is set to `pricing.totalAmount` from pickup/drop coordinates.
2. **Bid Acceptance** — The `accept_bid_tx` RPC overwrites `total_amount` to `p_bid_amount` (the driver's winning bid).
3. **Escrow Funding** — The blockchain escrow deposit is calculated from `bid.bid_amount`.
4. **Change-Drop** — The guard only blocks if `escrow_status === 'funded'` (line 1180), but between bid acceptance and escrow funding there is NO guard. Change-drop recomputes pricing and overwrites `total_amount` (line 1219).
5. **Trip Completion** — The `complete_trip_tx` RPC pays out `v_order.total_amount`, which now differs from the escrowed bid amount.

---

### Actual Behavior

Between bid acceptance (step 2) and escrow funding confirmation (step 3), a customer can change the drop location. This recomputes `total_amount` to a new value. The blockchain escrow was already locked for the original bid amount. When the trip completes, the driver is paid the new `total_amount` from the database wallet, but the blockchain escrow only holds the original bid amount. The platform must cover the difference.

Example: Order A→B, bid = ₹8,000, escrow funded for ₹8,000. Customer changes drop → `total_amount` = ₹15,000. Trip completes → driver paid ₹15,000 but escrow only holds ₹8,000. Platform loses ₹7,000.

---

### Expected Behavior

The `total_amount` should be immutable after bid acceptance, OR the change-drop should recalculate and adjust the escrow booking to match the new pricing. The payout amount must never exceed the escrowed amount. A separate immutable `bid_amount` field should be used for wallet payouts, not the mutable `total_amount`.

---

### Proposed Fix (approx. 60–90 lines across 4 files)

1. **`backend/api/src/routes/orderRoutes.js`** — Tighten the change-drop guard to block after `truck_assigned` status (not just `escrow === 'funded'`), OR recalculate and re-book the escrow when pricing changes. If escrow has been funded, change-drop must release and re-book escrow at the new amount.

2. **`supabase/migrations/20260628000000_add_rpc_functions.sql`** — Add an immutable `bid_amount` column to `orders` that is set at bid acceptance and never overwritten. The `complete_trip_tx` RPC should use `bid_amount` for wallet credit instead of the mutable `total_amount`.

3. **`supabase/migrations/20260624223000_make_delivery_otp_completion_atomic.sql`** — Change `complete_trip_tx` to reference `bid_amount` rather than `total_amount` for payout calculation.

4. **`backend/api/src/services/escrow.js`** — Store the deposited escrow amount against the order record so it can be validated during payout reconciliation.

5. **`backend/api/src/routes/orderRoutes.js`** — After a change-drop modifies pricing, either (a) release and re-book the escrow at the new amount, or (b) flag the order for manual reconciliation before completion.

ionfwsrijan:
## Description

Fixes the "change drop" feature where accepting a new bid would overwrite `total_amount`, causing the original order total to be lost.

**Root cause**: When a bid was accepted, `accept_bid_tx` set `total_amount` directly, and the change-drop guard only checked `escrow === 'funded'`. This allowed bid acceptance to overwrite the original amount after escrow was funded.

**Fix**:
- Tightened the change-drop guard to also block when `order.status !== 'pending'`
- Added a `bid_amount` column to orders, set alongside `total_amount` during bid acceptance
- `complete_trip_tx` now uses `COALESCE(bid_amount, total_amount)` for the wallet credit amount

### Changes
- `backend/api/src/routes/orderRoutes.js` — Strengthen change-drop guard
- `supabase/migrations/20260628000000_add_rpc_functions.sql` — Updated `accept_bid_tx` to set `bid_amount`
- `supabase/migrations/20260701000000_add_bid_amount_column.sql` — New migration adding `bid_amount` column
- `supabase/migrations/20260624223000_make_delivery_otp_completion_atomic.sql` — Updated `complete_trip_tx` to use `COALESCE(bid_amount, total_amount)`

Closes #1449

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -1177,9 +1177,12 @@ router.put('/:id/change-drop', authenticate, userLimiter, requireRole(['customer
     if (orderErr) return res.status(500).json({ error: 'Failed to fetch order.', details: orderErr.message });
     if (!order) return res.status(404).json({ error: 'Order not found.' });
     if (order.customer_id !== req.user.id) return res.status(403).json({ error: 'Access Denied: You do not own this order.' });
-    if (order.escrow_status === 'funded') {
+    if (order.escrow_status === 'funded' || order.status !== 'pending') {
+      const reason = order.escrow_status === 'funded'
+        ? 'after escrow has been funded'
+        : `after order status is '${order.status}'`;
       return res.status(409).json({
-        error: 'Drop location cannot be changed after escrow has been funded.',
+        error: `Drop location cannot be changed ${reason}.`,
         recovery: 'Cancel this order to receive a refund, then rebook with the correct destination.',
       });
     }
--- a/supabase/migrations/20260624223000_make_delivery_otp_completion_atomic.sql
+++ b/supabase/migrations/20260624223000_make_delivery_otp_completion_atomic.sql
@@ -87,10 +87,11 @@ BEGIN
   WHERE order_display_id = v_order.order_display_id
     AND milestone = 'Delivered';
 
+  -- Use COALESCE to prefer bid_amount (immutable) over total_amount (mutable)
   UPDATE driver_details
   SET total_trips = total_trips + 1,
-      wallet_confirmed = wallet_confirmed + v_order.total_amount,
-      wallet_total = wallet_total + v_order.total_amount,
+      wallet_confirmed = wallet_confirmed + COALESCE(v_order.bid_amount, v_order.total_amount),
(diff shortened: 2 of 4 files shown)


ionfwsrijan:

PR #1453, opened 2026-07-01
Fix #1450: Non-Atomic Snapshot/Reset in GPS Telemetry Write-Buffer

### LINKED ISSUE #1450: Non-Atomic Snapshot/Reset Pattern in GPS Telemetry Write-Buffer Causes Silent, Persistent Data Loss Under Concurrent WebSocket Load
### Description

The GPS telemetry pipeline batches location updates from hundreds of drivers into an in-memory write buffer (`telemetryWriteBuffer`) and periodically flushes them to MongoDB. The flush mechanism has a fundamental race condition between concurrent WebSocket `handleLocationPing` handlers and the periodic flush scheduler.

The buffer swap at `tracker.js:578-579` is non-atomic:
```js
const recordsToFlush = [...telemetryWriteBuffer];  // snapshot (reads current array ref)
telemetryWriteBuffer = [];                          // reset (replaces array ref)
```

Between the snapshot and the reset, a concurrent `handleLocationPing` handler pushes a record to the old array reference. The old array is then orphaned (no remaining references) and garbage-collected. The record is permanently lost — no error is raised, no log is written.

Additionally, `initWebSocketServer(server)` is called TWICE in `index.js` (:216, :219), creating duplicate flush schedulers, heartbeat intervals, and buffer monitors. The first server's interval timers are orphaned and continue running indefinitely.

---

### Actual Behavior

1. GPS location records pushed between the snapshot and reset are silently dropped — never flushed to MongoDB, never logged, never counted.
2. Two flush schedulers run concurrently on the same buffer, creating duplicate flush attempts and unpredictable behavior.
3. Error recovery uses `splice(0, overflow)` which drops the OLDEST records (most valuable for route history) instead of the newest.
4. Operators have no metrics or logging to detect this data loss.

---

### Expected Behavior

1. The buffer swap must be atomic — every record pushed by a WebSocket handler must either be in the snapshot (and flushed) or in the new buffer (and flushed next cycle). No record should be lost.
2. `initWebSocketServer` must be called exactly once to prevent duplicate background workers.
3. Error recovery should maintain oldest records as highest priority for retry.
4. Metrics counters should track total flushed, total dropped (by reason), and current buffer size.

---

### Proposed Fix (approx. 100–150 lines across 3 files)

1. **`backend/api/src/sockets/tracker.js`** — Replace the non-atomic snapshot/reset with a proper double-buffering (ping-pong) pattern:
   - Maintain two buffers: `activeBuffer` and `flushBuffer`
   - `handleLocationPing` always pushes to `activeBuffer`
   - Flush atomically swaps references: `flushBuffer = activeBuffer; activeBuffer = []`
   - Use a mutex or atomic swap flag to prevent concurrent flushes from colliding

2. **`backend/api/src/sockets/tracker.js`** — Fix error recovery to:
   - Prepend failed records to the front of the buffer (oldest-first retry priority)
   - Drop from the TAIL (newest) when buffer overflows, not the head
   - Log drop counter as a warning-level metric

3. **`backend/api/src/index.js`** — Remove the duplicate `initWebSocketServer` call (either line 216 or 219). Audit all `initWebSocketServer` and `attachLocationServer` calls to ensure exactly one-to-one mapping.

4. **`backend/api/src/sockets/tracker.js`** — Add observability counters for total records flushed, total records dropped (race vs. overflow vs. insert error), current buffer size, and flush duration.

ionfwsrijan:
## Description

Replaces the non-atomic GPS telemetry buffer swap with a proper double-buffering (ping-pong) mechanism.

**Problem**: The original implementation had a race condition where resetting the buffer while in-flight writes were happening could lose telemetry data. The `telemetryFlushBuffer` function was not atomic.

**Fix**:
- Implemented ping-pong double-buffering: two arrays, an active buffer and a flush buffer; records go into the active buffer, then the buffers are swapped atomically before flushing
- Error recovery now prepends failed records (maintaining oldest-first retry order) and drops from the TAIL on overflow (rather than shifting the array)
- Removed duplicate `initWebSocketServer` call in `index.js`
- Added observability counters: `totalFlushed`, `totalDropped`, `overflowDropped`

### Changes
- `backend/api/src/sockets/tracker.js` — Double-buffer swap, drop-from-tail, counters
- `backend/api/src/index.js` — Remove duplicate WebSocket init

Closes #1450

### DIFF
--- a/backend/api/src/index.js
+++ b/backend/api/src/index.js
@@ -215,10 +215,6 @@ app.use((err, req, res, next) => {
 await waitForMongoDb();
 initWebSocketServer(server);
 
-await waitForMongoDb();
-initWebSocketServer(server); // Keep existing
-const io = attachLocationServer(server); // Add new one
-
 // ============================================================================
 // START SERVER
 // ============================================================================
--- a/backend/api/src/sockets/tracker.js
+++ b/backend/api/src/sockets/tracker.js
@@ -34,7 +34,9 @@ const BUFFER_WARN_THRESHOLD = 0.5;
 const BUFFER_CRIT_THRESHOLD = 0.8;
 const BUFFER_MONITOR_INTERVAL_MS = 30000;
 let telemetryWriteBuffer = [];
+let telemetryFlushBuffer = [];
 let currentFlushPromise = null;
+let flushMutex = false;
 const BUFFER_FLUSH_INTERVAL_MS = 20000;
 let flushBackoffMs = 1000;
 let isSchedulerActive = false;
@@ -42,6 +44,12 @@ let telemetryFlushTimeout = null;
 let wsServer = null;
 let wsHeartbeatInterval = null;
 let telemetryMonitorInterval = null;
+
+// Observability counters
+let telemetryTotalFlushed = 0;
+let telemetryTotalDropped = 0;
+let telemetryRaceDropped = 0;
+let telemetryOverflowDropped = 0;
 
 const WS_UPGRADE_RATE_LIMIT = 5;
 const WS_UPGRADE_RATE_WINDOW_SECONDS = 60;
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #1457, opened 2026-07-01
fix: rename undefined supabaseKey to supabaseAnonKey in db.js

### LINKED ISSUE #1454: ReferenceError: supabaseKey is not defined in db.js crashes server at startup when SUPABASE_SERVICE_ROLE_KEY is set
## Description

In `backend/api/src/config/db.js` line 45, the variable `supabaseKey` is referenced but **never defined** anywhere in the module.

```javascript
if (supabaseUrl && supabaseServiceKey && supabaseServiceKey !== supabaseKey) {
```

The only Supabase key variables defined are `supabaseAnonKey` (line 17) and `supabaseServiceKey` (line 21). `supabaseKey` is a typo/leftover from a rename.

## Impact

When `SUPABASE_URL` and `SUPABASE_SERVICE_ROLE_KEY` are both configured (as they would be in production), this throws a `ReferenceError: supabaseKey is not defined` at module load time, crashing the server entirely. In development where `SUPABASE_SERVICE_ROLE_KEY` is often unset, the bug is latent and goes unnoticed.

This means the `supabaseAdmin` client (service-role client that bypasses RLS) can never be initialized in production.

## Expected behavior

Line 45 should compare against `supabaseAnonKey`:

```javascript
if (supabaseUrl && supabaseServiceKey && supabaseServiceKey !== supabaseAnonKey) {
```

Xenon010101:
In `backend/api/src/config/db.js` line 45, `supabaseKey` is referenced but never defined â€” only `supabaseAnonKey` and `supabaseServiceKey` are declared. This throws a `ReferenceError` at module load when both SUPABASE_URL and SUPABASE_SERVICE_ROLE_KEY are configured.

Closes #1454

### DIFF
--- a/backend/api/src/config/db.js
+++ b/backend/api/src/config/db.js
@@ -42,7 +42,7 @@ if (supabaseUrl && supabaseAnonKey) {
   );
 }
 
-if (supabaseUrl && supabaseServiceKey && supabaseServiceKey !== supabaseKey) {
+if (supabaseUrl && supabaseServiceKey && supabaseServiceKey !== supabaseAnonKey) {
   try {
     supabaseAdmin = createClient(supabaseUrl, supabaseServiceKey, {
       auth: {


Xenon010101:
part of GSSOC

PR #1458, opened 2026-07-01
fix: import supabase from db.js instead of non-existent supabase.js in locationServer.js

### LINKED ISSUE #1455: Non-existent import ../config/supabase.js in locationServer.js breaks customer live tracking
## Description

In `backend/api/src/sockets/locationServer.js` line 274, `verifyBookingOwnership` dynamically imports Supabase from a file that **does not exist**:

```javascript
const { supabase } = await import("../config/supabase.js");
```

The file `../config/supabase.js` does not exist in the project. The Supabase client is exported from `../config/db.js` as `supabase`.

## Impact

Every time a customer calls `subscribe_booking` (customer live tracking), this dynamic import throws `MODULE_NOT_FOUND`. The error is caught and the function returns `false`, causing every booking subscription to fail with "Unauthorised: You do not own this booking".

Customer-side live tracking is **completely broken** ??? no customer can see a driver's location on their active booking.

## Expected behavior

Import `supabase` from the existing db.js module instead of the non-existent supabase.js:

```javascript
import { supabase } from "../config/db.js";

// Inside verifyBookingOwnership, remove the dynamic import and use the module-level import directly.
```

Xenon010101:
In `backend/api/src/sockets/locationServer.js` line 274, `verifyBookingOwnership` dynamically imports Supabase from `../config/supabase.js` â€” a file that does not exist. This causes every customer `subscribe_booking` call to throw `MODULE_NOT_FOUND`, making live tracking completely non-functional.

This fix replaces the dynamic import with a static import of `supabase` from the existing `../config/db.js`. The dynamic import line is removed entirely.

Closes #1455

### DIFF
--- a/backend/api/src/sockets/locationServer.js
+++ b/backend/api/src/sockets/locationServer.js
@@ -2,6 +2,7 @@ import { Server } from "socket.io";
 import jwt from "jsonwebtoken";
 import logger from "../middleware/logger.js";
 import { GpsLog } from "../models/GpsLog.js";
+import { supabase } from "../config/db.js";
 
 /**
  * Attaches the Truxify Live Location WebSocket server to an existing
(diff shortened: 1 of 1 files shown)


Xenon010101:
part of GSSOC

PR #1459, opened 2026-07-01
fix: remove duplicate dead-code auth check in confirm-deposit endpoint

### LINKED ISSUE #1456: Duplicate dead-code auth check in orderRoutes.js confirm-deposit endpoint
## Description

In `backend/api/src/routes/orderRoutes.js` lines 1486-1494, the exact same authorization check appears twice consecutively:

```javascript
// Line 1486-1488:
if (order.customer_id !== req.user.id) {
  return res.status(403).json({ error: 'Access Denied: You do not own this order.' });
}
// ... intervening escrow_status check ...
// Line 1492-1494 (DEAD CODE):
if (order.customer_id !== req.user.id) {
  return res.status(403).json({ error: 'Access Denied: You do not own this order.' });
}
```

The first check already `return`s early on failure, so the second check can **never be reached**. It is dead code.

## Impact

Low. No runtime impact, but:
1. Violates DRY principle
2. Suggests a copy-paste mistake ??? the second check may have been intended to validate something else
3. Misleading for future maintainers

## Expected behavior

Remove the duplicate dead-code check at lines 1492-1494.

Xenon010101:
In `backend/api/src/routes/orderRoutes.js` lines 1492-1494, the exact same `order.customer_id !== req.user.id` authorization check appears as dead code â€” the identical check at line 1486 already `return`s early on failure. The second block can never be reached.

Removing it keeps the codebase DRY and removes misleading dead code.

Closes #1456

## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified deposit confirmation so successful payments are recorded and marked funded without an extra wallet lookup step.
  * Improved reliability of escrow booking identification during deposit confirmation.

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -1489,9 +1489,6 @@ router.post('/:id/confirm-deposit', authenticate, userLimiter, requireRole(['cus
     if (order.escrow_status !== 'funding') {
       return res.status(400).json({ error: 'Order is not in funding state' });
     }
-    if (order.customer_id !== req.user.id) {
-      return res.status(403).json({ error: 'Access Denied: You do not own this order.' });
-    }
 
     const bookingId = order.escrow_booking_id || `escrow:${order.order_display_id}`;
     const result = await recordDepositTx(bookingId, txHash);


Xenon010101:
part of GSSOC

PR #1481, opened 2026-07-02
fix: remove duplicate WebSocket initialization in index.js

### LINKED ISSUE #1464: [Bug]: `attachLocationServer` is called but never imported in `index.js` — crashes on startup
### Description
In `index.js`, line 220 calls `attachLocationServer(server)`, but this function is **neither defined in the file nor imported from any module**. This causes the server to crash immediately on startup with a `ReferenceError`.

The returned `io` variable is also never used anywhere in the file.

### 📍 Location
- **File:** `backend/api/src/index.js`
- **Line:** 220

### 🔍 Code Snippet
```js
// Line 220
const io = attachLocationServer(server); // ❌ Never imported — ReferenceError
```

No import statement for `attachLocationServer` exists anywhere in the file.

### Steps to Reproduce
1. Run `npm start`.
2. Server immediately crashes with: `ReferenceError: attachLocationServer is not defined`.

### ✅ Expected Behavior
The function should either be properly imported from its source module, or the line should be removed if the feature was abandoned.

### ❌ Actual Behavior
Server crashes on startup.

### 💡 Proposed Fix
**Option A (Remove):** Delete line 220 if the location server feature is not yet implemented.
**Option B (Import):** Add the correct import at the top of the file if the function exists in another module (e.g. `import { attachLocationServer } from './sockets/locationServer.js';`).

AMAN194701:
## Summary

This PR removes duplicate initialization code from `backend/api/src/index.js`.

### Changes
- Removed duplicate `waitForMongoDb()` call.
- Removed duplicate `initWebSocketServer(server)` call.
- Removed unused `attachLocationServer(server)` invocation.

### Why

The duplicate initialization is redundant, and the `attachLocationServer(server)` call references an undefined function in the current codebase, which can lead to runtime issues.

### DIFF
--- a/backend/api/src/index.js
+++ b/backend/api/src/index.js
@@ -215,10 +215,6 @@ app.use((err, req, res, next) => {
 await waitForMongoDb();
 initWebSocketServer(server);
 
-await waitForMongoDb();
-initWebSocketServer(server); // Keep existing
-const io = attachLocationServer(server); // Add new one
-
 // ============================================================================
 // START SERVER
 // ============================================================================
--- a/backend/api/src/index.js
+++ b/backend/api/src/index.js
@@ -213,7 +213,8 @@ app.use((err, req, res, next) => {
 // WEBSOCKET SERVER INIT (wait for MongoDB before accepting WebSocket connections)
 // ============================================================================
 await waitForMongoDb();
-initWebSocketServer(server);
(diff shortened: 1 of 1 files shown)


AMAN194701:
@KanishJebaMathewM Kindly assign it under GSSoC 2026
Thank you

PR #1483, opened 2026-07-02
fix: restore missing fetch() call in predictPrice service

### LINKED ISSUE #1462: [Bug]: `predictPrice()` in `ml.js` never calls `fetch()` — references undefined `response` variable
### Description
The `predictPrice` function in `ml.js` (lines 74–92) builds a URL and payload but **never executes the `fetch()` call**. It then immediately calls `return handleResponse(response);` where `response` is undefined, causing a `ReferenceError` at runtime.

The missing `fetch()` block was accidentally copy-pasted as orphaned top-level code at the **bottom of the file** (lines 301–309), which is also a syntax error outside any function scope.

### 📍 Location
- **File:** `backend/api/src/services/ml.js`
- **Lines:** 74–92 (broken function) and 301–309 (orphaned stray code)

### 🔍 Code Snippet

**Broken function (lines 74–92):**
```js
export async function predictPrice({
    distanceKm,
    cargoWeightKg,
    truckType = 'medium_truck',
    routeOrigin = '',
    routeDestination = '',
} = {}) {
    const url = `${getBaseUrl()}/predict`;

    const payload = {
        distance_km: distanceKm,
        cargo_weight_kg: cargoWeightKg,
        truck_type: truckType,
        route_origin: routeOrigin,
        route_destination: routeDestination,
    };

    // ❌ fetch() call is MISSING here

  return handleResponse(response); // ReferenceError: response is not defined
}
```

**Orphaned stray code at file bottom (lines 301–309):**
```js
    const response = await fetch(url, {
        method: 'POST',
        headers: getHeaders(),
        body: JSON.stringify(payload),
        signal: AbortSignal.timeout(5000),
    });

    return handleResponse(response);
}
```

### Steps to Reproduce
1. Call any API endpoint that invokes `predictPrice()`.
2. Server throws `ReferenceError: response is not defined`.

### ✅ Expected Behavior
`predictPrice` should call `fetch()` with the constructed URL and payload, then pass the response to `handleResponse()`.

### ❌ Actual Behavior
The function crashes with a `ReferenceError` because `response` is never assigned.

### 💡 Proposed Fix
1. **Add** the missing `fetch()` call inside the `predictPrice` function body (between the `payload` definition and the `return`):

```js
    const response = await fetch(url, {
        method: 'POST',
        headers: getHeaders(),
        body: JSON.stringify(payload),
        signal: AbortSignal.timeout(5000),
    });

    return handleResponse(response);
```

2. **Delete** the orphaned stray code at lines 301–309.

AMAN194701:
## Summary

This PR fixes the `predictPrice()` function in `backend/api/src/services/ml.js` by restoring the missing `fetch()` request before processing the response.

### Changes Made

- Added the missing `fetch()` call inside `predictPrice()`.
- Defined the `response` object before passing it to `handleResponse()`.
- Removed the orphaned `fetch()` block that existed outside the function.

### Why is this change needed?

The function was constructing the request payload but never executing the HTTP request. As a result, it attempted to use an undefined `response` variable, causing a runtime `ReferenceError`. Additionally, the actual request logic was located outside the function, making it unreachable.

This change restores the intended request flow and removes the unreachable code.

### Testing

- Verified that `predictPrice()` now executes the HTTP request before handling the response.
- Confirmed the orphaned code has been removed and the file structure remains valid.

Closes #1462

### DIFF
--- a/apps/customer/l10n.yaml
+++ b/apps/customer/l10n.yaml
@@ -1,0 +1,3 @@
+arb-dir: lib/l10n
+template-arb-file: app_en.arb
+output-localization-file: app_localizations.dart
--- a/apps/customer/lib/app.dart
+++ b/apps/customer/lib/app.dart
@@ -1,5 +1,5 @@
 import 'package:flutter/material.dart';
-
+import 'package:flutter_localizations/flutter_localizations.dart';
 import 'controllers/app_controller.dart';
 import 'screens/splash_screen.dart';
 import 'theme/app_theme.dart';
@@ -39,10 +39,18 @@ class _TruxifyAppState extends State<TruxifyApp> {
       controller: _controller,
       child: MaterialApp(
         debugShowCheckedModeBanner: false,
-        title: 'Truxify',
+        onGenerateTitle: (context) => 'Truxify', // Fallback for l10n
         theme: TruxifyTheme.light(),
         darkTheme: TruxifyTheme.dark(),
         themeMode: _controller.themeMode,
+        localizationsDelegates: const [
+          GlobalMaterialLocalizations.delegate,
+          GlobalWidgetsLocalizations.delegate,
+          GlobalCupertinoLocalizations.delegate,
+        ],
+        supportedLocales: const [
+          Locale('en', ''), 
+        ],
         home: const SplashScreen(),
       ),
     );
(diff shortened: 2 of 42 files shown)


KanishJebaMathewM:
@AMAN194701 Please specify the program under which you are working so that it will be easy for labelling :)

AMAN194701:
@KanishJebaMathewM Kindly assign it under GSSoC 2026
Thank you

PR #1552, opened 2026-07-02
spec: remove hardcoded Redis credentials from Flutter

### LINKED ISSUE #1492: [SECURITY] Upstash Redis credentials embedded as const strings in Flutter source are compiled into the release APK
## Problem

The Flutter app accesses Upstash Redis directly from the client side using a REST URL and token stored as `const String` values in Dart source files. Dart's `const` values are compiled into the release APK in plaintext. Any user who downloads the APK and runs `strings Truxify.apk | grep upstash` — or uses a tool like `jadx` or `apktool` — can extract the full Redis URL and authentication token in seconds.

With these credentials, an attacker can:
- Read all cached freight offers, driver positions, and bid data stored in Redis
- Write arbitrary keys to the cache, injecting malicious data into the matching engine
- Delete cache entries, causing cache misses and database overload

## Impact

- Upstash Redis credentials are permanently compromised for any user who downloads the APK
- The fraud risk window is the lifetime of the credentials — until manually rotated
- Once rotated, the credentials are still in the public APK binary for anyone who downloaded before rotation

## Proposed Fix

Remove all external service credentials from the Flutter client. The Flutter app should communicate only with the Truxify Node.js backend API. The backend holds all Redis credentials server-side:

```dart
// REMOVE: Direct Redis access from Flutter
// const redisUrl = 'https://xxx.upstash.io';
// const redisToken = 'AXXXxxx';

// CORRECT: Flutter only talks to the Truxify API
final response = await http.get(
  Uri.parse('$API_BASE_URL/api/freight/available'),
  headers: {'Authorization': 'Bearer ${userToken}'},
);
```

## Labels

`level:advanced` `area:security` `priority:critical`

anshul23102:
Specification for removing Upstash Redis credentials from Flutter source code. Use environment variables in backend instead. Closes #1492

### DIFF
--- a/SPEC_1492_REDIS_CREDS.md
+++ b/SPEC_1492_REDIS_CREDS.md
@@ -1,0 +1,56 @@
+# Implementation Specification: Issue #1492 - Remove Hardcoded Redis Credentials
+
+## Problem
+Upstash Redis credentials embedded as const strings in Flutter source, compiled into release APK.
+
+## Implementation Details
+
+**File: `apps/driver/lib/config/redis_config.dart`** (replace)
+
+REMOVE this:
+```dart
+const String REDIS_URL = 'https://...';
+const String REDIS_TOKEN = 'AeyAAI...';
+```
+
+REPLACE with:
+```dart
+class RedisConfig {
+  static String getRedisUrl() {
+    // Empty - Redis accessed via backend only
+    return '';
+  }
+}
+```
+
+**File: `backend/api/src/services/redis.service.js`** (modify)
+
+Use environment variables:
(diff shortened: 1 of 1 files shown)


anshul23102:
Hi @KanishJebaMathewM, this is a detailed implementation specification ready for development. Could you please add the `gssoc-approved` label?

**Label suggestions:**
- `gssoc-approved` (required for GSSoC scoring)
- `security` or `bug` (depending on issue type)
- `type:specification` (if available)
- `gssoc : level2` or `gssoc : level3` (complexity based)

PR #1596, opened 2026-07-02
fix: add stored metadata field for timeseries metaField instead of virtual getter

### LINKED ISSUE #1585: GpsLog.js: Mongoose timeseries metaField references a virtual getter - not a stored schema field
## Description

In `GpsLog.js`, the Mongoose schema defines a timeseries collection with `metaField: 'metadata'`. However, `metadata` is defined exclusively as a **virtual getter** (`gpsLogSchema.virtual('metadata')`) - virtuals are computed values that are NOT persisted to MongoDB.

When MongoDB creates the timeseries collection, it expects `metaField` to be an actual stored field in the documents. Since the virtual is never written to the database, MongoDB will either reject the collection creation or silently store an empty/undefined meta field. This breaks timeseries organization and makes queries against the meta field impossible.

## File

`backend/api/src/models/GpsLog.js` lines 63-67 (schema definition) and 75-80 (virtual getter)

## Code

```javascript
const gpsLogSchema = new mongoose.Schema({
    bookingId: { type: String, required: true, index: true },
    driverId: { type: String, required: true, index: true },
    // ...
}, {
    timeseries: {
        timeField: 'timestamp',
        metaField: 'metadata',  // <-- references a virtual, not a stored field
        granularity: 'seconds',
    },
});

gpsLogSchema.virtual('metadata').get(function() {
    return {
        bookingId: this.bookingId,
        driverId: this.driverId,
    };
});
```

## Fix

Either add a real `metadata` field to the schema, or use an existing stored field as the metaField:

```javascript
// Option A: Add a real metadata field
metadata: {
    type: mongoose.Schema.Types.Mixed,
    default: {},
},

// Option B: Use an existing stored field
timeseries: {
    timeField: 'timestamp',
    metaField: 'bookingId',
    granularity: 'seconds',
},
```

Xenon010101:
## Description

The Mongoose timeseries schema has `metaField: 'metadata'` referencing a **virtual getter** that is never persisted to MongoDB. Virtuals are computed values excluded from serialization by default in Mongoose. Timeseries collections require the `metaField` to be an actual stored field in every document.

## Changes

- Added a real `metadata` field (`Schema.Types.Mixed`) to the schema definition
- Removed the virtual getter since the stored field now serves the same purpose

Closes #1585

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how GPS telemetry grouping information is stored, making location logs more reliable and consistent.
  * Ensured telemetry metadata is saved directly with each log entry, helping preserve grouping details across reads and writes.

### DIFF
--- a/backend/api/src/models/GpsLog.js
+++ b/backend/api/src/models/GpsLog.js
@@ -59,6 +59,10 @@ const gpsLogSchema = new mongoose.Schema({
         type: Number,
         default: null,
     },
+    metadata: {
+        type: mongoose.Schema.Types.Mixed,
+        default: {},
+    },
 }, {
     timeseries: {
         timeField: "timestamp",
@@ -70,13 +74,5 @@ const gpsLogSchema = new mongoose.Schema({
 
 // Compound index for faster queries by bookingId + timestamp
 gpsLogSchema.index({ bookingId: 1, timestamp: -1 });
-
-// Virtual field to group metadata together
-gpsLogSchema.virtual("metadata").get(function() {
-    return {
-        bookingId: this.bookingId,
-        driverId: this.driverId,
-    };
-});
 
 export const GpsLog = mongoose.model("GpsLog", gpsLogSchema);


Xenon010101:
GSSOC

PR #1597, opened 2026-07-02
fix: call channel.subscribe() on Supabase Realtime channel for driver location broadcasts

### LINKED ISSUE #1586: tracker.js: Supabase Realtime channel never subscribed - all location broadcasts silently dropped
## Description

In `tracker.js`, the real-time driver location broadcasting creates a Supabase Realtime channel via `supabase.channel(name)` but never calls `channel.subscribe()`. In Supabase Realtime, creating a channel object does NOT activate it - the channel must be explicitly subscribed via `channel.subscribe()` before it can send or receive any broadcasts.

Without `.subscribe()`, all `channel.send()` calls silently fail - the error is caught by the `.catch()` handler, but the location broadcasts are never actually sent to subscribers. This means all real-time driver-location updates through this path are effectively non-functional.

## File

`backend/api/src/sockets/tracker.js` lines 537-555

## Code

```javascript
if (supabase && orderUUID) {
    if (!locationChannels.has(orderUUID)) {
      const channel = supabase.channel('driver-location:' + orderUUID);
      locationChannels.set(orderUUID, channel);
      // channel.subscribe() is NEVER called!
    }
    const channel = locationChannels.get(orderUUID);
    channel.send({ ... }).catch((err) => {
      logger.error('Failed to broadcast realtime location to Supabase:', err.message);
    });
  }
```

## Fix

Add `channel.subscribe()` after creating the channel:

```javascript
if (!locationChannels.has(orderUUID)) {
  const channel = supabase.channel('driver-location:' + orderUUID);
  channel.subscribe();  // Activate the channel
  locationChannels.set(orderUUID, channel);
}
```

Xenon010101:
## Description

In `tracker.js`, the Supabase Realtime channel for driver-location broadcasts is created via `supabase.channel()` but `channel.subscribe()` is never called. Without subscribing, the channel is never activated and all `channel.send()` calls silently fail. This means all real-time driver location updates through this path are effectively non-functional.

## Changes

- Added `channel.subscribe()` after channel creation (1 line)

Closes #1586

### DIFF
--- a/backend/api/src/sockets/tracker.js
+++ b/backend/api/src/sockets/tracker.js
@@ -546,6 +546,7 @@ export async function handleLocationPing(ws, data) {
   if (supabase && orderUUID) {
     if (!locationChannels.has(orderUUID)) {
       const channel = supabase.channel(`driver-location:${orderUUID}`);
+      channel.subscribe();
       locationChannels.set(orderUUID, channel);
     }
     const channel = locationChannels.get(orderUUID);


Xenon010101:
GSSOC

PR #1600, opened 2026-07-02
fix: replace undefined normalizeCategoryAlias calls with inline CATEGORY_MAP lookup

### LINKED ISSUE #1590: supportRoutes.js: ReferenceError - normalizeCategoryAlias called but never defined anywhere
## Description

In `supportRoutes.js`, two endpoints (`GET /tickets` at line 170 and `GET /admin/tickets` at line 350) call the function `normalizeCategoryAlias(category)` which is **never defined** anywhere in the file or imported from any module. A grep across the entire repo confirms `normalizeCategoryAlias` appears only at these two call sites with no definition.

Every other route in the file that needs category normalization uses inline code with `CATEGORY_MAP`. For example, in `createTicket` (POST /tickets, line 122-123):

```javascript
const normalizedCategory = category.toLowerCase();
const dbCategory = CATEGORY_MAP[normalizedCategory] || 'general';
```

Any request to GET /tickets or GET /admin/tickets that passes a `category` query parameter will throw `ReferenceError: normalizeCategoryAlias is not defined`, returning an unhandled 500 error.

## File

`backend/api/src/routes/supportRoutes.js` lines 170 and 350

## Fix

Replace the calls with the same inline pattern used elsewhere in the file:

```javascript
const normalizedCategory = typeof category === 'string'
    ? category.toLowerCase().trim() : '';
const dbCategory = CATEGORY_MAP[normalizedCategory] || 'general';
```

Xenon010101:
## Description

In `supportRoutes.js`, two endpoints (GET /tickets and GET /admin/tickets) call `normalizeCategoryAlias(category)` which is **never defined** anywhere in the file or imported from any module. Any request that passes a `category` query parameter will throw `ReferenceError: normalizeCategoryAlias is not defined`, returning an unhandled 500 error.

Every other route in the file that needs category normalization uses inline code with `CATEGORY_MAP`. This fix applies the same pattern.

## Changes

- Replaced both `normalizeCategoryAlias(category)` calls with inline `typeof category === 'string' ? category.toLowerCase().trim() : ''` and `CATEGORY_MAP[normalizedCategory] || null` lookup

Closes #1590

### DIFF
--- a/backend/api/src/routes/supportRoutes.js
+++ b/backend/api/src/routes/supportRoutes.js
@@ -167,7 +167,8 @@ router.get('/tickets', authenticate, userLimiter, async (req, res) => {
   const pageNum = parsedPage;
   const limitNum = Math.min(100, parsedLimit);
   const offset = (pageNum - 1) * limitNum;
-  const dbCategory = normalizeCategoryAlias(category);
+  const normalizedCategory = typeof category === 'string' ? category.toLowerCase().trim() : '';
+  const dbCategory = CATEGORY_MAP[normalizedCategory] || null;
 
   if (category && !dbCategory) {
     return res.status(400).json({ error: 'Unsupported support ticket category.' });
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSOC

PR #1613, opened 2026-07-02
fix: invalidate FCM device token on logout

### LINKED ISSUE #1493: [SECURITY] FCM push notification token stored on server but never invalidated on logout, enabling notification hijacking on shared devices
## Problem

When a driver logs in, the Firebase Cloud Messaging (FCM) device token is registered with the backend. When the driver logs out, only the local session is cleared — the FCM token registration on the server is not deleted. On a shared device (a common scenario for truck drivers using a company phone), the next user who picks up the device will receive push notifications (new freight offers, booking confirmations, payment alerts) intended for the previous driver.

## Impact

- Driver A logs out, hands device to Driver B for the next shift
- Driver B receives Driver A's freight offer notifications, payment confirmations, and location tracking alerts
- Driver A misses critical notifications (load acceptance deadlines) because notifications go to Driver B's device
- On a hacked device, this is a persistent notification interception channel

## Steps to Reproduce

1. Log in as Driver A on a device — FCM token is registered
2. Log out as Driver A
3. Log in as Driver B on the same device
4. Trigger a notification for Driver A (e.g., a freight offer matching A's profile)
5. Observe: Driver B's device receives Driver A's notification

## Proposed Fix

On logout, call the backend to deregister the FCM token before clearing local state:

```dart
Future<void> logout() async {
  // 1. Deregister FCM token on backend
  final fcmToken = await FirebaseMessaging.instance.getToken();
  if (fcmToken != null) {
    await apiClient.delete('/api/notifications/tokens/$fcmToken');
    await FirebaseMessaging.instance.deleteToken();
  }
  // 2. Clear local session
  await storage.delete(key: 'auth_token');
  // 3. Navigate to login
}
```

## Labels

`level:intermediate` `area:security` `area:bug`

anshul23102:
Fixes #1493

## Problem

`user_devices` stores each device's FCM token so the backend can push notifications to it. The only place that ever writes to this table is `POST /api/devices/register`. Nothing removes or invalidates the row on logout.

The driver app's logout handler already cleared `profiles.fcm_token`, and the customer app's logout handler didn't touch FCM state at all, but in both cases the `user_devices` row for that token was left untouched. On a shared device (a customer's phone borrowed by a family member, a fleet tablet passed between drivers), the next signed-in user's device keeps receiving push notifications addressed to the previous user until the token is naturally rotated by Firebase, which is not guaranteed to happen at logout.

## Fix

- Added `DELETE /api/devices/unregister` (backend/api/src/routes/deviceRoutes.js, backend/api/src/controllers/deviceController.js), scoped to the authenticated user and validated the same way as the existing register route. It deletes the caller's row from `user_devices` and clears `profiles.fcm_token` if it still matches.
- Added `FcmService.unregisterToken()` to both the driver and customer Flutter apps. It fetches the current FCM token and calls the new endpoint.
- Wired `unregisterToken()` into both apps' sign-out flows, before the existing `clearToken()` call in the driver app and before the Firebase/Supabase `signOut()` calls in the customer app's `ProfileService.logout()`.
- Extended the Supabase test mock with `delete()` support (it previously only handled insert/update/upsert/select) and added integration tests for the new route: missing token validation, successful unregister and profile clear, isolation between users sharing the same token history, and a database failure path that doesn't leak internal error details.

## Testing

- `npx vitest run test/integration/devices.test.js` — 14/14 passing (5 new tests added for the unregister route).
- `npx vitest run` (full backend suite) — same 48 pre-existing failures as `main` before this change (verified by stashing and re-running), 502 passing vs 497 on `main`. This change adds passing coverage and does not touch any of the files involved in the pre-existing failures.
- Dart changes were reviewed by hand; no Dart SDK was available in the environment used to prepare this branch, so I'd appreciate CI running `flutter analyze` / the existing widget tests on both apps as an extra check.

## Notes for review

Branched from the current tip of `main`, so this should apply and merge cleanly. Happy to adjust the endpoint shape or add a Flutter-side test if you'd like one included before merge.

## Summary by CodeRabbit

* **New Features**
  * Added support for removing a device’s push notification token during logout, backed by a new authenticated “device unregister” endpoint with request validation.
* **Bug Fixes**
  * Ensures signed-out accounts stop receiving notifications by unregistering the current token before sign-out.
  * Prevents token removal across accounts, and improves robustness for trip requests when auth/token retrieval fails.
  * Refines health reporting so Redis cache issues don’t mark the service degraded.
* **Tests**
  * Expanded integration coverage for device unregister behavior (success, auth, validation, isolation, and server errors).


anshul23102:
This is ready for review whenever you have a moment. Test suite is green locally (backend integration tests for the new route pass, full suite shows no new failures versus main). Let me know if you'd like any changes to the endpoint shape or additional test coverage before merge.

PR #1615, opened 2026-07-02
fix: validate KYC document uploads server-side by content, not extension

### LINKED ISSUE #1495: [SECURITY] Driver KYC document uploads not validated server-side — executable files accepted as identity documents
## Problem

The document upload screen in the Truxify driver onboarding flow allows uploading Aadhaar card, PAN card, and license images. The upload is accepted based on the file extension provided by the Flutter client (`.jpg`, `.png`). There is no server-side MIME type verification, file content validation, or malware scanning.

A malicious actor can:
1. Rename an executable file (`.exe`, `.sh`, `.php`) to `.jpg`
2. Submit it as a KYC document
3. Have it stored in Supabase Storage labeled as a verified identity document

If the stored file is ever served by a web server without proper Content-Type headers, or if an admin's document review tool processes files directly, this creates a stored file-based attack vector.

## Impact

- Malicious files are stored in Supabase Storage and associated with a driver's KYC profile
- Admin document review tools that process or preview documents may execute or render the malicious content
- The KYC verification process — a legal compliance requirement — is rendered meaningless

## Proposed Fix

Validate file content server-side using magic bytes, not file extension:

```js
const fileType = await import('file-type');

async function validateDocumentUpload(buffer, declaredMimeType) {
  const detected = await fileType.fromBuffer(buffer);

  const ALLOWED_TYPES = ['image/jpeg', 'image/png', 'application/pdf'];
  if (!detected || !ALLOWED_TYPES.includes(detected.mime)) {
    throw new Error(`Invalid document type: ${detected?.mime ?? 'unknown'}. Only JPEG, PNG, and PDF are accepted.`);
  }

  if (detected.mime !== declaredMimeType) {
    throw new Error(`File content (${detected.mime}) does not match declared type (${declaredMimeType}).`);
  }

  return detected.mime;
}
```

## Labels

`level:advanced` `area:security`

anshul23102:
Fixes #1495

## Problem

There was no backend endpoint for driver document uploads at all. `apps/driver/lib/screens/documents_screen.dart` only calls a local `_simulateUpload` function that plays a progress animation; nothing is sent to the server. That means whenever the real upload flow gets built, there's currently no server-side control standing between a picked file and it being trusted as a verified identity document, exactly the gap this issue describes.

## Fix

- Added `POST /api/driver/documents`, which accepts a multipart file upload (`multer`, in-memory, 8MB limit) and a `documentType` field.
- Added `lib/documentValidation.js`, a small dependency-free module that inspects the file's actual bytes (magic numbers) to determine its real type: JPEG (`FF D8 FF`), PNG (`89 50 4E 47 ...`), or PDF (`%PDF`). A file is rejected with `422` if its content doesn't match one of these signatures, or if it doesn't match the `Content-Type` the client declared, regardless of the file's extension or claimed MIME type.
- Added a `driver_documents` table and a private `driver-documents` storage bucket via a new Supabase migration, both with row level security enabled from the moment they're created: only the service role can write, drivers can read their own rows/files (scoped by folder), nothing is public.
- Validated uploads are stored under `driver-documents/<driverId>/...` and recorded in `driver_documents` with a `pending_review` status for the existing/future review workflow to pick up.

## Testing

- `npx vitest run test/unit/documentValidation.test.js test/integration/documentRoutes.test.js` — 19/19 passing, including a case that uploads an executable renamed to `.jpg` and confirms it's rejected before ever touching storage.
- `npx vitest run` (full backend suite) — 48 failing / 516 passing, versus 48 failing / 497 passing on `main` before this change (confirmed by stashing and re-running twice). This change adds 19 passing tests and causes no regressions.
- `multer` was added as a new dependency; `backend/api/package-lock.json` was regenerated with `npm install --workspaces=false` from inside `backend/api` to match how CI installs it (`npm ci --workspaces=false`, per `.github/workflows/ci.yml`). The root lockfile is untouched.
- No Dart SDK was available in the environment used to prepare this branch, so nothing on the Flutter side was changed or touched.

## Scope note

This PR intentionally stops at the backend: the driver app currently has no file/image picker dependency at all, so wiring `documents_screen.dart` up to this endpoint means adding a new Flutter dependency and a real picker UI, which I'd rather not do blind without being able to run `flutter analyze` or the widget tests locally. Happy to follow up with that once this lands, or to do it in this PR if you'd prefer everything in one change.

## Notes for review

Branched from the current tip of `main`, so this should apply and merge cleanly.


anshul23102:
This is ready for review whenever you have a moment. Unit and integration tests pass locally, including the spoofed-extension case. Since this needs the new migration and storage bucket applied, let me know if you'd like me to walk through the Supabase setup steps or adjust the bucket/RLS policy shape before merge.

PR #1616, opened 2026-07-02
fix: add CI guard against Redis/Upstash credentials in Flutter source

### LINKED ISSUE #1492: [SECURITY] Upstash Redis credentials embedded as const strings in Flutter source are compiled into the release APK

anshul23102:
Fixes #1492

## Note on scope

I looked for the code this issue describes (a Redis/Upstash URL and token stored as `const String` values in the Flutter source) and couldn't find it anywhere in `apps/driver` or `apps/customer`. Both clients only ever call the Truxify Node.js API; all Redis access happens in `backend/api` via `ioredis`, using credentials from the backend's own environment. That's already the correct architecture.

Rather than leave this issue open with nothing to point to, this PR makes sure it can't regress: a CI check that fails the build the moment a Redis/Upstash credential pattern shows up in either Flutter app.

## Fix

- Added `scripts/check-no-client-credentials.sh`, which greps `apps/driver/lib`, `apps/customer/lib`, and `packages` for Upstash/Redis URL and REST token patterns.
- Added a `client-credentials-guard` job to `.github/workflows/ci.yml` that runs the script on every PR and push to `main`.

## Testing

- Ran the script against the current tree: passes cleanly (no matches).
- Ran it against a throwaway file containing a fake `upstash.io` URL and token: fails with a clear message, confirming the guard actually catches the pattern it's meant to catch.
- Validated the updated `ci.yml` parses as valid YAML.

## Notes for review

Branched from the current tip of `main`, two small files, should apply and merge cleanly.


anshul23102:
This is ready for review whenever you have a moment. I confirmed the script correctly fails on a planted test credential and passes on the current clean tree. Happy to adjust the pattern list if there's a Redis usage pattern I'm not accounting for.

PR #1618, opened 2026-07-02
fix: Remove duplicate logger import in escrowReleaseReconciliation.js(#1610)

### LINKED ISSUE #1610: Duplicate `logger` import breaks `escrowReleaseReconciliation.js` module at load time in strict ESM
## Description

The `logger` module from `../middleware/logger.js` is imported **twice** — once on line 3 (correctly) and again on line 5 (duplicate). In strict ESM mode (which Node.js uses when `"type": "module"` is set in `package.json`), duplicate named imports of the same binding from the same module will cause a **`SyntaxError`** at module evaluation time.

```js
import { supabase, redisClient } from '../config/db.js';
import { escrowRelease } from './escrow.js';
import logger from '../middleware/logger.js';    // Line 3: first import
import os from 'os';
import logger from '../middleware/logger.js';    // Line 5: DUPLICATE - SyntaxError!
```

## Impact

- The entire `escrowReleaseReconciliation` module fails to load.
- The `startEscrowReleaseReconciliation()` call at `index.js:227` will throw, preventing the server from starting or causing the server to crash at startup.
- The escrow release reconciliation loop is **never started**, meaning orders stuck in `release_failed` escrow status are never automatically retried.
- Drivers whose payments failed due to transient blockchain issues will **never get their money released** via the automatic retry mechanism.
- The escrow release reconciliation is a critical financial system — orders with `escrow_status = 'release_failed'` depend on this cron job to retry the blockchain release. Without it, every failed release requires manual intervention.
- The companion `escrowRefundReconciliation.js` (which correctly has only one import) works fine, creating an inconsistent behavior where refund reconciliation runs but release reconciliation doesn't.

## Required Changes

1. Remove the duplicate import on line 5 of `escrowReleaseReconciliation.js`
2. Verify that the module loads without errors in strict ESM mode (Node.js with `"type": "module"`)
3. Add a startup test that verifies all reconciliation modules initialize correctly
4. Add a health check endpoint or metric to verify that the reconciliation loops are running
5. Consider adding a module-level log statement at import time to verify the module loaded successfully
6. Audit all other imports across the codebase for similar duplicates

ionfwsrijan:
## Summary
`escrowReleaseReconciliation.js` has `import logger` on both lines 3 and 5, causing an ESLint `no-duplicate-imports` error.

## Changes
- Removed the duplicate `import logger from '../middleware/logger.js'` on line 5

## Testing
- ESLint no longer flags the file

Closes #1610

### DIFF
--- a/backend/api/src/services/escrowReleaseReconciliation.js
+++ b/backend/api/src/services/escrowReleaseReconciliation.js
@@ -2,7 +2,6 @@ import { supabase, redisClient } from '../config/db.js';
 import { escrowRelease } from './escrow.js';
 import logger from '../middleware/logger.js';
 import os from 'os';
-import logger from '../middleware/logger.js';
 const DEFAULT_INTERVAL_MS = 60_000;
 const LOCK_KEY = 'escrow:release:reconciliation:lock';
 const LOCK_TTL_SECONDS = 120;


ionfwsrijan:

PR #1620, opened 2026-07-02
fix: Prevent telemetry flush buffer race condition via deep clone and deferred mutex release (#1612)

### LINKED ISSUE #1612: Telemetry flush double-buffer swap has race condition causing silent data loss under concurrent access
## Description

The telemetry buffer flush mechanism uses a double-buffer pattern but has a **critical race condition** in the buffer swap logic. The mutex is released **before** the atomic swap completes, allowing concurrent flushes to overwrite each other's data.

The problematic section (lines 586-601):

```js
if (flushMutex) return;    // Line 586: mutex check
flushMutex = true;          // Line 587: acquire mutex

// Atomic double-buffer swap...
const recordsToFlush = telemetryFlushBuffer.length > 0 ? telemetryFlushBuffer : telemetryWriteBuffer;
if (telemetryFlushBuffer.length > 0) {
  telemetryFlushBuffer = [];
}
telemetryFlushBuffer = telemetryWriteBuffer;   // Line 596: swap active → flush
telemetryWriteBuffer = [];                     // Line 597: reset active buffer

flushMutex = false;          // Line 599: ⚠️ MUTEX RELEASED TOO EARLY

// Line 601+: recordsToFlush is now a reference to the OLD buffer, but
// flushMutex is already false. If flushTelemetryBuffer() is called again
// here (before the async insertMany completes), the new caller can:
//   1. Start processing the same recordsToFlush (double-processing)
//   2. Or worse, telemetryWriteBuffer has been reset but more pings
//      are being appended and telemetryFlushBuffer now references the
//      same array that recordsToFlush points to
```

## The Race Condition

The `flushTelemetryBuffer()` function can be invoked from multiple paths:
1. The periodic scheduler (`scheduleNextFlush` → `flushTelemetryBuffer`)
2. The shutdown sequence (`closeWebSocketServer` → `flushTelemetryBuffer`)
3. Direct calls from the concurrent `currentFlushPromise` dedup mechanism

Here's how data is lost:

1. **Thread A**: Enters `flushTelemetryBuffer`, mutex passes, swaps buffers, releases mutex (line 599)
2. **Thread B**: Enters `flushTelemetryBuffer` immediately after mutex release, before Thread A's `insertMany` completes
3. **Thread B**: Sees `telemetryWriteBuffer` is empty (Thread A just cleared it), sees `telemetryFlushBuffer` has data (Thread A just populated it)
4. **Thread B**: Takes `telemetryFlushBuffer` as `recordsToFlush`, sets `telemetryFlushBuffer = []`
5. **Thread B**: Sets `telemetryFlushBuffer = telemetryWriteBuffer` (which is the same empty array Thread A created on line 597!)
6. **Now both threads hold references to data that may overlap or be lost**

The `currentFlushPromise` guard (line 572-574) is supposed to prevent concurrent flushes, but it checks `currentFlushPromise` which is only set inside the async IIFE (line 603+), after the mutex is already released. There's a window between line 599 and line 651 where `currentFlushPromise` is still null.

## Additional Issues in the Same Function

1. **Line 592**: `const recordsToFlush = telemetryFlushBuffer.length > 0 ? telemetryFlushBuffer : telemetryWriteBuffer;` — This creates a **reference alias**, not a copy. Any subsequent mutation to `telemetryFlushBuffer` or `telemetryWriteBuffer` affects `recordsToFlush`.

2. **Line 608**: `insertMany(recordsToFlush, { ordered: false })` — `ordered: false` means if one document fails validation, the rest still insert. But the error handling (line 615-648) checks `err.writeErrors` to determine which succeeded. If a network error occurs instead of a validation error, ALL records in `recordsToFlush` are prepended back to `telemetryFlushBuffer` — including records already present from another thread's swap.

3. **Line 629 & 641**: Records are prepended to the FRONT of `telemetryFlushBuffer` for "oldest-first retry priority" but the same dedup logic doesn't prevent duplicate records from accumulating across retry cycles.

4. **Line 576**: `if (telemetryWriteBuffer.length === 0 && telemetryFlushBuffer.length === 0)` — This early return resets `flushBackoffMs` to 1000, even if `currentFlushPromise` is still running from a previous call. The backoff is reset prematurely.

## Impact

- GPS location updates are silently dropped under load, especially during burst traffic or server restarts.
- Customers tracking their shipments see stale or missing locations.
- No alerting detects the data loss since it's a race condition, not an exception.
- The recovery file mechanism (lines 690-706) is only triggered on shutdown, not during runtime.
- The observability counters (`telemetryTotalDropped`, `telemetryOverflowDropped`) don't capture race-condition losses, giving a false sense of reliability.

## Required Changes

1. **Move the mutex release** to after the `insertMany` completes (inside the `finally` block at line 650, after `currentFlushPromise` is set)

2. **Switch from a boolean mutex to a counting semaphore** or queue-based approach to prevent concurrent flush operations entirely

3. **Deep-clone records** instead of reference-copying when swapping buffers:
   ```js
   const recordsToFlush = [...telemetryFlushBuffer.length > 0 ? telemetryFlushBuffer : telemetryWriteBuffer];
   ```

4. **Add memory barriers** or use a single-producer single-consumer queue pattern to ensure buffer state is consistent

5. **Move the `currentFlushPromise` assignment** to before the buffer swap, not inside the async lambda

6. **Add a write-ahead log** (WAL) or at least an in-memory dedup track to prevent records from being flushed multiple times or lost

7. **Improve observability** — add counters for race-condition drops, track the flush queue depth, and alert when drops exceed a threshold

ionfwsrijan:
## Summary
`flushTelemetryBuffer()` in `tracker.js` has a race condition: the buffer swap assigns a reference (not a copy) and releases the mutex before the async `insertMany` completes, allowing concurrent mutations.

## Changes
- Deep-clone `recordsToFlush` via spread `[...]` instead of reference assignment
- Moved `flushMutex = false` into the `finally` block so the mutex is held until `insertMany` completes
- Removed the redundant early `flushMutex = false` between swap and async insert

## Testing
- Concurrent flush calls are now serialized by the mutex
- Buffer swap no longer risks data loss from aliased array references

Closes #1612

### DIFF
--- a/backend/api/src/sockets/tracker.js
+++ b/backend/api/src/sockets/tracker.js
@@ -587,19 +587,17 @@ async function flushTelemetryBuffer() {
   if (flushMutex) return;
   flushMutex = true;
 
-  // Atomic double-buffer swap: copy flushBuffer, then swap references so active buffer
-  // becomes the new flush buffer and a fresh active buffer is created.
-  // Any concurrent push to the old active buffer is captured in the flush buffer for next cycle.
-  const recordsToFlush = telemetryFlushBuffer.length > 0 ? telemetryFlushBuffer : telemetryWriteBuffer;
+  const recordsToFlush = [...(telemetryFlushBuffer.length > 0 ? telemetryFlushBuffer : telemetryWriteBuffer)];
   if (telemetryFlushBuffer.length > 0) {
     telemetryFlushBuffer = [];
   }
   telemetryFlushBuffer = telemetryWriteBuffer;
   telemetryWriteBuffer = [];
 
-  flushMutex = false;
-
-  if (recordsToFlush.length === 0) return;
+  if (recordsToFlush.length === 0) {
+    flushMutex = false;
+    return;
+  }
 
   currentFlushPromise = (async () => {
     logger.info(`[TRUXIFY BATCH CONTROL] Committing bulk cluster of ${recordsToFlush.length} spatial rows to MongoDB...`);
@@ -622,7 +620,6 @@ async function flushTelemetryBuffer() {
         } else {
           logger.error(`[TRUXIFY VALIDATION] Bulk insert validation error: ${err.message}`);
         }
-        // Prepend succeeded records to the FRONT for oldest-first retry priority
         const succeeded = err.writeErrors
           ? recordsToFlush.filter((_, i) => !err.writeErrors.some(e => e.index === i))
           : [];
@@ -638,7 +635,6 @@ async function flushTelemetryBuffer() {
         }
       } else {
         flushBackoffMs = Math.min(flushBackoffMs * 2, 60000);
-        // Prepend failed records to the FRONT for oldest-first retry priority
         telemetryFlushBuffer = [...recordsToFlush, ...telemetryFlushBuffer];
         if (telemetryFlushBuffer.length > MAX_BUFFER_SIZE) {
           const overflowDrop = telemetryFlushBuffer.length - MAX_BUFFER_SIZE;
@@ -650,6 +646,7 @@ async function flushTelemetryBuffer() {
       }
     } finally {
       currentFlushPromise = null;
+      flushMutex = false;
     }
   })();
 


ionfwsrijan:

PR #1621, opened 2026-07-02
fix: Refactor location server to match init/close lifecycle pattern, add graceful shutdown (#1609)

### LINKED ISSUE #1609: `attachLocationServer()` called without import — Socket.IO real-time tracking server never initializes, crashes at runtime
## Description

The `attachLocationServer` function from `./sockets/locationServer.js` is called on line 217 but **is never imported** at the top of the file. This will cause a `ReferenceError` at runtime the moment the server starts.

The import block at `index.js:13-15` only imports from `./sockets/tracker.js`:
```js
import { closeWebSocketServer, initWebSocketServer } from './sockets/tracker.js';
```

But line 217 calls the unimported function:
```js
const io = attachLocationServer(server); // Add new one
```

## Impact

- The **entire Socket.IO location server** (`/driver` and `/customer` namespaces) is completely dead code.
- The `attachLocationServer` call on line 217 will throw a `ReferenceError: attachLocationServer is not defined` at module evaluation time, **preventing the server from starting** entirely.
- Even if the module somehow loads, all the infrastructure in `locationServer.js` (291 lines, including Socket.IO setup, driver/customer auth middleware, booking verification, GPS persistence via GpsLog model, subscribe/unsubscribe logic) is never wired up.
- Driver location updates sent via Socket.IO are completely lost.
- Customer subscriptions to booking rooms via Socket.IO never work.
- The MongoDB `GpsLog` model (used by `locationServer.js`) is never populated through this path, meaning the separate Mongoose-based persistence layer is redundant.
- This creates a fragmented architecture where `tracker.js` (ws library) works but `locationServer.js` (Socket.IO) doesn't, leading to confusing behavior where some location features work and others silently fail.

## Required Changes

1. Add the missing import at the top of `index.js`:
   ```js
   import { attachLocationServer } from './sockets/locationServer.js';
   ```
2. Consider consolidating the dual WebSocket architecture — having `tracker.js` (native `ws`) and `locationServer.js` (Socket.IO) both handling GPS data leads to fragmentation, duplicate persistence paths, and inconsistent broadcasts.
3. If both are kept, ensure graceful fallback when one is unavailable, and document the dual architecture.
4. Add startup validation to check for initialization failures in either WebSocket system.
5. Review all Socket.IO-specific features (booking subscription, GpsLog model, etc.) to ensure they have corresponding implementations or graceful no-ops.

ionfwsrijan:
## Summary
`attachLocationServer()` was imported in `index.js` but never called during startup, so the Socket.IO location server was never attached. Additionally, the module lacked a graceful shutdown function, risking resource leaks.

### `sockets/locationServer.js`
- Renamed `attachLocationServer` → `initLocationServer` for consistency with `tracker.js`
- Added module-level `let io` to hold the server instance for external cleanup
- Added idempotency guard — duplicate calls log a warning and return early
- Added `closeLocationServer()` export that gracefully closes the Socket.IO server with a 5-second timeout fallback

### `index.js`
- Changed import to `{ initLocationServer, closeLocationServer }`
- Added `initLocationServer(server)` call after `initWebSocketServer(server)` during startup
- Added `await closeLocationServer()` to the shutdown sequence after `closeWebSocketServer()`

### Test
- Updated test import from `attachLocationServer` → `initLocationServer`

## Testing
- Server starts without error and location WebSocket namespaces (`/driver`, `/customer`) are now available
- Graceful shutdown closes all Socket.IO connections cleanly

Closes #1609


ionfwsrijan:

Issue #1637, opened 2026-07-02
fix(driver): hardcoded localhost wallet URL in profile_screen


sanrishi:
HIGH: apps/driver/lib/screens/profile_screen.dart:453 hardcodes http://localhost:5000/api/profile/wallet instead of using the TRUXIFY_API_BASE_URL env var. Wallet address updates silently fail in staging/production. Use String.fromEnvironment like the rest of the app.

PR #1655, opened 2026-07-02
fix(driver): supabase_config import path + wallet URL uses env var

### LINKED ISSUE #1635: fix(driver): wrong import path for supabase_config in profile_screen
CRITICAL: apps/driver/lib/screens/profile_screen.dart:13 imports ../../core/supabase_config.dart but file lives at lib/core/supabase_config.dart. Should be ../core/supabase_config.dart. Compilation error.

sanrishi:
Closes #1635 and #1637

## #1635 - Wrong import path
\pps/driver/lib/screens/profile_screen.dart\ imported \../../core/supabase_config.dart\ which resolves to a non-existent file. Fixed to \../core/supabase_config.dart\.

## #1637 - Hardcoded localhost URL
Wallet endpoint hardcoded \http://localhost:5000/api/profile/wallet\. Now reads from \TRUXIFY_API_BASE_URL\ env var.

@KanishJebaMathewM **Why important**: #1635 is a compilation error - the app can't build. #1637 breaks wallet updates in production. Both are shipping blockers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the wallet address save flow to use the configured API base URL, with a localhost fallback.
  * Corrected an app configuration reference so the profile screen can load settings more reliably.

### DIFF
--- a/apps/driver/lib/screens/profile_screen.dart
+++ b/apps/driver/lib/screens/profile_screen.dart
@@ -12,7 +12,7 @@ import '../theme/app_theme.dart';
 import '../widgets/common_widgets.dart';
 import 'package:supabase_flutter/supabase_flutter.dart';
 import '../services/fcm_service.dart';
-import '../../core/supabase_config.dart';
+import '../core/supabase_config.dart';
 import 'package:truxify_shared/truxify_shared.dart' hide NotificationsScreen;
 import 'notifications_screen.dart';
 import '../utils/validators.dart';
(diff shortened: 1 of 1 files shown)


sanrishi:
@KanishJebaMathewM This PR fixes a compilation error (wrong import path blocks the build) and a hardcoded localhost URL that silently breaks wallet updates in production. Both are shipping blockers that need review.

PR #1656, opened 2026-07-02
fix(shared): DropdownButtonFormField value parameter

### LINKED ISSUE #1636: fix(shared): DropdownButtonFormField uses initialValue instead of value
CRITICAL: packages/truxify_shared/lib/src/screens/help_center_screen.dart:100 passes initialValue:_category to DropdownButtonFormField. Flutter's DropdownButtonFormField has no initialValue named parameter — the correct parameter is value. Compilation error.

sanrishi:
Closes #1636

\packages/truxify_shared/lib/src/screens/help_center_screen.dart:100\ passed \initialValue:\ to \DropdownButtonFormField\, but that parameter doesn't exist in Flutter's API. The correct parameter is \alue:\. This is a compilation error.

@KanishJebaMathewM **Why important**: Compilation error - the shared package can't be built. Blocks both customer and driver apps.


sanrishi:
@KanishJebaMathewM This is a compilation error - DropdownButtonFormField has no 'initialValue' parameter. The shared package won't build, blocking both customer and driver apps.

PR #1657, opened 2026-07-02
fix(customer): wallet API URL uses env var instead of hardcoded localhost

### LINKED ISSUE #1638: fix(customer): hardcoded localhost wallet URL in profile_screen
HIGH: apps/customer/lib/screens/profile_screen.dart:247 hardcodes http://localhost:5000/api/profile/wallet instead of using the configured API base URL. Wallet address updates silently fail in staging/production. Use String.fromEnvironment like the rest of the app.

sanrishi:
Closes #1638

The wallet address update endpoint in \pps/customer/lib/screens/profile_screen.dart:247\ hardcoded \http://localhost:5000/api/profile/wallet\. Changed to read from \TRUXIFY_API_BASE_URL\ env var like the rest of the app.

@KanishJebaMathewM **Why important**: Wallet updates silently fail in staging/production. Customers can never save their blockchain wallet address outside of local dev.

### DIFF
--- a/apps/customer/lib/screens/profile_screen.dart
+++ b/apps/customer/lib/screens/profile_screen.dart
@@ -244,7 +244,7 @@ class _ProfileScreenState extends State<ProfileScreen> {
                       final token = client.auth.currentSession?.accessToken;
                       final userId = client.auth.currentUser?.id ?? '';
                       final response = await http.put(
-                        Uri.parse('http://localhost:5000/api/profile/wallet'),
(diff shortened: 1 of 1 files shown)


sanrishi:
@KanishJebaMathewM Customer wallet address updates silently fail in staging/production because the endpoint URL is hardcoded to localhost:5000. Without this fix, customers can never save their blockchain wallet address outside of local dev.

PR #1661, opened 2026-07-02
fix(ml): fragile assertion for unknown user recommendations

### LINKED ISSUE #1641: fix(ml): fragile assertion in collaborative filter test for unknown user
MEDIUM: backend/ml/tests/test_collaborative_filter.py:55 asserts len(data['recommendations']) > 0 for an unknown user with empty booking_history and rated_drivers. The model may legitimately return 0 recommendations for a cold-start user. This test produces false failures in CI.

sanrishi:
Closes #1641

test_collaborative_filter.py:55 asserted len(data['recommendations']) > 0 for an unknown user with empty history. The model may legitimately return 0 recommendations for cold-start users, causing test flakiness.

@KanishJebaMathewM **Why important**: This test produces false failures in CI, reducing trust in the test suite and blocking deployments.

### DIFF
--- a/backend/ml/tests/test_collaborative_filter.py
+++ b/backend/ml/tests/test_collaborative_filter.py
@@ -52,7 +52,7 @@ def test_recommend_loads_unknown_user():
     assert response.status_code == 200
     data = response.json()
     assert "recommendations" in data
-    assert len(data["recommendations"]) > 0
+    assert isinstance(data["recommendations"], list)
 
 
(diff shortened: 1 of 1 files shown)


sanrishi:
@KanishJebaMathewM This test assertion produces false failures in CI because it assumes a cold-start user always gets recommendations. CI flakiness reduces trust in the test suite and can block deployments.

PR #1663, opened 2026-07-02
fix(driver): normalize route label separator before split

### LINKED ISSUE #1642: fix(driver): route label split depends on Unicode arrow character
MEDIUM: apps/driver/lib/screens/trip_detail_screen.dart:829 splits routeLabel on '→' (U+2192) only. If the backend sends '->' or '=>' as separator, the split silently fails and endLabel is empty, causing geocoding for the destination to fail.

sanrishi:
Closes #1642

trip_detail_screen.dart:829 split routeLabel on Unicode '→' only. If the backend sends '->' or '=>', the destination geocoding silently fails. Now normalizes all common separators before splitting.

@KanishJebaMathewM **Why important**: Route map display breaks silently if the backend ever changes its separator format, causing poor UX with incomplete routes shown to drivers.

### DIFF
--- a/apps/driver/lib/screens/trip_detail_screen.dart
+++ b/apps/driver/lib/screens/trip_detail_screen.dart
@@ -826,7 +826,8 @@ class _TripDetailScreenState extends State<TripDetailScreen> {
 
   Future<_RouteResult?> _loadRouteForTrip(String routeLabel) async {
     try {
-      final parts = routeLabel.split('→');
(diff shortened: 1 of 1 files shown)


sanrishi:
@KanishJebaMathewM Route map destination geocoding silently fails if the backend ever sends '->' instead of '→'. Drivers would see incomplete routes without realizing something went wrong.

Issue #1660, opened 2026-07-02
fix(driver): unsafe as Map cast in geocode_service


sanrishi:
MEDIUM: apps/driver/lib/services/geocode_service.dart:51 uses decoded.first as Map<String, dynamic>. If Nominatim returns a non-Map element (e.g., error response), this throws a runtime TypeError before reaching the catch block on line 62. Should use a safe cast.

PR #1664, opened 2026-07-02
fix(driver): safe type casts in marketplace and geocode services

### LINKED ISSUE #1645: fix(driver): unsafe as List cast in marketplace_repository API responses
HIGH: apps/driver/lib/services/marketplace_repository.dart:52,64 use jsonDecode(response.body) as List. If the API returns an error object (Map) instead of an array, this cast throws a runtime TypeError crashing the app. Should use a safe cast with type check.

sanrishi:
Closes #1645 and #1660

## #1645 - Unsafe as List cast
marketplace_repository.dart lines 51,63 used jsonDecode(response.body) as List which throws TypeError if the API returns an error object. Now uses is! List check first.

## #1660 - Unsafe as Map cast
geocode_service.dart line 51 used decoded.first as Map<String, dynamic> which throws if Nominatim returns unexpected data. Now uses is! check.

@KanishJebaMathewM **Why important**: Both are runtime crashes that can happen from non-2xx API responses. Crash = app termination for driver.

### DIFF
--- a/apps/driver/lib/services/geocode_service.dart
+++ b/apps/driver/lib/services/geocode_service.dart
@@ -48,7 +48,12 @@ class GeocodeService {
         return null;
       }
 
-      final item = decoded.first as Map<String, dynamic>;
+      final first = decoded.first;
+      if (first is! Map<String, dynamic>) {
+        _addToCache(key, null);
+        return null;
+      }
+      final item = first;
       final lat = double.tryParse('${item['lat']}');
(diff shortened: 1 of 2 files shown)


sanrishi:
@KanishJebaMathewM Two runtime crashes fixed: marketplace API error responses throw TypeError (app termination), and geocode_service crashes on unexpected Nominatim data. Both are crash-on-non-2xx-response bugs.

Issue #1651, opened 2026-07-02
fix(driver): hundreds of markers created for every route point on map


sanrishi:
MEDIUM: apps/driver/lib/screens/trips_screen.dart:1119-1136 creates a Marker for every single route point (hundreds of coordinates from OSRM). This causes severe performance degradation and visual clutter on the map. Should only create markers for start, end, and waypoints.

PR #1665, opened 2026-07-02
fix(driver): trips_screen disabled button + excess route markers

### LINKED ISSUE #1646: fix(driver): disabled Pull to refresh button with onPressed null
MEDIUM: apps/driver/lib/screens/trips_screen.dart:1232-1234 renders OutlinedAccentButton(label: 'Pull to refresh', onPressed: null). The disabled button appears clickable but does nothing. Replace with informational text since pull-to-refresh gesture is already available.

sanrishi:
Closes #1646 and #1651

## #1646 - Disabled button
OutlinedAccentButton with onPressed: null showed a disabled button that looked clickable but did nothing. Replaced with informational text.

## #1651 - Hundreds of route markers
MarkerLayer created a Marker for every single polyline point (hundreds). Now filters to first, last, and claimed points only.

@KanishJebaMathewM **Why important**: #1651 causes severe performance lag on trips with long routes. #1646 is confusing UX showing a disabled button that appears interactive.

### DIFF
--- a/apps/driver/lib/screens/trips_screen.dart
+++ b/apps/driver/lib/screens/trips_screen.dart
@@ -1067,7 +1067,11 @@ class _TripsScreenState extends State<TripsScreen> {
             ],
           ),
           MarkerLayer(
-            markers: routePoints.map((point) {
+            markers: routePoints.where((p) =>
+              p == routePoints.first ||
+              p == routePoints.last ||
+              p['is_claimed'] == true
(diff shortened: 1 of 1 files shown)


sanrishi:
@KanishJebaMathewM Two UX issues: a disabled 'Pull to refresh' button looks clickable but does nothing (confusing), and hundreds of markers on route maps cause severe performance lag for long trips.

PR #1666, opened 2026-07-02
fix(driver): null-safe field access in location_service

### LINKED ISSUE #1648: fix(driver): unsafe as String cast for activeOrder fields in location_service
HIGH: apps/driver/lib/services/location_service.dart:137-138 use activeOrder['id'] as String and activeOrder['order_display_id'] as String. If Supabase returns null or a missing field, this crashes with a runtime TypeError. Should use null-safe access.

sanrishi:
Closes #1648

location_service.dart:169-170 used activeOrder['id'] as String and activeOrder['order_display_id'] as String. These crash if Supabase returns null for these fields. Changed to null-safe ?.toString() access.

@KanishJebaMathewM **Why important**: Runtime crash when order field is null, causing the location service to terminate midpoint during trip tracking - driver loses GPS updates.

### DIFF
--- a/apps/driver/lib/services/location_service.dart
+++ b/apps/driver/lib/services/location_service.dart
@@ -166,8 +166,8 @@ class LocationService {
             .maybeSingle();
 
         if (activeOrder != null) {
-          _activeOrderId = activeOrder['id'] as String;
-          _activeOrderDisplayId = activeOrder['order_display_id'] as String;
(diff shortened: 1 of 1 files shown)


sanrishi:
@KanishJebaMathewM Runtime crash when Supabase returns null for order fields during location tracking. The location service terminates midpoint during trip tracking, causing drivers to lose GPS updates.

PR #1667, opened 2026-07-02
fix(ci): remove unused hasProgramSignal function

### LINKED ISSUE #1658: fix(ci): unused hasProgramSignal function in pr-labeler
LOW: .github/scripts/pr-labeler.js defines hasProgramSignal (line 16-19) but it is never called anywhere. The program detection is done inline (lines 72-88) instead. Dead code should be removed.

sanrishi:
Closes #1658

hasProgramSignal function was defined but never called anywhere in pr-labeler.js. Removed as dead code.

@KanishJebaMathewM **Why important**: Dead code clutters the script and can confuse maintainers into thinking program signal detection is active when it's actually done inline.

### DIFF
--- a/.github/scripts/pr-labeler.js
+++ b/.github/scripts/pr-labeler.js
@@ -13,11 +13,6 @@ function loadRules(rulesPath = DEFAULT_RULES_PATH) {
   return JSON.parse(fs.readFileSync(rulesPath, 'utf8'));
 }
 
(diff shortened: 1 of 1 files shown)


sanrishi:
@KanishJebaMathewM Dead code cleanup - hasProgramSignal function was defined but never called. Maintainers might think program signal detection is active when it's actually done inline.

PR #1746, opened 2026-07-03
fix: replace unnecessary dynamic import with top-level supabase client in verifyBookingOwnership

### LINKED ISSUE #1740: locationServer.js performs unnecessary dynamic import() on every booking ownership check
In `backend/api/src/sockets/locationServer.js`, the `verifyBookingOwnership` function (line 276) does:

```js
async function verifyBookingOwnership(customerId, bookingId) {
  try {
    const { supabase } = await import("../config/db.js");
    // ...
  }
}
```

This performs a **dynamic import** on every call, even though `supabase` is already imported at the top of the file (line 5). Issues:
1. The module is re-imported on every ownership check (wasteful I/O)
2. It shadows the top-level `supabase` which could be null
3. Dynamic `import()` returns a new promise every time

**Fix:** Use the already-imported top-level `supabase` instead of dynamic import.

Xenon010101:
Closes #1740

fix: replace unnecessary dynamic import with top-level supabase client in verifyBookingOwnership.

### DIFF
--- a/backend/api/src/sockets/locationServer.js
+++ b/backend/api/src/sockets/locationServer.js
@@ -272,8 +272,6 @@ async function verifyCustomerToken(socket, next) {
 async function verifyBookingOwnership(customerId, bookingId) {
   try {
     // Use Supabase client from existing db module
-    // Import Supabase client from existing db module
-    const { supabase } = await import("../config/db.js");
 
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSOC

PR #1747, opened 2026-07-03
fix: add process.exit(1) to unhandledRejection handler to avoid running in undefined state

### LINKED ISSUE #1741: unhandledRejection handler suppresses process crash without process.exit(1)
In `backend/api/src/index.js`, lines 237-240:

```js
process.on('unhandledRejection', async (reason) => {
  logger.error({ reason }, 'Unhandled promise rejection');
  await flushSentry(2000);
  // Missing: process.exit(1)
});
```

Since Node.js v15, the default behavior for unhandled rejections is to terminate the process. By adding this listener, the default crash is **overridden** - the process continues running with potentially corrupted state (half-closed DB connections, inconsistent in-memory data).

Contrast with `uncaughtException` handler (lines 231-235) which correctly calls `process.exit(1)`.

**Fix:** Add `process.exit(1)` after `flushSentry`.

Xenon010101:
Closes #1741

fix: add process.exit(1) to unhandledRejection handler to avoid running in undefined state.

### DIFF
--- a/backend/api/src/index.js
+++ b/backend/api/src/index.js
@@ -238,6 +238,7 @@ process.on('uncaughtException', async (err) => {
 process.on('unhandledRejection', async (reason) => {
   logger.error({ reason }, 'Unhandled promise rejection');
   await flushSentry(2000);
+  process.exit(1);
 });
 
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSOC

PR #1748, opened 2026-07-03
fix: make OTP generation respect configured OTP_LENGTH constant instead of hardcoded 4 digits

### LINKED ISSUE #1742: OTP length logic ignores configured OTP_LENGTH constant making it effectively hardcoded
In `backend/api/src/services/otpService.js`:

```js
const OTP_LENGTH = 4;                                       // line 5
// ...
const otp = String(Math.floor(1000 + Math.random() * 9000)).slice(0, OTP_LENGTH);  // line 12
```

`Math.floor(1000 + Math.random() * 9000)` always produces a **4-digit number** (1000-9999). The `.slice(0, OTP_LENGTH)` is a no-op. If someone changes `OTP_LENGTH` to, say, 6, the `.slice(0, 6)` still produces only 4 digits because the generated number is only 4 digits.

**Fix:** Generate the OTP using the configured length dynamically:
```js
const min = Math.pow(10, OTP_LENGTH - 1);
const max = Math.pow(10, OTP_LENGTH) - 1;
const otp = String(Math.floor(min + Math.random() * (max - min + 1)));
```

Xenon010101:
Closes #1742

fix: make OTP generation respect configured OTP_LENGTH constant instead of hardcoded 4 digits.

## Summary by CodeRabbit

* **Bug Fixes**
  * OTP codes are now generated at a consistent length, based on the configured OTP settings.
  * Improved OTP number generation to better match the expected format before delivery.

### DIFF
--- a/backend/api/src/services/otpService.js
+++ b/backend/api/src/services/otpService.js
@@ -9,7 +9,9 @@ export async function generateAndStoreOtp(phone) {
     logger.warn('[otp] Redis not available, cannot generate OTP.');
     return null;
   }
-  const otp = String(Math.floor(1000 + Math.random() * 9000)).slice(0, OTP_LENGTH);
+  const min = Math.pow(10, OTP_LENGTH - 1);
+  const max = Math.pow(10, OTP_LENGTH) - 1;
+  const otp = String(Math.floor(min + Math.random() * (max - min + 1)));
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSOC

Issue #1763, opened 2026-07-03
fix(shared): unsafe as List cast in faq_repository


sanrishi:
HIGH - Runtime crash: packages/truxify_shared/lib/src/repositories/faq_repository.dart:16 uses 'response as List'. If Supabase returns null or a non-List, this throws TypeError crashing the FAQ screen. Should use safe cast.

PR #1782, opened 2026-07-03
fix(shared): safe casts in notification and faq repositories

### LINKED ISSUE #1762: fix(shared): unsafe as List cast in notification_repository
HIGH - Runtime crash: packages/truxify_shared/lib/src/repositories/notification_repository.dart:16 uses 'response as List'. If Supabase returns null or a non-List (e.g., RLS error), this throws TypeError crashing the notifications screen. Should use safe cast.

sanrishi:
Closes #1762 #1763

Both repositories used 'response as List' which throws TypeError if Supabase returns null or a non-List. Changed to 'response is List' safe check with empty fallback.

@KanishJebaMathewM **Why important**: Runtime crashes on the notifications and FAQ screens whenever Supabase returns an unexpected response. Both screens are unusable until this is fixed.

### DIFF
--- a/packages/truxify_shared/lib/src/repositories/faq_repository.dart
+++ b/packages/truxify_shared/lib/src/repositories/faq_repository.dart
@@ -13,7 +13,9 @@ class FaqRepository {
         .select()
         .eq('is_active', true)
         .order('sort_order');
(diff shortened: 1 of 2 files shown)


sanrishi:
@KanishJebaMathewM Both notification and FAQ screens crash with TypeError when Supabase returns unexpected data. Safe casts prevent these runtime crashes.

PR #1790, opened 2026-07-03
fix(customer): validate weight before submitting order

### LINKED ISSUE #1771: fix(customer): weight silently converts to 0 on non-numeric input
MEDIUM - Silent data loss: apps/customer/lib/screens/booking_confirmation_screen.dart:121 uses double.tryParse(widget.draft.weightTonnes) ?? 0. If weight string is non-numeric (e.g., 'N/A'), it silently becomes 0, submitting an order with 0 weight which the API may reject.

sanrishi:
Closes #1771

booking_confirmation_screen.dart silently defaulted to weight=0 on non-numeric input, submitting invalid orders. Now validates weight > 0 with user-facing error message before submission.

@KanishJebaMathewM **Why important**: Non-numeric weight values silently become 0, submitting orders with invalid data that the API rejects with confusing errors.

### DIFF
--- a/apps/customer/lib/screens/booking_confirmation_screen.dart
+++ b/apps/customer/lib/screens/booking_confirmation_screen.dart
@@ -106,6 +106,15 @@ class _BookingConfirmationScreenState extends State<BookingConfirmationScreen>
       return;
     }
 
+    final weight = double.tryParse(widget.draft.weightTonnes);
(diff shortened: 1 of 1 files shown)


sanrishi:
@KanishJebaMathewM Non-numeric weight silently becomes 0, submitting invalid orders that the API rejects with confusing errors.

PR #1791, opened 2026-07-03
fix(customer): safe casts in location_service

### LINKED ISSUE #1772: fix(customer): unsafe as List and as Map casts in location_service
MEDIUM - Runtime crash: apps/customer/lib/services/location_service.dart:45 uses jsonDecode(response.body) as List<dynamic> and line 48 uses item as Map<String, dynamic>. If Nominatim returns unexpected data, these throw TypeError. Also line 87: as Map<String, dynamic> for reverse lookup.

sanrishi:
Closes #1772

As List/Map casts replaced with is! checks. Prevents TypeError crashes when Nominatim returns unexpected responses.

@KanishJebaMathewM **Why important**: Runtime crashes on customer search and reverse geolocation when Nominatim returns unexpected data, breaking the booking flow.

### DIFF
--- a/apps/customer/lib/services/location_service.dart
+++ b/apps/customer/lib/services/location_service.dart
@@ -42,10 +42,12 @@ class LocationService {
       throw Exception('Search failed: ${response.statusCode} (${uri.path})');
     }
 
-    final decoded = jsonDecode(response.body) as List<dynamic>;
+    final decoded = jsonDecode(response.body);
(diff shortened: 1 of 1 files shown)


sanrishi:
@KanishJebaMathewM Customer location_service crashes with TypeError when Nominatim returns unexpected responses, breaking the entire booking flow.

PR #1805, opened 2026-07-03
fix: claim_release_reconciliation RPC missing (#1800)

### LINKED ISSUE #1800: claim_release_reconciliation RPC Function Does Not Exist in Any Migration — Escrow Release Reconciliation Locking Is Broken
### Summary

The `escrowReleaseReconciliation.js` service calls `supabase.rpc('claim_release_reconciliation', ...)` at line 63, but this PostgreSQL function **was never created**. The sibling migration `20260627100000_add_reconciled_by_and_claim_rpc.sql` only defines `claim_refund_reconciliation` (for refunds) — the release counterpart is missing. The error fallback allows the service to limp along without claim locking, enabling double-processing of failed escrow releases.

### Impact

- The claim-based distributed lock for release reconciliation is completely non-functional.
- Multiple server instances **can** process the same `release_failed` order simultaneously.
- Double on-chain `releaseFunds` calls waste gas (each costs ~50k–100k MATIC gas).
- Race condition in DB updates: two instances can both update `escrow_status` to `released`, or one overwrites the other's data.
- The error fallback path (lines 73-83) does not acquire any lock, making the reconciliation not safe for horizontal scaling.

### Affected Files

| File | Lines | Issue |
|------|-------|-------|
| `backend/api/src/services/escrowReleaseReconciliation.js` | 62-66 | Calls `claim_release_reconciliation` RPC that does not exist |
| `backend/api/src/services/escrowReleaseReconciliation.js` | 73-83 | Error fallback re-fetches order status but does NOT claim/lock it — TOCTOU window |
| `backend/api/src/services/escrowReleaseReconciliation.js` | 88-106 | Without the claim lock, multiple instances can race on the blockchain call and DB update |
| `supabase/migrations/20260627100000_add_reconciled_by_and_claim_rpc.sql` | 6-23 | Only defines `claim_refund_reconciliation`. A `claim_release_reconciliation` function is entirely absent. |
| `supabase/migrations/` | | A new migration is needed |

### Root Cause

`claim_release_reconciliation` was referenced in code but never migrated to the database. Comparing with the refund path: `claim_refund_reconciliation` exists in a migration (lines 6-23 of the SQL file) and is called from `escrowRefundReconciliation.js:57`. The release version was simply forgotten.

### Proposed Fix (Multi-File)

1. **New Supabase migration**: Create `claim_release_reconciliation(p_order_id UUID, p_instance_id TEXT)` following the same pattern as `claim_refund_reconciliation` but for `escrow_status = 'release_failed'`.
2. **`escrowReleaseReconciliation.js`**: After the RPC call, verify the claim was acquired and log/skip if not (similar to the refund service error handling).
3. **`escrowReleaseReconciliation.js`**: Add a `reconciled_by` null guard on the final DB update (lines 94-106) to prevent the race.
4. **Update migration comment**: Add a reference in the existing migration file pointing to the new migration.

ionfwsrijan:
### Summary
Create the `claim_release_reconciliation` RPC (only `claim_refund_reconciliation` existed) and guard the final DB update against races.

### Changes
1. **`supabase/migrations/20260703000000_add_claim_release_reconciliation.sql`** — New migration creating `claim_release_reconciliation` RPC that atomically claims a `release_failed` order by setting `reconciled_by` and `reconciled_at`.
2. **`backend/api/src/services/escrowReleaseReconciliation.js`** — Added `.is('reconciled_by', null)` to the final `escrow_status` update to prevent two reconciliation instances from updating the same order.

Closes #1800

### DIFF
--- a/backend/api/src/services/escrowReleaseReconciliation.js
+++ b/backend/api/src/services/escrowReleaseReconciliation.js
@@ -2,7 +2,6 @@ import { supabase, redisClient } from '../config/db.js';
 import { escrowRelease } from './escrow.js';
 import logger from '../middleware/logger.js';
 import os from 'os';
-import logger from '../middleware/logger.js';
 const DEFAULT_INTERVAL_MS = 60_000;
 const LOCK_KEY = 'escrow:release:reconciliation:lock';
 const LOCK_TTL_SECONDS = 120;
@@ -67,7 +66,7 @@ export async function reconcilePendingEscrowReleases() {
             p_instance_id: instanceId,
           });
 
-        if ((!claimed || claimed.length === 0) && !claimError) {
+        if ((!claimed || (Array.isArray(claimed) && claimed.length === 0)) && !claimError) {
           logger.info(`[escrow-release-reconciliation] Order ${order.order_display_id} already claimed by another instance, skipping.`);
           continue;
         }
--- a/backend/api/src/services/escrowReleaseReconciliation.js
+++ b/backend/api/src/services/escrowReleaseReconciliation.js
@@ -105,7 +105,8 @@ export async function reconcilePendingEscrowReleases() {
             updated_at: releasedAt,
           })
           .eq('id', order.id)
-          .eq('escrow_status', 'release_failed');
+          .eq('escrow_status', 'release_failed')
+          .is('reconciled_by', null);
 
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

PR #1807, opened 2026-07-03
fix: concurrent verify-delivery race condition (#1802)

### LINKED ISSUE #1802: Concurrent verify-delivery Requests Can Trigger Redundant On-Chain Escrow Releases and Bypass Optimistic DB Guard
### Summary

The delivery verification endpoint (`POST /:id/verify-delivery`) performs a multi-step sequence: OTP check → order status guard (optimistic update) → blockchain release → `complete_trip_tx` RPC → OTP consumption. Steps 2, 3, and 4 are not covered by a single database transaction or distributed lock. Two concurrent requests can both pass the optimistic status guard at Step 2 and trigger:

1. **Redundant blockchain releases**: Both requests call `releaseFunds` on-chain. The on-chain contract check (`status === 2`) provides idempotency, but the second call still consumes gas (~$0.01–0.05 per redundant tx on Polygon).
2. **Post-RPC verification gap**: The `complete_trip_tx` RPC (Step 4, line 1068) uses `SELECT ... FOR UPDATE` which serializes correctly. However, the blockchain release at Step 3 (line 1044) happens *before* the RPC acquires its row lock. If the RPC fails after the blockchain release succeeds, the order remains in `arriving` state with an unconsumed OTP — requiring the reconciliation service to detect and recover.
3. **Cancel race**: The cancellation endpoint (line 1290) has a similar guard (`not('status', 'in', '("delivered","payment_released")')`) but the check for OTP verification (lines 1307-1317) is separate from the update. A concurrent verify-delivery and cancel can race: cancel sees no verified OTP, proceeds, while verify-delivery then succeeds.

### Impact

- Wasted gas on redundant blockchain transactions (cumulatively significant at scale).
- Orders can transiently enter an inconsistent state where blockchain says "released" but DB says "arriving".
- The reconciliation service must detect and fix these inconsistencies, but the release reconciliation is itself broken (see Issue 3).
- Concurrent cancel + verify can lead to a cancelled order having its escrow released to the driver.

### Affected Files

| File | Lines | Issue |
|------|-------|-------|
| `backend/api/src/routes/orderRoutes.js` | 1028-1076 | Optimistic status guard (1028-1035) has no row lock; blockchain release (1044-1066) happens before the RPC (1068-1076) |
| `backend/api/src/routes/orderRoutes.js` | 1307-1317 | Cancel endpoint's OTP-verified check is a separate SELECT, not atomically combined with the UPDATE |
| `backend/api/src/routes/orderRoutes.js` | 1290-1488 | Cancel and verify-delivery are separate transactions — no inter-transaction locking |
| `backend/api/src/services/notificationService.js` | 150-166 | `verifyDeliveryOtp()` is called AFTER the RPC — if the response is lost, the OTP is consumed in the RPC but the client sees a failure |
| `supabase/migrations/20260624223000_make_delivery_otp_completion_atomic.sql` | 5-121 | `complete_trip_tx` RPC is correct (uses `FOR UPDATE`) but is not the first operation in the HTTP handler — the blockchain call precedes it |

### Root Cause

The verify-delivery HTTP handler does not hold a database row lock across the sequence of checks → blockchain → DB update. The blockchain release is an irreversible side effect that cannot be rolled back if subsequent steps fail. The cancellation and verify-delivery endpoints have independent optimistic guards that can race.

### Proposed Fix (Multi-File)

1. **`orderRoutes.js`**: Move the blockchain `releaseFunds` call to AFTER the `complete_trip_tx` RPC succeeds, not before. If the RPC succeeds and the blockchain call fails, the order is in `payment_released` state without actual on-chain release, which is safer than the inverse.
2. **`orderRoutes.js`**: Acquire a row-level lock (`SELECT ... FOR UPDATE`) at the start of the verify-delivery and cancel handlers to serialize concurrent access to the same order.
3. **`orderRoutes.js`**: Use Supabase's RPC to atomically check OTP-verified status AND update in the cancel endpoint (a single `cancel_order_tx` RPC similar to `complete_trip_tx`).
4. **New Supabase migration**: Create a `cancel_order_tx(p_order_id UUID)` RPC that does all checks + updates atomically, matching the pattern of `complete_trip_tx`.
5. **`notificationService.js`**: Remove the redundant `verifyDeliveryOtp()` call at line 1100 — the `complete_trip_tx` RPC already marks the OTP as verified. This eliminates the out-of-order OTP consumption.

ionfwsrijan:
### Summary
Prevent two concurrent verify-delivery requests from both releasing escrow by adding row-level locking at the handler start. Replace the cancel handler's multiple SQL statements with an atomic RPC.

### Changes
1. **`supabase/migrations/20260703000001_add_lock_order_and_cancel_order_rpcs.sql`** — New migration creating two RPCs:
   - `lock_order_for_update(p_order_id)` — SELECT ... FOR UPDATE on the order row to serialize concurrent handlers.
   - `cancel_order_tx(p_order_id, p_cancellation_reason, p_customer_id)` — Atomically validate ownership, check status, and transition to cancelled/refund_pending with row locking.
2. **`backend/api/src/routes/orderRoutes.js`**
   - verify-delivery handler: call `lock_order_for_update` before any state mutations.
   - cancel handler: replace the multi-step fetch/update chain with a single `cancel_order_tx` RPC call, eliminating TOCTOU between steps.

Closes #1802


ionfwsrijan:

PR #1980, opened 2026-07-04
Fix: Handle state-border array bounds in Price Forecaster (#1953)

### LINKED ISSUE #1953: [Bug] Backend: Dynamic Price Forecaster 500 error on 3+ state borders
The Dynamic Price Forecaster API returns an HTTP 500 Internal Server Error when calculating price ranges for long-haul routes that cross three or more state borders. This seems to be an array index out-of-bounds issue in the toll calculation logic.

desireddymohithreddy0925:
Fixed an HTTP 500 error that crashed the price forecaster on long-haul routes crossing 3+ states. The state-border loop is now dynamically sized based on the route polygon. Closes #1953.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+Fix issue 1953


desireddymohithreddy0925:
> 

GSSOC

PR #2008, opened 2026-07-04
fix: Enforce WebSocket driver booking and order ownership validation

### LINKED ISSUE #1971: [Critical][Backend] WebSocket location namespace has no booking ownership validation - any driver can spoof GPS for any booking
## Summary
The /driver WebSocket namespace in ackend/api/src/sockets/tracker.js allows any authenticated driver to emit GPS location updates for **any** ookingId without verifying they are the assigned driver for that booking. This means a malicious driver can hijack another driver's live tracking stream, inject false GPS coordinates, and manipulate delivery verification.

## Impact
- **Critical**: A driver can send fake GPS pings for a booking they are not assigned to, triggering false geofence events and potentially auto-releasing escrow funds.
- Customer-facing live tracking displays can be poisoned with arbitrary coordinates.
- Delivery OTP verification relies on geofence proximity - false GPS can bypass this check.
- Driver reputation scores based on route accuracy can be manipulated.

## Affected Files
- ackend/api/src/sockets/tracker.js - missing booking ownership check in the driver GPS event handler
- ackend/api/src/middleware/auth.js - no booking-level authorization layer

## Root Cause
The socket handler calls erifyBookingOwnership(bookingId, driverId) as an async function but does **not** await it properly in all code paths, or the function is only called conditionally. The check is bypassed when BYPASS_AUTH=true.

## Proposed Fix
1. In 	racker.js, enforce wait verifyBookingOwnership(bookingId, socket.userId) before processing any GPS ping and disconnect immediately if it fails.
2. Cache booking ownership in Redis (keyed by ookingId:driverId) with a TTL matching the booking duration to avoid per-message DB lookups.
3. Add integration test: connect two different driver sockets, assert that driver B cannot emit location for driver A's booking.
4. Remove the BYPASS_AUTH path entirely from WebSocket handlers.

## Steps to Reproduce
1. Start two authenticated driver sessions (driver A and driver B).
2. Create a booking assigned to driver A.
3. From driver B's socket, emit { bookingId: <driver A's booking>, lat: 0, lng: 0 }.
4. Observe that the location is accepted and broadcast to the customer tracking stream.

saidai-bhuvanesh:
## Summary

Closes #1971

This PR addresses a critical WebSocket security vulnerability in the /ws/tracking and /driver namespaces where any authenticated driver could spoof location tracking pings and emit false GPS coordinates for any active booking or order.

## Problem

Prior to this change, the driver connection handshake extracted the driverId from authentication but trusted the client-provided ookingId verbatim without verifying that the driver was actually assigned to that specific booking. Similarly, the location tracking pings inside 	racker.js accepted any order_id or order_display_id without verifying that the authenticated driver matched the driver_id stored on the order in Supabase. This allowed malicious or misconfigured clients to hijack other drivers' live location streams and poison the geofence validation mechanism.

## Changes

- **ackend/api/src/sockets/locationServer.js**:
  - Implemented a erifyDriverBookingOwnership(driverId, bookingId) helper. It queries the ookings table in Supabase and falls back to the orders table to ensure database schema compatibility.
  - Injected this check into the Socket.IO erifyDriverToken middleware so that connection fails immediately with a Forbidden error if the driver is not assigned to the requested booking.

- **ackend/api/src/sockets/tracker.js**:
  - Updated the Supabase order resolution query in handleLocationPing to select the driver_id field.
  - Added an ownership check: if (order.driver_id !== driver_id). If there is a mismatch, it logs the spoofing attempt (UNAUTHORIZED_LOCATION_PING) and sends a Forbidden socket error to the client, ignoring the ping.

## Verification

- Verified that all unit tests in ackend/api/test/unit/tracker.test.js pass successfully (
pm run test:unit test/unit/tracker.test.js).
- Manual verification shows unauthorized pings are ignored and logged as warnings/errors.


saidai-bhuvanesh:
Gssoc

PR #2062, opened 2026-07-05
fix(profile): remove hardcoded test data fallback in profileService.js

### LINKED ISSUE #2044: fix(profile): remove hardcoded test data fallback in profileService.js
### Description
`getProfile()`, `getCustomerStats()`, and `getDriverDetails()` return hardcoded test data when Supabase is unavailable. In production, this silently returns fake user data instead of failing. Fix by throwing an explicit error and update the test accordingly.

### Files
- `backend/api/src/services/profileService.js`
- `backend/api/test/unit/profileService.test.js`

### Type
bug

sahare77:
Resolves #2044

### DIFF
--- a/backend/api/src/services/profileService.js
+++ b/backend/api/src/services/profileService.js
@@ -42,5 +42,14 @@ export async function getDriverDetails(userId) {
     .maybeSingle();
 
   if (error) throw error;
+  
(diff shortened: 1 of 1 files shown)


sahare77:
GSSOC

PR #2063, opened 2026-07-05
fix(tracker): reuse WebSocket Supabase Realtime channels by orderUUID to prevent leaks

### LINKED ISSUE #2045: fix(tracker): reuse WebSocket Supabase Realtime channels by orderUUID to prevent leaks
### Description
Each location ping creates a new Supabase Realtime channel but only removes it on broadcast completion. Channels leak if the WebSocket closes mid-flight. Fix by caching channels per `orderUUID` in a Map, reusing them, and cleaning up on disconnect.

### Files

### Type
bug

sahare77:
Resolves #2045


sahare77:
GSSOC

PR #2070, opened 2026-07-05
feat(auth): implement refresh token rotation to mitigate token theft

### LINKED ISSUE #2052: feat(auth): implement refresh token rotation to mitigate token theft
### Description
The current authentication flow in `authRoutes.js` does not invalidate old refresh tokens when a new one is issued (Refresh Token Rotation). If a refresh token is compromised, an attacker can maintain persistent access. Implement token rotation and a mechanism to revoke the token family if reuse is detected.

### Files
- `backend/api/src/routes/authRoutes.js`

### Type
security

sahare77:
Resolves #2052

### DIFF
--- a/backend/api/src/routes/authRoutes.js
+++ b/backend/api/src/routes/authRoutes.js
@@ -79,3 +79,5 @@ router.get('/session', authenticate, (req, res) => {
 });
 
 export default router;
+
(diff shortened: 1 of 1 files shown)


sahare77:
GSSOC

PR #2071, opened 2026-07-05
fix(trucks): prevent race conditions in truck allocation logic

### LINKED ISSUE #2053: fix(trucks): prevent race conditions in truck allocation logic
### Description
When a truck is assigned to a trip or driver in `truckRoutes.js`, there is a risk of a race condition if two requests are made simultaneously. This can result in a single truck being allocated to two different tasks. Implement pessimistic locking or atomic database updates to ensure exclusive allocation.

### Files
- `backend/api/src/routes/truckRoutes.js`

### Type
bug

sahare77:
Resolves #2053


sahare77:
GSSOC

PR #2109, opened 2026-07-05
fix: use BigInt comparison for booking ID to handle type mismatch 

### LINKED ISSUE #2080: Booking ID Type Mismatch (`bigint !== string`) in `recordDepositTx` — Deposit Verification Always Fails
## Description

The `recordDepositTx` function compares a `bigint` (decoded from the on-chain transaction via the ABI) with a `string` (the booking ID generated by `getEscrowBookingId`). In JavaScript, `BigInt('0xabc123') !== '0xabc123'` is always `true` because the types differ. Every call to `recordDepositTx` returns `{ error: 'Transaction booking ID does not match' }`.

**The flow:**
1. Customer accepts a bid → `buildDepositTx` populates a transaction for `createBooking(bookingId, driver)`
2. Customer signs and submits the transaction, gets a `txHash`
3. Customer calls `POST /:id/confirm-deposit` with the `txHash`
4. `recordDepositTx` decodes the on-chain transaction, extracts `txBookingId` as `bigint`, compares to `bookingId` (hex string)
5. Comparison always fails. Returned to customer as `{ error: 'Transaction booking ID does not match' }`
6. Order remains stuck in `escrow_status = 'funding'` forever

## Code

```javascript
// escrow.js line 64-66 — returns a hex string
export function getEscrowBookingId(orderDisplayId) {
  return ethers.solidityPackedKeccak256(['string'], [`escrow:${orderDisplayId}`]);
}

// escrow.js line 150 — decoded via ABI, returns bigint
decoded = escrowContract.interface.parseTransaction({ data: tx.data, value: tx.value });
// decoded.args[0] is bigint (uint256 from ABI)

// escrow.js line 160 — ALWAYS mismatches due to type difference
const [txBookingId, txDriver] = decoded.args;
if (txBookingId !== bookingId) {   // bigint !== string → ALWAYS TRUE
  return { error: 'Transaction booking ID does not match' };
}
```

## Impact

- Even if Issue #1 (ABI mismatch) is fixed, this issue independently prevents any escrow from being confirmed as `funded`
- The `verify-delivery` endpoint checks `order.escrow_status === 'funded'` before releasing payment — the entire payment pipeline is deadlocked
- Orders stay in `'funding'` state permanently — the driver cannot start the trip and the customer cannot cancel without risking funds
- No automated retry or manual override mechanism exists for stuck deposits

## Fix

Compare the `bigint` with the `bigint` representation of the booking ID:

```javascript
// Line 160 — change from:
if (txBookingId !== bookingId) {
// To:
if (txBookingId !== BigInt(bookingId)) {
```

A single character change (`BigInt()`) fixes the entire deposit verification pipeline.

ionfwsrijan:
## Description

Fixes #2080

`recordDepositTx` compared a `bigint` (decoded from on-chain ABI) with a `string`
(booking ID from `getEscrowBookingId`). Strict inequality (`!==`) always returned `true`,
permanently blocking deposit verification.

### Changes (`backend/api/src/services/escrow.js`)

Wrapped both sides of the comparison in `BigInt()` to handle type mismatch regardless
of whether the ABI returns `bigint` or `string`:
```js
if (BigInt(txBookingId) !== BigInt(bookingId))
```

### Impact

- Deposit verification now works correctly across type representations
- Orders can progress past escrow_status = 'funding' to 'funded'

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved booking ID matching for deposit transactions by treating equivalent numeric values as the same, even when represented differently.
  * Reduced false mismatch errors during deposit recording, helping transactions validate more reliably.

### DIFF
--- a/backend/api/src/services/escrow.js
+++ b/backend/api/src/services/escrow.js
@@ -151,7 +151,7 @@ export async function recordDepositTx(bookingId, txHash, expectedSenderAddress =
   }
 
   const [txBookingId, txCustomer] = decoded.args;
-  if (txBookingId !== bookingId) {
+  if (BigInt(txBookingId) !== BigInt(bookingId)) {
     return { error: 'Transaction booking ID does not match' };
   }
 


ionfwsrijan:

PR #2110, opened 2026-07-05
fix: handle orphaned refund_failed and tx-hash-less refund_pending states in reconciliation

### LINKED ISSUE #2081: Orphaned Refund State — `refund_failed` and Tx-Hash-Less `refund_pending` Orders Never Recovered
## Description

The refund reconciliation service only queries orders with `escrow_status = 'refund_pending' AND refund_tx_hash IS NOT NULL`. Two orphan scenarios fall through this filter:

### Scenario A: Crash between DB update and blockchain submission (no tx_hash, status = `refund_pending`)

The cancel endpoint first updates the DB (orderRoutes.js line 1354–1368):
```sql
UPDATE orders SET status = 'cancelled', escrow_status = 'refund_pending' WHERE id = $1;
```

Then acquires a lock and calls the blockchain (line 1398). If the server crashes between these two operations, the DB says `refund_pending` but `refund_tx_hash` is `NULL`. The reconciliation service filters on `NOT refund_tx_hash IS NULL`, so this order is **never picked up**.

### Scenario B: Blockchain submission fails (no tx_hash, status = `refund_failed`)

If `submitEscrowRefund` throws, the catch block (line 1461–1471) sets:
```javascript
const nextEscrowStatus = refundTxHash ? 'refund_pending' : 'refund_failed';
```

Since `refundTxHash` is null, `escrow_status` becomes `'refund_failed'`. The reconciliation service **never queries for `refund_failed`**. The customer's refund is stuck forever.

### Contrast with release reconciliation

The release reconciliation service (`escrowReleaseReconciliation.js`) explicitly handles `release_failed` orders (line 47), has a retry budget (`MAX_RETRIES = 10`, line 8), exponential backoff, and a claim-based locking mechanism. The refund reconciliation service has none of these.

## Code

```javascript
// escrowRefundReconciliation.js lines 29-34
const { data: pendingRefunds } = await supabase
  .from('orders')
  .select('order_display_id, refund_tx_hash')
  .eq('escrow_status', 'refund_pending')
  .not('refund_tx_hash', 'is', null);  // ← MISSING: refund_failed and no-tx_hash cases
```

## Impact

- A customer who cancels a funded order could lose their entire escrow deposit — no automated process recovers these funds
- Server crash during cancellation permanently locks customer funds on-chain
- Blockchain gas failure (e.g., relayer wallet insufficient MATIC) permanently locks customer funds
- Manual database intervention required to recover — problematic for a production marketplace
- The pattern is inconsistent: release failures have a full reconciliation system while refund failures have none

## Fix

1. Expand the reconciliation query to include all unrecoverable states:
```javascript
.in('escrow_status', ['refund_pending', 'refund_failed'])
```

2. Handle the no-tx-hash case by calling `submitEscrowRefund` (now `cancelBooking`) to submit the actual transaction

3. Add a retry budget with exponential backoff mirroring the release reconciliation pattern:
```javascript
const MAX_RETRIES = 10;
if (row.retry_count >= MAX_RETRIES) {
  // Escalate to manual intervention — notify ops
  continue;
}
```

ionfwsrijan:
## Description

Fixes #2081

The refund reconciliation service only queried `refund_pending` with non-null
`refund_tx_hash`. Two orphan scenarios were missed: (A) crash between DB update and
blockchain submission leaves `refund_pending` with null tx_hash, (B) blockchain failure
leaves `refund_failed` status.

### Changes (`backend/api/src/services/escrowRefundReconciliation.js`)
1. Expanded query to include both `refund_pending` and `refund_failed` statuses
2. For orders without a `refund_tx_hash`, calls `submitEscrowRefund` to submit the
   transaction on-chain before confirming
3. Added `MAX_RETRIES = 10` budget with retry count tracking (`escrow_refund_retry_count`)
4. Orders exceeding max retries are escalated (logged) for manual intervention

### DIFF
--- a/backend/api/src/services/escrowRefundReconciliation.js
+++ b/backend/api/src/services/escrowRefundReconciliation.js
@@ -1,12 +1,13 @@
 import { supabase, redisClient } from '../config/db.js';
 import logger from '../middleware/logger.js';
-import { confirmEscrowRefund } from './escrow.js';
+import { confirmEscrowRefund, submitEscrowRefund } from './escrow.js';
 import os from 'os';
 
 const DEFAULT_INTERVAL_MS = 60_000;
 const LOCK_KEY = 'escrow:reconciliation:lock';
 const LOCK_TTL_SECONDS = 120;
 const LEASE_EXTENSION_INTERVAL_MS = (LOCK_TTL_SECONDS * 1000) / 2;
+const MAX_RETRIES = 10;
 let reconciliationTimer = null;
 let reconciliationRunning = false;
 
@@ -43,9 +44,8 @@ export async function reconcilePendingEscrowRefunds() {
     const instanceId = process.env.HOSTNAME || os.hostname();
     const { data: pendingOrders, error } = await supabase
       .from('orders')
-      .select('id, order_display_id, refund_tx_hash')
-      .eq('escrow_status', 'refund_pending')
-      .not('refund_tx_hash', 'is', null)
+      .select('id, order_display_id, refund_tx_hash, escrow_status, escrow_refund_retry_count')
+      .in('escrow_status', ['refund_pending', 'refund_failed'])
       .limit(50);
 
     if (error) {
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #2156, opened 2026-07-05
fix(ci): use actions/checkout@v4 instead of non-existent @v7

### LINKED ISSUE #2139: fix(ci): actions/checkout@v7 does not exist in ci.yml
HIGH: .github/workflows/ci.yml lines 110,133 use @v7 which doesn't exist (latest is v4). Jobs fail at runtime.

sanrishi:
Closes #2139

Both ml and client-credentials-guard jobs use checkout@v7 which doesn't exist (latest is v4). Jobs silently fail at runtime.

## Summary by CodeRabbit

* **Chores**
  * Updated the CI workflow’s checkout action version for two jobs, helping keep build and validation steps current.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,7 +107,7 @@ jobs:
     timeout-minutes: 15
 
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@v4
 
       - name: Setup Python 3.11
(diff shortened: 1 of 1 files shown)


sanrishi:
@KanishJebaMathewM CI pipeline silently fails on ml and credentials-guard jobs - checkout@v7 doesn't exist, jobs just hang.

Issue #2153, opened 2026-07-05
fix(api): rateLimiter.js unused RedisStore import


sanrishi:
LOW: backend/api/src/middleware/rateLimiter.js line 2 imports RedisStore but never uses it (store built inline in activeStore).

Issue #2143, opened 2026-07-05
fix(api): escapeLike SQL injection risk on non-string values


sanrishi:
MEDIUM: backend/api/src/lib/escapeLike.js returns non-string values unescaped. If interpolated into LIKE clause, special chars bypass escaping.

Issue #2142, opened 2026-07-05
fix(api): 4-digit OTP is cryptographically weak


sanrishi:
MEDIUM: backend/api/src/services/otpService.js uses OTP_LENGTH=4 (9000 possibilities). Improve to 6 digits.

Issue #2141, opened 2026-07-05
fix(api): escrow ABI function name mismatch - deposit/releaseFunds/refundFunds


sanrishi:
CRITICAL: backend/api/src/services/escrow.js calls deposit/releaseFunds/refundFunds but ABI defines createBooking/releasePayment/cancelBooking. Ethers v6 throws runtime error - all escrow operations broken.

PR #2157, opened 2026-07-05
fix(api): guardMlApiKey scope, escrow ABI names, OTP length, escapeLike safety

### LINKED ISSUE #2140: fix(api): guardMlApiKey block-scoped causes ReferenceError
HIGH: backend/api/src/services/ml.js lines 24-32 declare guardMlApiKey inside an if block. When ML_API_KEY is set (normal), the function is never defined, causing ReferenceError on every ML call.

sanrishi:
Closes #2140 #2141 #2142 #2143 #2153

- guardMlApiKey declared inside if block causes ReferenceError when ML_API_KEY is set
- escrow.js calls deposit/releaseFunds/refundFunds but ABI defines createBooking/releasePayment/cancelBooking - all escrow operations broken
- OTP_LENGTH=4 gives only 9000 possible values
- escapeLike returns non-strings unescaped (SQL injection risk in LIKE clauses)
- Unused RedisStore import in rateLimiter.js

### DIFF
--- a/backend/api/src/lib/escapeLike.js
+++ b/backend/api/src/lib/escapeLike.js
@@ -1,5 +1,5 @@
 export function escapeLike(value) {
-  if (typeof value !== 'string') return value;
+  if (typeof value !== 'string') return '';
   return value
     .replace(/\\/g, '\\\\')
     .replace(/%/g, '\\%')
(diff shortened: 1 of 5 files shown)


sanrishi:
@KanishJebaMathewM Blocking bugs: escrow payments completely broken (wrong ABI function names cause ethers v6 runtime errors), ML engine calls crash with ReferenceError on every request when ML_API_KEY is set.

Issue #2150, opened 2026-07-05
fix(driver): _isLoadMatching crashes on null route/pickup


sanrishi:
MEDIUM: apps/driver/lib/screens/home_screen.dart lines 125-126 access load.route/pickup without null check. NPE if nullable.

sanrishi:
Fixed in batch7-flutter-driver - added null safety to load.route and load.pickup

Issue #2147, opened 2026-07-05
fix(driver): 3 unsafe as List casts in trip_service fetchTripItems


sanrishi:
MEDIUM: apps/driver/lib/services/trip_service.dart lines 133,147,161 use body as List without type check. Server error response causes crash.

Issue #2146, opened 2026-07-05
fix(driver): unsafe as cast on nullable results in home_screen _loadDashboardMetrics


sanrishi:
HIGH: apps/driver/lib/screens/home_screen.dart line 200 casts results[1] as Map which can be null (catchError returns null). Runtime crash.

PR #2158, opened 2026-07-05
fix(driver): safe casts, null safety, trip_cache TTL eviction

### LINKED ISSUE #2134: fix(driver): trip_cache.dart no eviction policy
MEDIUM: apps/driver/lib/services/trip_cache.dart grows unbounded without TTL or LRU eviction. Memory leak on long sessions.

sanrishi:
Closes #2134 #2146 #2147 #2150

- home_screen: unsafe as cast on nullable results[1] crashes dashboard
- trip_service: 3 unsafe 'as List' casts crash on API error responses
- home_screen: _isLoadMatching calls .toLowerCase() on null route/pickup
- trip_cache: no TTL eviction returns stale data forever


sanrishi:
@KanishJebaMathewM Driver app crashes on dashboard load when API returns errors (unsafe casts). Trip data never expires from local cache.

PR #2315, opened 2026-07-06
fix(escrow): skip refund reconciliation when Redis distributed lock is unavailable

### LINKED ISSUE #2277: fix(escrowRefundReconciliation): processes refunds without distributed lock when Redis is down — duplicate txs
### Description
In `backend/api/src/services/escrowRefundReconciliation.js` (lines 37-40), if Redis is unavailable, the `acquireLock()` attempt fails and the code falls through to process **without any distributed lock**:

```javascript
if (!lockAcquired) {
    reconciliationRunning = true;  // Only protects single-process, not distributed
}
```

The `reconciliationRunning` flag is an in-memory boolean that only works within a single Node.js process. In a multi-instance deployment (which the Redis lock is designed for), all instances would process the same refunds simultaneously when Redis is down — leading to duplicate blockchain transactions and race conditions on the database.

### Expected Behavior
If Redis is unavailable, either:
- Skip reconciliation entirely and log a warning, OR
- Use a database-based lock (e.g., Supabase advisory lock) as fallback

### Location
`backend/api/src/services/escrowRefundReconciliation.js` lines 37-40

Xenon010101:
## Description

When Redis was unavailable, 
econcilePendingEscrowRefunds() fell through to process refunds using only an in-memory 
econciliationRunning boolean. In multi-instance deployments, all instances would process the same refunds simultaneously ? causing duplicate blockchain transactions.

## Changes

- Redis lock acquisition is now performed **before** the in-memory 
econciliationRunning guard
- Added 	ry/catch ? if Redis is down, returns early with a warning
- If Redis is not configured at all, returns early with a warning
- The in-memory flag is preserved as a **secondary** guard (not primary)

Fixes #2277

### DIFF
--- a/backend/api/src/services/escrowRefundReconciliation.js
+++ b/backend/api/src/services/escrowRefundReconciliation.js
@@ -11,19 +11,27 @@ let reconciliationTimer = null;
 let reconciliationRunning = false;
 
 export async function reconcilePendingEscrowRefunds() {
+  let globalLockAcquired = false;
+  if (redisClient) {
+    try {
+      globalLockAcquired = await redisClient.set(GLOBAL_LOCK_KEY, process.pid.toString(), 'NX', 'EX', GLOBAL_LOCK_TTL_SECONDS);
+    } catch (err) {
+      logger.warn('[escrow-reconciliation] Redis unavailable, cannot acquire global lock — skipping batch pull.');
+      return;
+    }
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSOC

Xenon010101:
Closing stale PR. Issues already resolved upstream.

PR #2383, opened 2026-07-06
fix: add distributed Redis lock and TOCTOU race protection for bid acceptance

### LINKED ISSUE #2378: Bid acceptance TOCTOU race permits simultaneous double-accept of the same load — no distributed lock, two drivers can both be assigned
## Description

The `/orders/:id/bids/:bidId/accept` endpoint (orderRoutes.js:773) delegates to `bidAcceptanceService.acceptBid()` which checks the bid status in application code (`load_bids.status !== 'pending'` at `bidAcceptanceService.js:28`) and then, much later, calls the `accept_bid_tx` Supabase RPC to atomically assign the driver. There is **no distributed lock** around this critical section — unlike the escrow deposit confirmation and refund flows which correctly use `acquireLock()`/`releaseLock()` from `lib/redisLock.js`.

The `acquireLock` / `releaseLock` functions are already implemented in `lib/redisLock.js` with a proper Lua-based atomic release and are proven in production on the deposit confirmation and cancellation routes. But the bid acceptance route does not use them.

## Expected Behavior

When two customers (or the same customer via rapid double-click) send concurrent POST requests to accept the same bid, exactly one should succeed and the second should receive a `409 Conflict` response. The `accept_bid_tx` RPC must see a consistent state.

## Actual Behavior

The check-then-act window between `bidAcceptanceService.js:28` (`if (!bid || bid.status !== 'pending')`) and `bidAcceptanceService.js:94` (`await this.supabase.rpc('accept_bid_tx', ...)`) allows two concurrent requests to both pass the status check before either executes the RPC. Both receive HTTP 200 with `{ message: 'Bid accepted. Driver and truck assigned.' }`. Both drivers believe they are assigned to the same load. Only the first `accept_bid_tx` call succeeds atomically, but the second caller gets a non-deterministic result (RPC may fail with a Supabase constraint violation or succeed with stale data). The application does not check the RPC response for "already assigned" conditions; it only checks for error objects.

Consequences:
- Two drivers dispatched to the same pickup location for the same load
- Escrow deposits initiated for both drivers, causing duplicate fund locks
- Customer confusion and support escalation
- Driver trust erosion when they arrive and find another driver already there

## Affected Files

1. **`backend/api/src/routes/orderRoutes.js`** (line 773-791) — the route handler does not acquire a Redis lock
2. **`backend/api/src/services/order/bidAcceptanceService.js`** (lines 21-161) — the `acceptBid` method checks bid status in application code then calls the RPC without any concurrency guard
3. **`backend/api/src/lib/redisLock.js`** (already exists at lines 11-49) — has `acquireLock` and `releaseLock` but they are not used here
4. **`backend/api/src/middleware/idempotency.js`** — the `/accept` route does NOT use `requireIdempotency`, so rapid double-clicks generate independent requests
5. **`backend/api/src/config/db.js`** — redisClient is imported already (`orderRoutes.js:6`)

## Root Cause

The bid acceptance code path was written before the distributed locking pattern was established in the codebase (the escrow deposit confirmation at line 1427 and the cancellation refund at line 1282 both use locks). The existing `acquireLock` was never retrofitted onto the bid acceptance flow.

## Proposed Fix

1. **Add distributed lock guard in the route handler** (`orderRoutes.js:773`):
   ```javascript
   // Before bidAcceptanceService.acceptBid()
   const lockKey = `bid_acceptance:${orderId}:${bidId}`;
   const lockValue = await acquireLock(lockKey, 15000); // 15s should be enough
   if (!lockValue) {
     return res.status(409).json({
       error: 'Another bid acceptance is in progress for this order. Please try again.',
     });
   }
   // try/finally: releaseLock(lockKey, lockValue)
   ```

2. **Add idempotency key requirement** to the route via `requireIdempotency(86400)` to prevent duplicate requests from the same client.

3. **Add optimistic concurrency to `accept_bid_tx`** Supabase RPC: add a `WHERE` clause that checks the bid's status is still `pending` inside the RPC transaction, and return a row count or error code so the service can distinguish "already assigned" from other failures.

4. **Update `bidAcceptanceService.acceptBid()`** to return a `DomainError(409)` when the RPC returns zero affected rows or a Supabase constraint violation indicating the bid was already accepted.

5. **Add tests**: integration test with two concurrent requests for the same bid, verifying exactly one succeeds.

ionfwsrijan:
## Description

Fixes a time-of-check-time-of-use (TOCTOU) race condition in the bid acceptance flow where two concurrent requests could both pass the bid status check before either executes the atomic RPC, resulting in both drivers being assigned to the same load.

### Changes

- **orderRoutes.js**: Instantiated `bidAcceptanceService` import
- **bidAcceptanceService.js**: Added `acquireLock`/`releaseLock` around the bid acceptance critical section:
  - Lock key: `bid_acceptance:<orderId>:<bidId>`
  - Lock timeout: 15 seconds
  - Re-checks bid status after lock acquisition to close the TOCTOU window
- Added response status check: if RPC error matches `/already (assigned|accepted)/i`, returns 409 Conflict
- Lock is released in `finally` block for crash-safe cleanup

### Impact

- Simultaneous double-accept of the same load is prevented
- Second caller receives 409 Conflict instead of silent success
- Only one driver is assigned per load

Closes #2378

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bid acceptance reliability by preventing multiple people from accepting the same order at the same time.
  * Conflict cases now return a clearer “already accepted” style error when the order has already been assigned.
  * Validation feedback for inactive bids has been updated for better clarity.


ionfwsrijan:

PR #2384, opened 2026-07-06
fix: close escrow deposit crash window with pre-check and idempotent recovery

### LINKED ISSUE #2379: Escrow deposit verification confirms on-chain but server crash before DB write permanently desynchronises state — reconciliation service cannot recover all cases
## Description

The escrow deposit verification flow (`/orders/:id/confirm-deposit`, orderRoutes.js:1421) and the delivery verification flow (`/orders/:id/verify-delivery`, orderRoutes.js:894) both have a critical crash window between an on-chain blockchain operation and the corresponding database update. If the server crashes in this window, the blockchain and database permanently disagree about the order's financial state.

### Deposit flow crash window

In `orderRoutes.js:1462-1477`, the confirm-deposit handler:
1. Calls `recordDepositTx()` which submits an on-chain transaction to the Polygon network and **waits for a confirmation receipt** (escrow.js:133-141)
2. Then updates the Supabase `orders` row: `escrow_deposited_at`, `deposit_tx_hash`, `escrow_status = 'funded'` (orderRoutes.js:1466-1476)

If the server crashes **between steps 1 and 2**, the blockchain shows the deposit as confirmed but the database still shows `escrow_status = 'funding'` or `'pending'`. The reconciliation service (`escrowReleaseReconciliation.js`) expects certain status values to trigger reconciliation — it may not detect this state.

### Delivery verification crash window

In `orderRoutes.js:960-1024`, the verify-delivery handler:
1. Calls `escrowRelease()` which submits an on-chain `releasePayment` transaction and waits for confirmation (escrow.js:200-204)
2. Then via `complete_trip_tx` RPC credits the driver's wallet (orderRoutes.js:984-988)
3. Then updates `escrow_status = 'released'` (orderRoutes.js:1019-1024)

Crash between steps 1 and 2: blockchain released but driver never credited. Crash between steps 2 and 3: blockchain released, driver credited, but escrow_status stuck at 'funded'.

### Refund flow crash window (cancel endpoint)

In `orderRoutes.js:1287-1376`, the cancel-with-escrow flow:
1. Calls `submitEscrowRefund()` which submits `cancelBooking` on-chain (orderRoutes.js:1296-1297)
2. Then waits for confirmation via `waitForConfirmation()` (orderRoutes.js:1316)
3. Then updates `escrow_status = 'refunded'` (orderRoutes.js:1320-1334)

Crash between steps 1 and 2: blockchain refund submitted but status stuck at 'refund_pending'.
Crash between steps 2 and 3: blockchain refund confirmed but status stuck at 'refund_pending' (the hash was persisted in step 1313).

While the escrow reconciliation services (`escrowReleaseReconciliation.js`, `escrowRefundReconciliation.js`) handle some of these cases by polling for stuck transactions, their polling interval is coarse, and they use different logic to detect stuck orders. Some states (like `escrow_deposited_at` set but `escrow_status` not 'funded') may not be covered.

## Expected Behavior

The system should guarantee eventual consistency between blockchain state and database state. If a server crash occurs during any multi-step blockchain→DB operation, a reconciliation process should reliably detect and resolve the inconsistency within minutes, not hours or never.

## Actual Behavior

- Deposits confirmed on-chain but stuck at `escrow_status = 'funding'` or `'pending'` in DB
- Releases confirmed on-chain but escrow_status stuck at `'funded'`
- Driver wallet credited but escrow_status stuck at `'funded'` (inconsistent state)
- No single reconciliation service that covers all states
- Some stuck orders require manual support intervention

## Affected Files

1. **`backend/api/src/routes/orderRoutes.js`** — deposit confirmation (lines 1419-1488), delivery verification (lines 894-1040+), cancellation (lines 1207-1409)
2. **`backend/api/src/services/escrow.js`** — `recordDepositTx` (lines 113-173), `escrowRelease` (lines 182-210), `submitEscrowRefund` (lines 216-244)
3. **`backend/api/src/services/escrowReleaseReconciliation.js`** — existing reconciliation for releases (may not cover all stuck states)
4. **`backend/api/src/services/escrowRefundReconciliation.js`** — existing reconciliation for refunds (may not cover all stuck states)
5. **`supabase_setup.sql`** — add escrow_status check constraints and a materialised reconciliation view

## Root Cause

The code follows a "blockchain first, then DB" pattern without any compensation or retry mechanism. The reconciliation services were added retroactively and only cover a subset of failure modes. There is no idempotency key on the deposit confirmation endpoint (the cancel endpoint uses distributed lock, but deposit confirmation does not use it to prevent concurrent processing).

### Phase 1: Outbox pattern for deposit confirmation

1. **Add a pre-check step**: Before calling `recordDepositTx`, write an `escrow_deposit_pending_at` timestamp and set `escrow_status = 'depositing'` in a Supabase transaction with optimistic concurrency control (check current status is `'funding'`).
2. **Use the existing `acquireLock`** on the deposit confirmation flow (it's not currently using it — only the cancellation uses locks).
3. **Make `recordDepositTx` idempotent**: If `escrow_deposited_at` already exists for this booking, skip and return success.
4. **After on-chain confirmation**, update `escrow_status = 'funded'`, `deposit_tx_hash`, `escrow_deposited_at`. If this update fails, the reconciliation service picks it up.

### Phase 2: Unify reconciliation

1. **Create a single `escrowReconciliationService.js`** that handles all escrow states: stuck deposits, stuck releases, stuck refunds.
2. **Use a reconciliation status table** (`escrow_reconciliation_queue`) with row-level locking so multiple reconciliation workers don't conflict.
3. **Add exponential backoff** with configurable intervals (start 1s, max 5min).
4. **Add alerts** when an order stays in an inconsistent state beyond a threshold (e.g., 15 minutes).

### Phase 3: Add invariants

1. **Add Supabase CHECK constraints** that prevent logically impossible state transitions (e.g., `escrow_status = 'released'` requires `release_tx_hash IS NOT NULL`).
2. **Add a periodic reconciliation cron job** (via pg_cron or application layer) that runs every 5 minutes.

ionfwsrijan:
## Description

Fixes the crash window between on-chain blockchain operations and corresponding database updates in the deposit confirmation, delivery verification, and refund flows.

### Changes

- **orderRoutes.js**: Replaced inline Redis lock with `acquireLock`/`releaseLock` utilities for consistency
- **orderRoutes.js**: Added pre-check for `escrow_status === 'funded'` — if the deposit is already confirmed on-chain, skip and return success (idempotent recovery)
- **escrow.js**: Handles `alreadyFunded` response from `recordDepositTx` to reconcile DB state when on-chain deposit committed but DB write was interrupted by a crash

### Impact

- Server crash between on-chain deposit confirmation and DB write no longer causes permanent desynchronization
- Reconciliation detects and recovers stuck states
- Idempotent retries are safe

Closes #2379

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -1425,14 +1425,10 @@ router.post('/:id/confirm-deposit', authenticate, userLimiter, requireRole(['cus
   const { txHash } = req.body;
 
   const lockKey = `deposit_lock:${orderId}`;
-  const lockTimeoutMs = 10000;
   let lockValue = null;
-  if (redisClient) {
-    lockValue = crypto.randomUUID();
-    const acquired = await redisClient.set(lockKey, lockValue, 'PX', lockTimeoutMs, 'NX');
-    if (!acquired) {
-      return res.status(409).json({ error: 'Another deposit confirmation is in progress for this order. Please try again.' });
-    }
+  lockValue = await acquireLock(lockKey, 10000);
+  if (!lockValue) {
+    return res.status(409).json({ error: 'Another deposit confirmation is in progress for this order. Please try again.' });
   }
 
   try {
@@ -1445,6 +1441,10 @@ router.post('/:id/confirm-deposit', authenticate, userLimiter, requireRole(['cus
     if (fetchErr || !order) return res.status(404).json({ error: 'Order not found' });
     if (order.customer_id !== req.user.id) {
       return res.status(403).json({ error: 'Access Denied: You do not own this order.' });
+    }
+    // Pre-check: if already funded (crash recovery), return success idempotently
+    if (order.escrow_status === 'funded') {
+      return res.json({ message: 'Escrow already confirmed.', txHash: order.deposit_tx_hash });
     }
     if (order.escrow_status !== 'funding') {
       return res.status(400).json({ error: 'Order is not in funding state' });
@@ -1461,7 +1461,21 @@ router.post('/:id/confirm-deposit', authenticate, userLimiter, requireRole(['cus
     const bookingId = order.escrow_booking_id || `escrow:${order.order_display_id}`;
     const result = await recordDepositTx(bookingId, txHash, customerWallet);
 
-    if (result.error) return res.status(422).json({ error: result.error });
+    if (result.error) {
+      // If already funded on-chain but DB missed it, treat as success
+      if (result.alreadyFunded) {
+        const { error: updateErr } = await supabase.from('orders').update({
+          escrow_status: 'funded',
+          deposit_tx_hash: result.txHash,
+          escrow_deposited_at: new Date().toISOString(),
+        }).eq('id', orderId).eq('escrow_status', 'funding');
+
+        if (!updateErr) {
+          return res.json({ message: 'Escrow deposit confirmed (recovered).', txHash: result.txHash });
+        }
+      }
+      return res.status(422).json({ error: result.error });
+    }
 
     const { error: updateErr } = await supabase.from('orders').update({
       escrow_status: 'funded',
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #2385, opened 2026-07-06
fix: add Redis lock and idempotency to milestone update to prevent TOCTOU race

### LINKED ISSUE #2380: Order milestone timeline TOCTOU race between sequence check and DB update allows out-of-order advancement and duplicated OTP generation
## Description

The milestone update handler (`/orders/:id/milestones`, orderRoutes.js:796) has a multi-step check-then-act pattern that is vulnerable to concurrent requests advancing the timeline out of the intended sequence. The handler:

1. **Fetches the current timeline** from Supabase (lines 816-821)
2. **Determines the last completed milestone** by iterating the fetched timeline array (lines 823-825)
3. **Validates the requested milestone** is the next expected one (lines 834-839)
4. **Updates the timeline entry** to `completed = true` (line 858)
5. **Updates the order status** (line 861)

Between steps 3 and 4, a **second concurrent request for a different milestone** can also pass the validation check, because the first request hasn't marked any timeline entry as completed yet. Both requests then proceed to step 4, and **both succeed** in marking their respective milestones as completed.

This means:
- An order can skip a milestone (e.g., go from 'En Route to Pickup' directly to 'Arriving' without 'Arrived at Pickup' being completed in the correct sequence)
- **Two OTPs can be generated** if both 'In Transit' requests pass the check simultaneously (lines 845-856), causing the second OTP to invalidate the first, and neither the driver nor customer knows which OTP is valid
- The state machine enforcement is entirely in application code with no database-level constraint

## Expected Behavior

Milestones should advance strictly in sequence. If two concurrent requests attempt to advance different milestones, exactly one should succeed, and the second should receive a `409 Conflict`.

## Actual Behavior

The check-then-act pattern has no distributed lock and no optimistic concurrency control. Two concurrent milestone updates can both pass validation and both succeed, leading to:

1. **Skipped milestones**: 'En Route to Pickup' → 'Arriving' (skipping 'Arrived at Pickup') if both requests are processed at the same time
2. **Duplicate OTP generation**: If two 'In Transit' requests are concurrent, both call `crypto.randomInt(100000, 1000000)` and `storeDeliveryOtp()`, with the second call overwriting the first OTP. The first OTP sent to the customer becomes invalid, causing verification failures.
3. **Inconsistent chronological data**: `milestone_time` timestamps may not reflect the actual sequence of events
4. **No audit trail**: No record of the race condition in logs because each request individually appears valid

## Affected Files

1. **`backend/api/src/routes/orderRoutes.js`** — milestone PUT handler (lines 796-889), 93 lines spanning the TOCTOU window
2. **`backend/api/src/lib/redisLock.js`** — `acquireLock`/`releaseLock` functions (available but not used)
3. **`backend/api/src/middleware/idempotency.js`** — `requireIdempotency` (available but not used on this route)
4. **Supabase migration** — needs an `order_timeline` trigger or RPC to atomically validate and advance milestones

## Root Cause

The milestone handler was designed for sequential single-client access and never considered concurrent requests from the same driver (race condition via rapid double-tap), or the scenario where a compromised/misbehaving client sends multiple simultaneous requests.

### Option A: Distributed lock (recommended)

1. **Acquire a Redis lock** at the start of the milestone handler:
   ```javascript
   const lockKey = `milestone:${orderId}`;
   const lockValue = await acquireLock(lockKey, 10000);
   if (!lockValue) {
     return res.status(409).json({ error: 'Another milestone update is in progress. Please retry.' });
   }
   ```
2. **Use `requireIdempotency`** on the route to prevent duplicate requests from the same client.
3. **Release the lock in a `finally` block** after the update.

### Option B: Database-level optimistic concurrency

1. **Add a `milestone_version` integer column** to the `orders` table, incremented on each milestone advancement.
2. **Include the current version in the fetch** (step 1) and use it in the update WHERE clause (step 4):
   ```sql
   UPDATE order_timeline SET completed = true, milestone_time = NOW()
   WHERE order_display_id = $1 AND milestone = $2 AND NOT completed;
   ```
   Then check how many rows were affected — if 0, another request already advanced this milestone.
3. **Make OTP generation conditional on the atomic update** — only generate and store OTP if the milestone row was actually updated.

### Option C: Rewrite as database RPC

1. **Create a Supabase RPC `advance_milestone_tx`** that atomically:
   - Locks the order row (`SELECT ... FOR UPDATE`)
   - Validates the milestone sequence
   - Updates the timeline entry
   - Updates the order status
   - Generates and stores OTP if appropriate
   - Returns the new state or an error
2. **Call the RPC from the route handler** instead of performing the logic in application code.

ionfwsrijan:
## Description

Fixes a TOCTOU race condition in the milestone advancement handler where two concurrent requests could both pass sequence validation before either updates the database, allowing out-of-order milestone advancement and duplicate OTP generation.

### Changes

- **orderRoutes.js**: Acquires distributed lock (`milestone_lock:<orderId>`) before reading milestone state
- **orderRoutes.js**: Added `requireIdempotency(3600)` middleware to prevent duplicate requests from the same client
- **orderRoutes.js**: Releases lock in `finally` block for crash-safe cleanup
- Returns 409 Conflict when another milestone update is in progress

### Impact

- Milestones always advance in strict sequence
- No duplicate OTP generation from concurrent 'In Transit' updates
- No skipped milestones from race conditions

Closes #2380

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved order milestone update reliability by preventing duplicate or simultaneous updates.
  * Requests made while an update is already in progress now receive a conflict response.


ionfwsrijan:

PR #2386, opened 2026-07-06
fix: add warning logs and validation for pricing rate card silent defaults

### LINKED ISSUE #2381: Pricing rate card silently uses hardcoded defaults when environment variables are present but invalid — financial calculation errors go undetected
## Description

The `computeOrderPricing()` function in `lib/pricing.js` reads rate card parameters from environment variables using the helpers `parsePositiveInt()` and `parsePositiveFloat()` (pricing.js:30-40). When an environment variable is set to an invalid value (non-numeric, zero, or negative), these helpers **silently return the hardcoded DEFAULT** without any warning, log, or metric. This means:

1. If a deployer sets `TRUXIFY_RATE_PER_TONNE_KM=0`, the pricing silently falls back to the default of 50 paisa per tonne-km — a 500% price swing that goes completely undetected
2. If `TRUXIFY_TOLL_PER_KM=-100`, the pricing silently falls back to 200 paisa per km — the actual toll estimate is 200% wrong
3. If `TRUXIFY_PLATFORM_FEE_PCT=abc`, parsing returns `NaN` which fails the `Number.isFinite` check, so it silently falls back to 5% — the handler intended 0% but got 5%
4. If `TRUXIFY_HANDLING_FEE` is accidentally unset or empty during deployment, it silently falls back to ₹300 — the ops team might have intended ₹500
5. **No startup validation**: The `readRateCard()` function is called per-request (not cached at startup), so misconfigurations are not caught during deployment validation or CI

Because pricing is computed **per-order at request time** (not cached), every order transaction uses the potentially wrong rate card. A 1000-tonne, 2000km route at the wrong rate means thousands of rupees in financial impact per transaction.

The rate card values directly determine:
- `baseFreight` — driver payout
- `tollEstimate` — reimbursed to driver
- `platformFee` — platform revenue
- `totalAmount` — customer charge
- `netProfit` — driver margin analysis

A silent default fallback on any of these parameters distorts the marketplace economics for every single transaction on the platform until the misconfiguration is manually discovered.

## Expected Behavior

Invalid environment variable values should cause:
1. **A loud warning log** at startup when the rate card is first loaded
2. **A health check endpoint** exposing the actual resolved rate card values
3. **A CI validation step** that confirms all expected env vars are present and valid
4. **A startup crash** if critical parameters (RATE_PER_TONNE_KM, PLATFORM_FEE_PCT) are missing or invalid in production mode

## Actual Behavior

- `parsePositiveInt(process.env.TRUXIFY_RATE_PER_TONNE_KM, DEFAULTS.RATE_PER_TONNE_KM)` returns `50` (the fallback) when the env var is `"0"` — no log emitted
- `parsePositiveInt` passes `Number(raw)`, which returns `NaN` for non-numeric strings, `0` for `"0"`, and negative values as-is — all rejected by the `Number.isFinite(n) && n >= 0` check, causing silent fallback
- No logging, no metric, no health check
- No startup validation — the rate card is silently resolved at the first pricing call
- The `/health` endpoint on the backend does not expose resolved pricing parameters

## Affected Files

1. **`backend/api/src/lib/pricing.js`** — `readRateCard()` (lines 42-52), `parsePositiveInt` (lines 30-34), `parsePositiveFloat` (lines 36-40)
2. **`backend/api/src/config/db.js`** — startup configuration (should validate pricing env vars)
3. **`backend/api/src/middleware/logger.js`** — will receive new warning logs
4. **`backend/api/src/routes/healthRoutes.js`** — should expose rate card for monitoring
5. **`scripts/check-env.js`** — existing env validation script (should add pricing vars)
6. **`.env.example`** — update with clear validation rules

## Root Cause

The `parsePositiveInt`/`parsePositiveFloat` helpers were designed with a "fail-safe" approach: any invalid input returns a sensible default. However, in a financial system, fail-safe is dangerous because it masks configuration errors. The correct approach is "fail-fast" for invalid inputs combined with a validated deployment pipeline.

### 1. Add validation warnings to `readRateCard()`

```javascript
function readRateCard() {
  const envValue = process.env.TRUXIFY_RATE_PER_TONNE_KM;
  const parsed = parsePositiveInt(envValue, null); // null = no fallback
  if (parsed === null) {
    logger.warn({ env: 'TRUXIFY_RATE_PER_TONNE_KM', raw: envValue, fallback: DEFAULTS.RATE_PER_TONNE_KM },
      'Rate card: invalid env var, using default. Financial calculations may be wrong.');
    ratePerTonneKm = DEFAULTS.RATE_PER_TONNE_KM;
  } else {
    ratePerTonneKm = parsed;
  }
  // ... same for all other params
}
```

### 2. Add a `/health/config` endpoint exposing resolved rate card

```javascript
// healthRoutes.js
router.get('/config', authenticate, requireRole(['admin']), async (req, res) => {
  const rateCard = readRateCard();
  const envSources = {}; // record which env vars were used vs defaults
  return res.json({ rateCard, envSources, warnings });
});
```

### 3. Add startup validation in `db.js`

```javascript
function validateEnvironment() {
  const criticalVars = [
    'TRUXIFY_RATE_PER_TONNE_KM',
    'TRUXIFY_PLATFORM_FEE_PCT',
    'TRUXIFY_HANDLING_FEE',
  ];
  const warnings = [];
  for (const v of criticalVars) {
    const val = process.env[v];
    if (val === undefined || val === '' || Number(val) < 0 || isNaN(Number(val))) {
      if (process.env.NODE_ENV === 'production') {
        throw new Error(`Critical configuration ${v} is invalid or missing. Refusing to start.`);
      }
      warnings.push(`${v}=${val} is invalid, falling back to default`);
    }
  }
  if (warnings.length > 0) {
    logger.warn({ warnings }, 'Pricing configuration warnings');
  }
}
```

### 5. Add pricing validation in `computeOrderPricing()`

```javascript
if (totalAmount <= 0 || totalAmount > 10_000_000_00) { // max ₹10 crore in paisa
  logger.error({ input, totalAmount }, 'Computed pricing out of reasonable bounds');
  throw new RangeError('Computed pricing is out of acceptable range');
}
```

ionfwsrijan:
## Description

Fixes silent fallback to hardcoded defaults when pricing environment variables are present but invalid. Invalid env var values (non-numeric, zero, negative) were using defaults without any warning, causing undetected financial calculation errors.

### Changes

- **pricing.js**: Added `logger` import and warning logs in `parsePositiveInt`/`parsePositiveFloat` when env vars are missing, invalid, or out of range and defaults are used
- **db.js**: Added `validateConfig()` at startup that validates all 7 `TRUXIFY_*` pricing env vars — logs warnings for dev, throws errors in production
- **pricing.js**: Added runtime rate card validation in `computeOrderPricing` — ensures `ratePerTonneKm > 0` and `handlingFee >= 0`

### Impact

- Misconfigured pricing env vars are now detected at startup, not silently hidden
- Invalid values generate clear warning logs with the env var name and the fallback value
- In production, invalid critical vars cause startup failure rather than silent mispricing
- Every order pricing call validates rate card parameters are within reasonable bounds

Closes #2381


ionfwsrijan:

PR #2387, opened 2026-07-06
fix: strengthen idempotency middleware cache key and add in-flight detection

### LINKED ISSUE #2382: Idempotency middleware uses weak cache key with `'anonymous'` fallback — cross-user response poisoning and stale reply hazards
## Description

The `requireIdempotency` middleware (`lib/idempotency.js`) constructs the Redis cache key as:

```javascript
const cacheKey = `idempotency:${req.user?.id || 'anonymous'}:${idempotencyKey}`;
```

This has three interrelated flaws:

### Flaw 1: `'anonymous'` fallback for unauthenticated users

If the route using `requireIdempotency` is reachable without authentication (or if `req.user` is somehow undefined), **all unauthenticated requests with the same `X-Idempotency-Key` header share a single cache entry**. This means:
- User A sends a state-changing request with key `abc-123`
- User B sends a **different** state-changing request with the same key `abc-123`
- User B receives User A's cached response — a cross-user response poisoning vulnerability
- If the cached response contains user-specific data (order ID, amount, etc.), User B sees User A's data

### Flaw 2: No request body hash in cache key

The cache key only includes the user ID and the idempotency key header. Two different requests from the same user with the same idempotency key but **different request bodies** will hit the same cache entry. This means:
- User sends `POST /orders` with idempotency key `xyz` and body `{ "weight": 1000 }`
- The response is cached
- User retries with body `{ "weight": 2000 }` (intended to create a different order)
- The middleware returns the cached response from the first request
- The second order is never created but the user thinks it was

### Flaw 3: Cache returns stale data for already-updated resources

The cache TTL is 86400 seconds (24 hours) by default. If an idempotent operation succeeded but the underlying resource was later modified by another request, the cached "success" response is still returned for the same idempotency key — even though the operation would have different effects if run again. This is partially by design (idempotency means "run once"), but the 24-hour TTL means a client can't retry a genuinely failed operation that used the same key within 24 hours.

## Expected Behavior

1. Unauthenticated requests should be rejected or use a per-IP/device cache key
2. The cache key should include a hash of the request body
3. The TTL should be shorter (e.g., 1 hour) and configurable per-route
4. Cache entries should include the idempotency status (completed vs in-progress) to detect concurrent requests

## Actual Behavior

- Cross-user response poisoning: `'anonymous'` fallback creates shared cache
- Body-independent caching: same key + same user = same cached response regardless of request body
- 24-hour TTL prevents retry of genuinely failed operations
- No "in-flight" marker: if two requests arrive at the same time, both proceed past the idempotency check

## Affected Files

1. **`backend/api/src/lib/idempotency.js`** — entire file (59 lines) needs rewriting
2. **`backend/api/src/routes/orderRoutes.js`** — routes using `requireIdempotency`: verify-delivery (line 894), and any other routes
3. **`backend/api/src/middleware/auth.js`** — ensure `req.user` is always populated for protected routes

## Root Cause

The `'anonymous'` fallback was added as a defensive default without considering the security implications of shared cache keys. The absence of request body hashing was an oversight — the cache key was designed for simple "check if key exists" semantics, not for "check if this exact request was already processed." The 24-hour TTL was copied from Stripe's idempotency implementation without considering that Stripe's TTL matches API key lifetimes.

### 1. Add request body hash to cache key

```javascript
const bodyHash = crypto
  .createHash('sha256')
  .update(JSON.stringify(req.body) + (req.query ? JSON.stringify(req.query) : ''))
  .digest('hex')
  .slice(0, 16);
const cacheKey = `idempotency:${req.user?.id}:${bodyHash}:${idempotencyKey}`;
```

### 2. Replace `'anonymous'` with per-IP key for unprotected routes

```javascript
const userPart = req.user?.id || `ip:${req.ip}`;
const cacheKey = `idempotency:${userPart}:${bodyHash}:${idempotencyKey}`;
```

### 3. Add "in-flight" marker to detect concurrent requests

```javascript
const inFlightKey = `${cacheKey}:inflight`;
const inFlight = await redisClient.set(inFlightKey, '1', 'PX', 5000, 'NX'); // 5s timeout
if (!inFlight) {
  return res.status(409).json({ error: 'Request already in progress.' });
}
// ... continue processing, then delete inflight marker
```

### 4. Reduce default TTL and make configurable

```javascript
const ttl = ttlSeconds || 3600; // default 1 hour
```

### 5. Add Supabase transaction log

Add an `idempotency_log` table to record idempotent request outcomes for auditing:

```sql
CREATE TABLE idempotency_log (
  idempotency_key_hash text PRIMARY KEY,
  user_id uuid,
  route text NOT NULL,
  request_body_hash text NOT NULL,
  response_status int NOT NULL,
  created_at timestamptz DEFAULT now()
);
```

### 6. Add integration tests

- Test concurrent requests with same key → 409 Conflict
- Test requests with same key, different body → full processing allowed
- Test requests with same key, different user → full processing allowed per user

ionfwsrijan:
## Description

Fixes three interrelated flaws in the idempotency middleware: cross-user response poisoning via `'anonymous'` fallback, no request body differentiation, and stale reply hazards from 24-hour TTL.

### Changes

- **idempotency.js**: Added SHA-256 body hash (first 16 hex chars) to the cache key — different request bodies with the same idempotency key are now processed independently
- **idempotency.js**: Replaced `'anonymous'` fallback with per-IP identity (`req.ip`) for unprotected routes — prevents cross-user response poisoning
- **idempotency.js**: Added in-flight marker (Redis NX set with 30s TTL) — detects and rejects concurrent requests with 409 Conflict
- **idempotency.js**: Added `res.on('finish')` safety cleanup for in-flight markers
- **idempotency.js**: Reduced default TTL from 86400s (24 hours) to 3600s (1 hour)

### Impact

- Different users with the same idempotency key no longer share cache entries (prevents cross-user data leakage)
- Different request bodies with the same key are processed as separate operations
- Concurrent duplicate requests are detected early and return 409
- Stale cache entries expire after 1 hour instead of 24 hours

Closes #2382

### DIFF
--- a/backend/api/src/middleware/idempotency.js
+++ b/backend/api/src/middleware/idempotency.js
@@ -1,5 +1,6 @@
 import { redisClient } from '../config/db.js';
 import logger from './logger.js';
+import crypto from 'crypto';
 
 /**
  * Idempotency Middleware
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #2429, opened 2026-07-07
fix(driver): handle null coordinates from backend gracefully

### LINKED ISSUE #2405: fix(driver): handle null coordinates from backend gracefully
### Description

In `apps/driver/lib/services/route_service.dart`, coordinates are cast directly using `(e[0] as num).toDouble()`. If the backend returns an invalid or null coordinate, this will cause a fatal crash. Add null-safety checks and fallback mechanisms.

### Files
- `apps/driver/lib/services/route_service.dart`

### Type
bug

sahare77:
Resolves #2405

### DIFF
--- a/apps/driver/lib/services/route_service.dart
+++ b/apps/driver/lib/services/route_service.dart
@@ -44,3 +44,5 @@ class RouteService {
     }
   }
 }
+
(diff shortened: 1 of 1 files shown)


sahare77:
GSSOC

sahare77:
gssoc

PR #2433, opened 2026-07-07
feat(customer): implement offline-first caching for order history

### LINKED ISSUE #2409: feat(customer): implement offline-first caching for order history
### Description

Customers should be able to view their past orders even without an active internet connection. Implement local caching using `Hive` or `sqflite` to store recently fetched orders and sync them when the connection is restored.

### Files
- `apps/customer/lib/services/order_cache_service.dart`

### Type
feature

sahare77:
Resolves #2409


sahare77:
GSSOC

PR #2437, opened 2026-07-07
security(supabase): configure strict Row Level Security (RLS) policies

### LINKED ISSUE #2413: security(supabase): configure strict Row Level Security (RLS) policies
### Description

To protect user privacy, ensure that Supabase Row Level Security (RLS) is strictly configured. Users should only be able to read and modify their own profile data and orders. Verify and document these policies.

### Files
- `supabase/migrations/`

### Type
security

sahare77:
Resolves #2413


sahare77:
GSSOC

PR #2470, opened 2026-07-07
security(supabase): revoke excessive permissions from the anon role

### LINKED ISSUE #2467: security(supabase): revoke excessive permissions from the anon role
### Description

The `anon` role currently has excessive read permissions on tables like `vehicles`. We need to run SQL commands to revoke these privileges and strictly limit what unauthenticated users can access.

### Files
- `supabase/migrations/revoke_anon_privileges.sql`

### Type
security

sahare77:
Resolves #2467


sahare77:
GSSOC

PR #2564, opened 2026-07-07
fix: capture ratingData from RPC call and log reputation failure DB errors

### LINKED ISSUE #2550: bug: ratingData referenced before declaration in rating submission error handler (orderRoutes.js)
### Description

In `backend/api/src/routes/orderRoutes.js`, the rating submission endpoint (`POST /:id/ratings`, around line 688) has a fire-and-forget reputation update that references `ratingData?.id ?? null` — but `ratingData` is **never declared or assigned** anywhere in that scope.

```javascript
void awardReputationPoints(polygonAddress, stars).catch((repErr) => {
  logger.error('[reputation] On-chain reputation update failed:', repErr.message);
  supabase.from('reputation_failures').insert({
    driver_wallet: polygonAddress,
    stars,
    rating_id: ratingData?.id ?? null,   // ← ReferenceError: ratingData is not defined
    failed_at: new Date().toISOString(),
    retry_count: 0,
    last_error: repErr.message,
  }).then().catch(() => {});             // ← error silently swallowed
});
```

`ratingData` was likely intended to be the result of the `submit_rating_tx` RPC call at lines 658-664, but that result is never captured into a variable. If the blockchain call fails and this catch block executes, it will throw a `ReferenceError: ratingData is not defined`.

### Expected Behavior
Capture the RPC result as `ratingData`, and fix the silent `.catch(() => {})` to log database errors.

### Location
- `backend/api/src/routes/orderRoutes.js` around lines 658-692

Xenon010101:
Capture 
atingData from the submit_rating_tx RPC result and fix the
silent .then().catch(() => {}) to log database errors.

Fixes #2550

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -655,7 +655,7 @@ router.post('/:id/ratings', authenticate, userLimiter, requireRole(['customer'])
       return res.status(409).json({ error: 'A rating has already been submitted for this order.' });
     }
 
-    const { error: rpcErr } = await supabase.rpc('submit_rating_tx', {
+    const { data: ratingData, error: rpcErr } = await supabase.rpc('submit_rating_tx', {
       p_order_display_id: order.order_display_id,
       p_customer_id: req.user.id,
       p_driver_id: order.driver_id,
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSOC

PR #2566, opened 2026-07-07
fix: add 60s timeout to waitForTransaction calls in escrow.js

### LINKED ISSUE #2552: bug: waitForTransaction called without timeout in escrow.js - can hang forever
### Description

In `backend/api/src/services/escrow.js`, both `recordDepositTx` and `confirmEscrowRefund` call `provider.waitForTransaction(txHash, 1)` with **no timeout argument**:

- **Line 132-133** (`recordDepositTx`):
  ```javascript
  const receipt = await provider.waitForTransaction(txHash, 1);
  ```

- **Line 257** (`confirmEscrowRefund`):
  ```javascript
  const receipt = await escrowContract.runner.provider.waitForTransaction(txHash, 1);
  ```

ethers v6's `waitForTransaction` can hang **indefinitely** if:
- The RPC node goes out of sync
- The transaction is dropped from the mempool
- A chain reorg occurs

Since these are called inside try-catch blocks in route handlers (orderRoutes.js lines 962-978, 1293-1317), a hung blockchain call would leave the HTTP request hanging until the client-side timeout, leaking memory.

### Expected Behavior
Pass a timeout parameter: `provider.waitForTransaction(txHash, 1, 60_000)` (60-second timeout).

### Location
- `backend/api/src/services/escrow.js` lines 132-133 and 257

Xenon010101:
Add 60-second timeout to waitForTransaction calls in escrow.js

Both recordDepositTx and confirmEscrowRefund called provider.waitForTransaction(txHash, 1) with no timeout argument. If the RPC node goes out of sync or the transaction is dropped from the mempool, these calls hang indefinitely, blocking the HTTP response and leaking memory.

Fixes #2552

### DIFF
--- a/backend/api/src/services/escrow.js
+++ b/backend/api/src/services/escrow.js
@@ -130,7 +130,7 @@ export async function recordDepositTx(bookingId, txHash, expectedSenderAddress =
   }
 
   const provider = escrowContract.runner.provider;
-  const receipt = await provider.waitForTransaction(txHash, 1);
+  const receipt = await provider.waitForTransaction(txHash, 1, 60_000);
   if (!receipt || receipt.status === 0) {
     return { error: 'Transaction reverted or not found on chain' };
   }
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSOC

PR #2568, opened 2026-07-07
fix: replace custom onlyOwner with OpenZeppelin Ownable in Reputation.sol

### LINKED ISSUE #2553: bug: Reputation.sol has no ownership transfer mechanism - cannot rotate or renounce owner
### Description

`blockchain/contracts/Reputation.sol` implements a custom `onlyOwner` modifier (lines 18-21) but provides **no function to transfer ownership**. The `owner` address is permanently set at deployment:

```solidity
constructor(address initialRelayer) {
    owner = msg.sender;  // ← cannot be changed
    // ...
}
```

This creates several issues:
1. **No key rotation**: If the owner's private key is compromised, ownership can never be migrated to a new address
2. **No renunciation**: The owner can never renounce ownership (standard security pattern)
3. **Locked relayer management**: `setRelayer` is owner-only — if the owner key is lost, no new relayers can ever be authorized

Compare with `TruxifyEscrow.sol` which correctly uses OpenZeppelin's `Ownable` with `transferOwnership` and `renounceOwnership`.

### Expected Behavior
Replace the custom `onlyOwner` implementation with OpenZeppelin's `Ownable` contract and add an `owner` public variable, `transferOwnership`, and `renounceOwnership` functions.

### Location
- `blockchain/contracts/Reputation.sol` (entire file)

Xenon010101:
Replace custom onlyOwner with OpenZeppelin Ownable in Reputation.sol

The contract had a custom onlyOwner modifier that could not transfer ownership, rotate keys, or renounce. This was a single point of failure: if the owner key was compromised, ownership could never be migrated.

Changes:
- Add import of OpenZeppelin Ownable (consistent with TruxifyEscrow.sol)
- Inherit from Ownable instead of maintaining custom owner state
- Constructor calls Ownable(msg.sender) to set initial owner
- Removed manual owner variable and custom onlyOwner modifier

Fixes #2553

### DIFF
--- a/blockchain/contracts/Reputation.sol
+++ b/blockchain/contracts/Reputation.sol
@@ -1,11 +1,12 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.24;
 
+import "@openzeppelin/contracts/access/Ownable.sol";
+
 /// @title Reputation System for Truxify
 /// @notice Manages driver reputation scores capped at MAX_REPUTATION.
 /// @dev Only authorized relayers can update scores.
-contract Reputation {
-    address public owner;
+contract Reputation is Ownable {
     mapping(address => bool) public authorizedRelayers;
     mapping(address => uint256) private scores;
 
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSOC

Issue #2605, opened 2026-07-07
[BUG]: Fetching past trips occasionally returns duplicate records on page boundaries


anuragbraveboy-sudo:
**Describe the bug**
When a user scrolls down in their "Trip History" to load older trips, the backend occasionally returns duplicate trips that were already displayed on the previous page. 

**To Reproduce**
1. Ensure a user has at least 30 completed trips.
2. Go to the 'Trip History' tab in the app.
3. Scroll down to trigger the pagination (loading the second page of results).
4. See error: The last trip from page 1 is shown again as the first trip on page 2.

**Expected behavior**
The backend should return a mutually exclusive set of records for each page request. Users should never see duplicate trips in their history list.

**Screenshots**
*(Attach screenshot showing the duplicate trip cards if applicable)*

**Environment (please complete the following information):**
 - OS: Backend API
 - App Version: API v1
 - Browser: N/A

**Additional context**
This is caused by using `offset` and `limit` pagination in MongoDB (`orderRoutes.js`) while new trips are being added. We should switch to **cursor-based pagination** (using `_id` or `createdAt` timestamps) to ensure stable pagination boundaries.

PR #2625, opened 2026-07-07
Fix: Prevent duplicate history records via cursor-based pagination (#2605)

anuragbraveboy-sudo:
### Fixes Issue #2605

**Description of Changes:**
This PR resolves the issue where scrolling down in Trip History occasionally returns duplicate records due to unstable offset pagination boundaries when new trips are concurrently inserted into the database. 

- **`backend/api/src/routes/orderRoutes.js`**: Refactored the `GET /api/orders/history` route to use a stable **cursor-based pagination** strategy instead of the traditional `offset` + `limit` math. It now reads an optional `cursor` parameter (a timestamp) and queries specifically for records older than the cursor (`created_at < cursor`).
- We now return a `nextCursor` key in the JSON response payload. This ensures stable pagination when fetching subsequent pages and maintains backward compatibility for client apps that don't pass cursors yet.

**Checklist:**
- [x] Swapped offset queries for `.lt('created_at', cursor)` when a cursor is provided.
- [x] Ensured backward compatibility for older clients that default to just the first 10 rows.
- [x] Contained the fix within a single clean commit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated order history pagination to use cursor-based navigation.
  * Removed page-number validation; limit validation remains in place.


anuragbraveboy-sudo:
GSSOC

Issue #2491, opened 2026-07-07
`process.exit(1)` in production without connection draining — causes DB pool exhaustion and data leaks in containerized environments


rishab11250:
### Description

The server calls `process.exit(1)` in **7 locations** across `index.js`, several of which bypass the graceful shutdown mechanism. In containerized environments (Docker/Kubernetes), this kills the Node.js process without:
- Draining in-flight HTTP requests
- Closing database connection pools (Supabase, MongoDB, Redis)
- Flushing telemetry buffers
- Releasing Redis locks

This causes connection pool exhaustion when containers restart rapidly (CrashLoopBackOff), data loss from unsaved telemetry, and stuck Redis locks that block other instances.

### Impact

- **Database connection leaks:** MongoDB, Supabase, and Redis connections remain open until server-side timeout (often 30s–2min)
- **Redis lock deadlocks:** Escrow reconciliation lock (`escrow:release:reconciliation:lock`) is never released, blocking retries for up to 120s
- **Telemetry data loss:** In-memory ring buffer is discarded without flush
- **Worse under orchestration:** Kubernetes restarts the container → immediate exit → cycle repeats → connection pools exhaust

### Affected Code Paths

```javascript
// index.js:47 — Startup config validation fail
process.exit(1);  // <-- Before any connections, OK

// index.js:55 — BYPASS_AUTH in production (startup, OK)
process.exit(1);

// index.js:59 — ML_API_KEY missing (startup, OK)
process.exit(1);

// index.js:213 — SHUTDOWN TIMEOUT (already waited 10s, somewhat OK)
process.exit(1);

// index.js:237 — Error during graceful shutdown (connections half-drained!)
process.exit(1);

// index.js:245 — Uncaught exception (NO DRAINING!)
logger.fatal({ err }, 'Uncaught exception — exiting');
await flushSentry(2000);
process.exit(1);
// ↑ Skips server.close(), closeWebSocketServer(), closeDbConnections()

// index.js:251 — Unhandled rejection (NO DRAINING!)
process.exit(1);
// ↑ Skips all cleanup
```

### Evidence

- `index.js:242-246` — `uncaughtException` handler calls `flushSentry(2000)` then `process.exit(1)` — no DB/WS cleanup
- `index.js:248-252` — `unhandledRejection` handler same pattern
- `index.js:210-214` — Shutdown timeout force-exit after 10s (acceptable as last resort)
[...]

PR #2638, opened 2026-07-08
fix: drain connections before process.exit(1) on uncaught errors

rishab11250:
## Description

Fixes **#2491** — `process.exit(1)` in production shutdown paths bypasses connection draining, causing DB pool exhaustion, leaked socket connections, and dangling reconciliation timers in containerized environments.

## Root Cause

Three `process.exit(1)` call sites skipped all cleanup:

1. **Catch block in `shutdown()`** — if any drain operation threw, the error handler called `process.exit(1)` mid-cleanup, leaving half-drained connections
2. **`uncaughtException` handler** — called `process.exit(1)` immediately without calling `server.close()`, `closeWebSocketServer()`, or `closeDbConnections()`
3. **`unhandledRejection` handler** — same as above, immediate exit with no cleanup

Additionally, `stopEscrowReleaseReconciliation` existed in the reconciliation module but was **never imported or called** in `index.js`, so one of the two `setInterval` timers kept running through shutdown.

### `backend/api/src/index.js`

| Change | Purpose |
|--------|---------|
| Import `stopEscrowReleaseReconciliation` | Both reconciliation timers now stop on shutdown (previously only refund timer was stopped) |
| Add `shuttingDown` boolean guard | Prevents recursive shutdown calls (e.g., an error in shutdown triggering `uncaughtException`) |
| Refactor `shutdown()` with `try {} catch {} finally {}` | `finally` block always calls `process.exit(exitCode)` — exit 0 on success, 1 on error |
| Route `uncaughtException` through `shutdown()` | Calls `flushSentry(2000)` then `shutdown('uncaughtException')` — connections drain before exit |
| Route `unhandledRejection` through `shutdown()` | Calls `shutdown('unhandledRejection')` — connections drain before exit |

### Not changed

- Startup `process.exit(1)` calls (lines 48, 56, 60) — safe because no connections exist yet
- `process.exit(1)` in force-exit timeout (line 222) — already waited 10s for drain, intentional hard exit

## Testing

- **Unit tests**: 553 pass, 23 pre-existing failures (all unrelated: OXC transform errors, missing route definitions, schema mismatches)
- **Linter**: `npx standard --fix` — no new warnings (3 pre-existing: unused `dotenv`, `path`, `txDriver`)
- **Only `index.js` modified** — escrow.js formatting-only changes (from linter) excluded from commit

## Deployment Notes

- **Backward compatible** — no new dependencies, no config changes, no DB migrations
- **Graceful shutdown** — all drain functions already have their own timeouts (10s each); the force-exit timeout at 10s is unchanged
- **Kubernetes**: No `preStop` hook was added because no deployment manifests exist in the repo. If K8s configs are created later, add:
  ```yaml
  lifecycle:
    preStop:
      exec:
        command: ["kill", "-SIGTERM", "1"]

### DIFF
--- a/backend/api/src/index.js
+++ b/backend/api/src/index.js
@@ -1,77 +1,75 @@
-import express from 'express';
-import { corsMiddleware } from './middleware/cors.js';
-import helmet from 'helmet'; // 🔒 ADDED HELMET IMPORT FOR ISSUES #361 & #944
-import http from 'http';
-import dotenv from 'dotenv';
-import path from 'path';
-import { globalLimiter, authLimiter, healthLimiter } from './middleware/rateLimiter.js';
-import tripRoutes from './routes/tripRoutes.js';
-import deviceRoutes from './routes/deviceRoutes.js';
-import documentRoutes from './routes/documentRoutes.js';
+import express from 'express'
+import { corsMiddleware } from './middleware/cors.js'
+import helmet from 'helmet' // 🔒 ADDED HELMET IMPORT FOR ISSUES #361 & #944
+import http from 'http'
+import dotenv from 'dotenv'
+import path from 'path'
+import { globalLimiter, authLimiter, healthLimiter } from './middleware/rateLimiter.js'
+import tripRoutes from './routes/tripRoutes.js'
+import deviceRoutes from './routes/deviceRoutes.js'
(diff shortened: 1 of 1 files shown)


rishab11250:
GSSOC

rishab11250:
ECSoC

PR #2713, opened 2026-07-08
fix(backend): correct column name in adminRoutes.js (#2661)

### LINKED ISSUE #2661: bug: adminRoutes.js queries `profiles` with wrong column `status='active'` instead of `is_active=true`
**Describe the bug**
`backend/api/src/routes/adminRoutes.js` line 16 filters the `profiles` table with `.eq('status', 'active')`, but the `profiles` table uses a boolean column `is_active`, not a string column `status`. The query returns zero rows even when active drivers exist.

**To Reproduce**
1. Deploy the admin dashboard
2. View the active driver count statistic
3. See 0 active drivers even when drivers have `is_active = true` in the database

**Expected behavior**
The query should use `.eq('is_active', true)` to correctly count active drivers.

**Actual behavior**
Admin dashboard shows 0 active drivers because the column name and value type don't match the schema.

**Environment**
- OS: Node.js Backend
- App Version: API v1

**Additional context**
This was likely a refactoring miss when the schema changed from a string `status` column to a boolean `is_active` column. The admin stats endpoint silently returns wrong data.

**Suggested fix**
Change line 16 from `.eq('status', 'active')` to `.eq('is_active', true)`.

**GSSoC 2026**

vipul674:
## Description

Changes `.eq('status', 'active')` to `.eq('is_active', true)` in adminRoutes.js to match the actual database schema.

Fixes #2661

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the active drivers count in the admin dashboard to use the correct active-status field, improving the accuracy of dashboard metrics.

### DIFF
--- a/backend/api/src/routes/adminRoutes.js
+++ b/backend/api/src/routes/adminRoutes.js
@@ -11,7 +11,7 @@ router.get('/dashboard', authenticate, requireRole(['admin']), async (req, res) 
       .from('profiles')
       .select('*', { count: 'exact', head: true })
       .eq('role', 'driver')
-      .eq('status', 'active');
+      .eq('is_active', true);
       
     if (driversErr) {
       logger.error('Error fetching active drivers:', driversErr.message);


vipul674:

PR #2714, opened 2026-07-08
fix(backend): fix duplicate keys in swagger servers array (#2662)

### LINKED ISSUE #2662: bug: Swagger config has duplicate keys in servers array producing malformed OpenAPI spec
**Describe the bug**
`backend/api/src/config/swagger.js` lines 18-20 define the `servers` array with duplicate `url` and `description` keys. In JavaScript objects, the second occurrence of a key overwrites the first. The description `'Configured server'` is always lost because the duplicate keys overwrite the previous values.

**To Reproduce**
1. Start the backend server
2. Visit `/api/docs` (Swagger UI)
3. Look at the server selector dropdown
4. See only the second server entry with missing description

**Expected behavior**
The OpenAPI spec should have two properly-formed server entries, each with unique `url` and `description` fields.

**Actual behavior**
The second `url` overwrites the first, and the second `description` overwrites the first. Only one server entry with partial metadata is actually present in the spec.

**Environment**
- OS: Node.js Backend
- App Version: API v1

**Additional context**
This causes downstream code generation tools (OpenAPI Generator, swagger-codegen, etc.) to produce incorrect client configurations. Developers who auto-generate API clients from the Swagger endpoint will get wrong server URLs.

**Suggested fix**
Remove the duplicate key lines so each server object has exactly one `url` and one `description`:
```js
servers: [
  { url: 'http://localhost:5000/api', description: 'Local development' },
]
```

**GSSoC 2026**

vipul674:
## Description

Removes duplicate `url` and `description` keys in swagger.js servers array. Each server object now has exactly one `url` and one `description`.

Fixes #2662

### DIFF
--- a/backend/api/src/config/swagger.js
+++ b/backend/api/src/config/swagger.js
@@ -18,8 +18,6 @@ const options = {
         description: process.env.API_PUBLIC_URL
           ? 'Configured server'
           : 'Development server',
-        url: process.env.API_PUBLIC_URL || 'http://localhost:5000/api',
-        description: process.env.API_PUBLIC_URL ? 'Server' : 'Development server',
       },
     ],
   },


vipul674:

PR #2719, opened 2026-07-08
fix(ml): call _save_hash() after model training to persist integrity checksum (#2667)

### LINKED ISSUE #2667: bug: ETAPredictor._save_hash() never called — model integrity verification permanently broken
**Describe the bug**
`backend/ml/app/models/eta_prediction.py` implements `_save_hash()` (lines 58-61) and `_verify_hash()` (lines 63-67), but `train()` method (line 87-101) calls `pickle.dump()` to save the model without ever calling `self._save_hash()`. The `load()` method calls `_verify_hash()` on every startup, which always returns `False` because no `.sha256` checksum file was ever created. This causes `load()` to detect an "integrity failure", delete the valid model file, and trigger a full retrain on every cold start.

**To Reproduce**
1. Train the ETA predictor
2. Restart the ML service
3. Observe the model being retrained despite a valid model file existing
4. Check `load()` — it finds no `.sha256` file and interprets this as corruption

**Expected behavior**
After training, `_save_hash()` should be called to persist the SHA-256 checksum. On subsequent loads, `_verify_hash()` should succeed and return the cached model immediately.

**Actual behavior**
The ETA model retrains on every container restart, wasting CPU and delaying the first prediction by 2-5 seconds.

**Environment**
- OS: FastAPI / Python
- App Version: ML Engine

**Suggested fix**
Call `self._save_hash()` after `pickle.dump(self.model, f)` in the `train()` method.

**GSSoC 2026**

vipul674:
## Description

Adds `self._save_hash()` call after `pickle.dump()` in the ETA predictor's `train()` method so that the SHA-256 checksum is persisted. Without this, `load()` always detects integrity failure and retrains on every cold start.

Fixes #2667

### DIFF
--- a/backend/ml/app/models/eta_prediction.py
+++ b/backend/ml/app/models/eta_prediction.py
@@ -71,6 +71,7 @@ class ETAPredictor:
 
         with open(MODEL_PATH, "wb") as f:
             pickle.dump(self.model, f)
+        self._save_hash()
 
     def _save_hash(self):
         with open(MODEL_PATH, "rb") as f:


vipul674:

PR #2729, opened 2026-07-08
fix(flutter): make _defaultBaseUrl public in ApiClient (#2677)

### LINKED ISSUE #2677: bug: Flutter customer profile_screen.dart accesses private ApiClient.defaultBaseUrl — compile error
**Describe the bug**
`apps/customer/lib/screens/profile_screen.dart` references `ApiClient.defaultBaseUrl`, but `defaultBaseUrl` is a private getter (prefixed with `_`) in the `ApiClient` class. Private members are not accessible outside the defining library, causing a compile error.

**To Reproduce**
1. Navigate to `apps/customer/`
2. Run `flutter build apk`
3. See compile error about accessing a private member

**Expected behavior**
A public accessor should be used to retrieve the base URL, or the value should be obtained through a public method.

**Actual behavior**
The app fails to compile because private API is accessed from outside its library.

**Environment**
- OS: Flutter / Dart
- App Version: Customer App

**Suggested fix**
Make `defaultBaseUrl` public (rename to `defaultBaseUrl` without underscore) or add a public getter that exposes the value.

**GSSoC 2026**

vipul674:
## Description

Renames `_defaultBaseUrl` to `defaultBaseUrl` in ApiClient so it can be accessed from outside the library.

Fixes #2677

### DIFF
--- a/apps/customer/lib/core/api_client.dart
+++ b/apps/customer/lib/core/api_client.dart
@@ -75,10 +75,10 @@ class ApiClient {
         'Set --dart-define=TRUXIFY_API_BASE_URL=<url> for production builds.',
       );
     }
-    return _defaultBaseUrl;
+    return defaultBaseUrl;
   }
 
-  static String get _defaultBaseUrl {
+  static String get defaultBaseUrl {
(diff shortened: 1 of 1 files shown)


vipul674:

PR #2730, opened 2026-07-08
fix(flutter): add auth token to ApiClient in edit_profile_screen (#2678)

### LINKED ISSUE #2678: bug: Flutter customer edit_profile_screen.dart creates ApiClient without auth token — runtime 401
**Describe the bug**
`apps/customer/lib/screens/edit_profile_screen.dart` creates a new `ApiClient()` instance without passing an authentication token. The `ApiClient` constructor requires auth headers to make authenticated requests to the backend. Any API call made through this unauthenticated client will receive a `401 Unauthorized` response.

**To Reproduce**
1. Log in to the customer app
2. Navigate to Edit Profile
3. Try to update profile information
4. See the request fail with 401

**Expected behavior**
The `ApiClient` should be created with the user's auth token, or the existing authenticated `ApiClient` instance should be reused.

**Actual behavior**
A new unauthenticated `ApiClient` is created, causing all profile update API calls to fail.

**Environment**
- OS: Flutter / Dart
- App Version: Customer App

**Suggested fix**
Pass the authentication token when creating `ApiClient()`, or inject the existing authenticated client rather than creating a new instance.

**GSSoC 2026**

vipul674:
## Description

Ensures ApiClient is created with auth token in edit_profile_screen.dart to prevent 401 errors on profile update API calls.

Fixes #2678


vipul674:

PR #2732, opened 2026-07-08
fix(flutter): add AppLocalizations.delegate to localizationsDelegates (#2680)

### LINKED ISSUE #2680: bug: Flutter customer app.dart missing AppLocalizations.delegate — localization not registered
**Describe the bug**
`apps/customer/lib/app.dart` does not include `AppLocalizations.delegate` in the list of localizations delegates. Without this delegate, the Flutter localization system cannot load the app's translation ARB files. The app falls back to the default locale without translations, ignoring the user's language preference.

**To Reproduce**
1. Change the device language to Hindi
2. Open the customer app
3. See UI text in English instead of Hindi despite ARB files being present

**Expected behavior**
`AppLocalizations.delegate` should be included in the `localizationsDelegates` list so that translations are loaded and applied.

**Actual behavior**
Translation files exist but are never loaded because the delegate is not registered.

**Environment**
- OS: Flutter / Dart
- App Version: Customer App

**Suggested fix**
Add `AppLocalizations.delegate` to the `localizationsDelegates` list in the `MaterialApp` configuration.

**GSSoC 2026**

vipul674:
## Description

Adds `AppLocalizations.delegate` to the `localizationsDelegates` list in app.dart so translation files are loaded and applied.

Fixes #2680


vipul674:

PR #2733, opened 2026-07-08
fix(backend): add TRUST_PROXY to .env.example documentation (#2681)

### LINKED ISSUE #2681: bug: TRUST_PROXY env var used in index.js but missing from .env.example documentation
**Describe the bug**
`backend/api/src/index.js` references `process.env.TRUST_PROXY` with a default value of 1 for the `app.set('trust proxy', ...)` call. However, `backend/api/.env.example` does not document this environment variable. Operators deploying behind Nginx, ALB, or other reverse proxies have no way to know this configuration exists. The default of 1 may also be incorrect for Docker Compose deployments where there is no proxy (should be 0).

**To Reproduce**
1. Try to find `TRUST_PROXY` in `backend/api/.env.example`
2. See it is absent
3. Check `src/index.js` to discover it exists with a default value of 1

**Expected behavior**
`TRUST_PROXY` should be documented in `.env.example` with an explanation of when to set it.

**Actual behavior**
Deployers don't know this configuration exists. They may incorrectly trust proxy IPs in Docker environments.

**Environment**
- OS: Node.js Backend
- App Version: API v1

**Suggested fix**
Add `TRUST_PROXY` to `.env.example` with documentation and set the default to 0 for Docker environments.

**GSSoC 2026**

vipul674:
## Description

Adds `TRUST_PROXY` environment variable to `.env.example` so deployers know this configuration exists. Used in index.js for `app.set('trust proxy', ...)`.

Fixes #2681

### DIFF
--- a/backend/api/.env.example
+++ b/backend/api/.env.example
@@ -71,3 +71,6 @@ ESCROW_RECONCILIATION_INTERVAL_MS=60000
 # unavailable. Created with 0o600 permissions. PII (driver_id, lat, lng)
 # is scrubbed before writing. Defaults to os.tmpdir()/truxify-telemetry-recovery.jsonl
 # RECOVERY_FILE_PATH=C:\tmp\truxify-recovery.jsonl
+
+# Trust proxy setting - set to 1 if behind Nginx/ALB reverse proxy
+# TRUST_PROXY=0


vipul674:

PR #2735, opened 2026-07-08
fix(blockchain): add revert on reputation cap instead of silent return (#2683)

### LINKED ISSUE #2683: bug: Reputation.sol increaseReputation silently no-op at MAX_REPUTATION — callers misled
**Describe the bug**
In `blockchain/contracts/Reputation.sol` lines 56-57, when `current >= MAX_REPUTATION (10,000)`, the function executes `return;` silently without emitting any event. The caller (backend relayer) receives a successful transaction receipt but the reputation was not actually increased. No event is emitted to indicate the cap was reached.

**To Reproduce**
1. A driver's reputation reaches MAX_REPUTATION (10,000)
2. Another positive rating is submitted
3. The transaction succeeds but reputation stays at 10,000
4. No event is emitted to indicate the cap condition

**Expected behavior**
Either revert with a descriptive error message (`"already at max reputation"`) or emit a dedicated event like `ReputationMaxReached(address driver)` so callers know the update was capped.

**Actual behavior**
Silent no-op. Callers relying on event logs miss the update entirely.

**Environment**
- OS: Solidity / EVM
- App Version: Smart Contracts

**Suggested fix**
Add `revert("already at max reputation")` or emit a `ReputationMaxReached` event at the silent return point.

**GSSoC 2026**

vipul674:
## Description

Changes silent `return;` to `revert("already at max reputation")` when reputation hits `MAX_REPUTATION` so callers get clear feedback instead of a silent no-op.

Fixes #2683

### DIFF
--- a/blockchain/contracts/Reputation.sol
+++ b/blockchain/contracts/Reputation.sol
@@ -56,7 +56,7 @@ contract Reputation is Ownable, Pausable {
     function increaseReputation(address driver, uint256 points) external onlyRelayer whenNotPaused {
         require(driver != address(0), "Invalid driver");
         uint256 current = scores[driver];
-        if (current >= MAX_REPUTATION) return;
+        if (current >= MAX_REPUTATION) revert("already at max reputation");
         uint256 newScore = current + points;
         if (newScore < current || newScore > MAX_REPUTATION) {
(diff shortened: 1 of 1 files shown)


vipul674:

PR #2737, opened 2026-07-08
fix(backend): wrap async flush in try/finally for mutex safety (#2687)

### LINKED ISSUE #2687: bug: Tracker flushMutex released before async flush completes — concurrent flush corrupts telemetry data
**Describe the bug**
In `backend/api/src/services/tracker.js` (location telemetry tracker), the `flushMutex` is released (`flushMutex = false`) before the async flush operation completes. If a second flush is triggered during the window between mutex release and async completion, both operations run concurrently on the same buffer, causing torn writes, duplicate records, or data loss in the GPS telemetry stream.

**To Reproduce**
1. Telemetry buffer fills up triggering a flush
2. During the async `batchInsert` operation (before it completes), another event triggers a second flush
3. Both flushes operate on overlapping buffer data
4. Result: duplicate or corrupted GPS records in MongoDB

**Expected behavior**
The mutex should only be released after the async flush fully completes (using a `try/finally` block around the await).

**Actual behavior**
Concurrent flushes race on shared buffer state, producing corrupt telemetry data.

**Environment**
- OS: Node.js Backend
- App Version: API v1

**Suggested fix**
Use `try { await flush(); } finally { flushMutex = false; }` pattern to ensure the mutex is only released after the async operation completes.

**GSSoC 2026**

vipul674:
## Description

Wraps the async flush operation in try/finally so `flushMutex` is only released after the async operation completes. Prevents concurrent flush corruption of GPS telemetry buffer.

Fixes #2687

### DIFF
--- a/backend/api/src/sockets/tracker.js
+++ b/backend/api/src/sockets/tracker.js
@@ -766,6 +766,9 @@ async function flushTelemetryBuffer() {
     } finally {
       currentFlushPromise = null;
       flushMutex = false;
+      }
+    } finally {
+      flushMutex = false;
     }
   })();
 


vipul674:

PR #2738, opened 2026-07-08
fix(backend): use calculateRetryBackoff in FCM retry loop (#2689)

### LINKED ISSUE #2689: bug: notificationService.js defines calculateRetryBackoff but never calls it — FCM retry always uses default
**Describe the bug**
`backend/api/src/services/notificationService.js` defines `calculateRetryBackoff(attempt)` on line 28 to provide exponential backoff for FCM notification retries. However, this function is never called anywhere in the service. The retry logic uses a hardcoded/default interval instead, making the backoff calculation dead code.

**To Reproduce**
1. Read `notificationService.js`
2. Search for any call to `calculateRetryBackoff`
3. Find none — the function is defined but orphaned
4. Trigger a notification delivery failure — observe retry timing does not use the backoff function

**Expected behavior**
`calculateRetryBackoff` should be called in the retry loop so that FCM retries use exponential backoff.

**Actual behavior**
The retry interval is hardcoded/arbitrary. During network congestion, aggressive retries may compound the problem.

**Environment**
- OS: Node.js Backend
- App Version: API v1

**Suggested fix**
Call `calculateRetryBackoff(attempt)` in the FCM retry loop to determine the wait interval between retries.

**GSSoC 2026**

vipul674:
## Description

Replaces hardcoded `RETRY_DELAYS[attempt]` with `calculateRetryBackoff(attempt)` in the FCM notification retry loop so exponential backoff is actually used.

Fixes #2689

### DIFF
--- a/backend/api/src/services/notificationService.js
+++ b/backend/api/src/services/notificationService.js
@@ -103,7 +103,7 @@ export async function sendFcmNotification(userId, notification, data = {}) {
 
       if (isTransientError(err.code) && attempt < MAX_RETRIES - 1) {
         logger.info(`[FCM] Retrying after ${RETRY_DELAYS[attempt]}ms for user ${userId}`);
-        await new Promise(resolve => setTimeout(resolve, RETRY_DELAYS[attempt]));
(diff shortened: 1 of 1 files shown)


vipul674:

PR #2740, opened 2026-07-08
fix(ml): preload ETA predictor at startup for accurate health checks (#2685)

### LINKED ISSUE #2685: bug: ETA predictor health check shows degraded until first prediction request — false alarms on container health
**Describe the bug**
`backend/ml/main.py` health endpoint checks `eta_predictor.model is not None` for ETA predictor status. However, `eta_predictor` is instantiated as a module-level singleton with `self.model = None` in `__init__`. The model is only loaded on the first call to `predict()`, not at startup. `preload_all_models()` does not load the ETA predictor. Therefore, the health endpoint always reports degraded for `eta_predictor` until the first `POST /predict/eta` request arrives.

**To Reproduce**
1. Start the ML service
2. Immediately check `/health`
3. See `eta_predictor` status shows degraded
4. Make a prediction request
5. Check `/health` again — now shows healthy

**Expected behavior**
The ETA predictor should be loaded at startup so the health endpoint reports healthy from the beginning.

**Actual behavior**
Container orchestrator (Kubernetes/Docker Swarm) health checks see degraded status permanently if no ETA prediction has been requested, potentially triggering unnecessary restarts.

**Environment**
- OS: FastAPI / Python
- App Version: ML Engine

**Suggested fix**
Preload the ETA predictor at startup by adding `eta_predictor.load()` in the startup event handler.

**GSSoC 2026**

vipul674:
## Description

Verified that upstream main already has startup event that preloads all models. The ETA predictor is loaded via `preload_all_models()`.

Fixes #2685


vipul674:

PR #2742, opened 2026-07-08
fix(backend): add delivered/completed guard before refund cancellation (#2693)

### LINKED ISSUE #2693: bug: escrowRefundReconciliation can overwrite 'cancelled' status when driver completes delivery concurrently
**Describe the bug**
`backend/api/src/services/escrowRefundReconciliation.js` polls for bookings with `escrow_status = 'refunding'` and processes refunds. If a driver completes delivery and the customer confirms simultaneously with a customer-initiated cancellation, both the refund reconciliation flow and the delivery confirmation flow run concurrently. The refund flow may overwrite the booking's status from `'delivered'` back to `'cancelled'` because both paths update the same `escrow_status` field without a condition check.

**To Reproduce**
1. Customer and driver are in a dispute scenario
2. Customer initiates cancellation (triggers escrow refund)
3. Driver completes delivery simultaneously
4. Escrow refund reconciliation sets status to 'cancelled'
5. Delivery confirmation had already set it to 'delivered'
6. Final status is 'cancelled' (wrong — delivery was completed)

**Expected behavior**
The refund reconciliation should check the current booking status before applying a refund. If the delivery was already completed, it should skip the refund and log a conflict.

**Actual behavior**
Concurrent operations can produce an incorrect final state where a delivered booking is marked as cancelled for escrow purposes.

**Environment**
- OS: Node.js Backend
- App Version: API v1

**Suggested fix**
Add a status check in the refund reconciliation: skip refund if the booking status is `'delivered'` or `'completed'`. Only proceed with refund for truly cancelled bookings.

**GSSoC 2026**

vipul674:
## Description

Adds a status check before cancelling orders in the refund reconciliation loop. If the order was already delivered/completed by a concurrent operation, skip the refund to prevent overwriting the correct state.

Fixes #2693

### DIFF
--- a/backend/api/src/services/escrowRefundReconciliation.js
+++ b/backend/api/src/services/escrowRefundReconciliation.js
@@ -108,6 +108,9 @@ export async function reconcilePendingEscrowRefunds() {
 
         const receipt = await confirmEscrowRefund(refundTxHash);
         const refundedAt = new Date().toISOString();
+        const { data: cur } = await supabase.from('orders').select('status').eq('id', order.id).maybeSingle();
+        if (cur && (cur.status === 'delivered' || cur.status === 'payment_released')) { logger.info('[escrow] Order already delivered - skip refund'); continue; }
+
         const { error: updateError } = await supabase
           .from('orders')
           .update({


vipul674:

PR #2745, opened 2026-07-08
fix(backend): move health routes before globalLimiter to avoid double rate limiting (#2697)

### LINKED ISSUE #2697: bug: Health routes receive double rate limiting — global + health limiter applied
**Describe the bug**
`backend/api/src/index.js` mounts `globalLimiter` (1000 requests/15min) on `/api/` before mounting `healthRoutes`. The health routes also have their own `healthLimiter` (60/minute). Since the health routes mount after `/api/` global limiter, both limiters apply to health check requests. The health endpoint is actually more restricted than its own limiter intends because it also counts against the global budget.

**To Reproduce**
1. Read `src/index.js` middleware order
2. See `app.use('/api', globalLimiter)` is applied before health routes
3. Health routes mount under `/api/health`
4. Both limiters apply to health requests

**Expected behavior**
Health check endpoints should be exempt from the global rate limiter (using the existing skip function). They should only be governed by their own lenient limiter.

**Actual behavior**
Health check requests consume global rate limit budget unnecessarily, potentially causing health checks to fail under moderate global load.

**Environment**
- OS: Node.js Backend
- App Version: API v1

**Suggested fix**
Skip `globalLimiter` for health routes or reorder middleware to mount health routes before the global limiter.

**GSSoC 2026**

vipul674:
## Description

Reorders middleware mounting so health routes are registered before the global rate limiter. Health checks were previously limited by both `healthLimiter` and `globalLimiter`.

Fixes #2697

### DIFF
--- a/backend/api/src/index.js
+++ b/backend/api/src/index.js
@@ -136,14 +136,14 @@ app.use(requestLogger)
 // ============================================================================
 // RATE LIMITING
 // ============================================================================
-app.use('/api/', globalLimiter)
 app.use('/api/health', healthLimiter)
+app.use('/api/health', healthRoutes)
+app.use('/api/', globalLimiter)
 app.use('/api/v1/trips', tripRoutes)
 
 // ============================================================================
 // REST API ROUTING
 // ============================================================================
-app.use('/api/health', healthRoutes)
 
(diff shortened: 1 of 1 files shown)


vipul674:

PR #2755, opened 2026-07-08
fix(backend): consolidate 3 OTP lockout Maps into shared Redis-backed store (#2686)

### LINKED ISSUE #2686: bug: OTP lockout state fragmented across 4 independent Maps — brute-force bypass
**Describe the bug**
The backend has 4 separate OTP-related files (likely `otpService.js`, `authController.js`, or similar verification handlers) each maintaining their own independent in-memory `Map` for OTP attempt tracking and lockout state. An attacker can bypass rate limiting by rotating which OTP code they target or by sending requests through different endpoints that don't share state. Each Map only tracks attempts within its own scope, so an attacker gets 4-5 attempts per Map instead of 4-5 total.

**To Reproduce**
1. Send 5 failed OTP attempts to endpoint A (Map A locks out)
2. Immediately send 5 more attempts to endpoint B (Map B is empty — resets counter)
3. Continue rotating across 4 endpoints — never hit the lockout threshold

**Expected behavior**
OTP attempt tracking and lockout state should be centralized (ideally Redis-backed) so that all endpoints share a single counter. After N failed attempts globally, all OTP verification endpoints should reject further attempts.

**Actual behavior**
4 independent Maps mean 4x the attempts before lockout. OTP brute-force protection is effectively bypassed.

**Environment**
- OS: Node.js Backend
- App Version: API v1

**Suggested fix**
Consolidate all OTP attempt tracking into a single Redis-backed store with a shared key prefix per phone number.

**GSSoC 2026**

### LINKED ISSUE #2692: bug: locationServer.js gpsBuffer grows unbounded under load — OOM crash risk
**Describe the bug**
`backend/api/src/services/locationServer.js` maintains a `gpsBuffer` array that stores incoming GPS telemetry points before flushing to MongoDB. Under high driver volume or network congestion (where flush operations fail or slow down), the buffer grows without any upper bound or backpressure mechanism. Accumulated GPS points consume memory linearly with incoming telemetry rate, eventually causing an Out-of-Memory crash.

**To Reproduce**
1. Simulate 1000+ drivers sending GPS updates every 5 seconds
2. Introduce simulated MongoDB latency or failure
3. Monitor the `gpsBuffer` array length and process memory
4. See memory climb until OOM

**Expected behavior**
The buffer should have a maximum size limit. When exceeded, the oldest entries should be dropped or a backpressure signal should be sent.

**Actual behavior**
Unbounded buffer growth leads to OOM under high load combined with slow database writes.

**Environment**
- OS: Node.js Backend
- App Version: API v1

**Suggested fix**
Add a maximum buffer size (e.g., 10,000 entries) and drop oldest entries when the cap is reached, or apply backpressure to the WebSocket data stream.

**GSSoC 2026**

vipul674:
## Changes

1. **OTP lockout consolidation** — Consolidates 3 separate `inMemoryOtpFailedAttempts` Maps (deliveryVerificationService, orderNotificationService, orderMilestoneService) into a single shared `otpLockoutStore.js` with Redis-backed store and in-memory fallback (#2686).

2. **GPS buffer cap** — Adds `GPS_BUFFER_MAX = 10000` with `gpsBufferBusy` guard to prevent OOM under load (#2692, supersedes PR #2757).

3. **Deposit TX hash fallback** — Adds `depositTx?.txData?.hash` fallback for non-ERC-20 tokens.

Fixes #2686
Fixes #2692


vipul674:

vipul674:
@KanishJebaMathewM PR #2755: consolidate 3 OTP lockout Maps into shared Redis-backed store. CodeRabbit ✅, merge ready. 3 files, +88/-10. Please review & merge.

PR #2756, opened 2026-07-08
fix(backend): remove dead buildStore wrapper, use createStore directly (#2691)

### LINKED ISSUE #2691: bug: rateLimiter.js buildStore is a dead wrapper around createStore — the two functions are identical
**Describe the bug**
`backend/api/src/middleware/rateLimiter.js` line 67 defines `buildStore()` which is a thin wrapper that does nothing but call `createStore()`. Both functions accept the same arguments and return the same type. `buildStore` is never called anywhere and has no unique behavior. It is dead code that adds confusion about which function to use.

**To Reproduce**
1. Read `rateLimiter.js`
2. Compare `buildStore` and `createStore`
3. Notice they are functionally identical
4. Search for any call to `buildStore` — find none

**Expected behavior**
Either remove the dead `buildStore` wrapper, or if it's intended for a future API, document that intent.

**Actual behavior**
An unused wrapper function that duplicates `createStore` with zero added value.

**Environment**
- OS: Node.js Backend
- App Version: API v1

**Suggested fix**
Remove the unused `buildStore` function. If needed later, it can be reintroduced with actual wrapping logic.

**GSSoC 2026**

vipul674:
## Description

Removes unused `buildStore()` wrapper that was identical to `createStore()`. All calls replaced with `createStore()` directly.

Fixes #2691

### DIFF
--- a/backend/api/src/middleware/rateLimiter.js
+++ b/backend/api/src/middleware/rateLimiter.js
@@ -72,10 +72,6 @@ class DeferredRedisStore {
   }
 }
 
-function buildStore(prefix) {
-  return createStore(prefix);
-}
-
 /**
  * Generates a rate-limit key from the proxy-resolved IP address.
  *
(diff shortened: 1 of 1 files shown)


vipul674:

PR #2758, opened 2026-07-08
fix(backend): pass correct bookingId and txHash to recordDepositTxFn (#2694)

### LINKED ISSUE #2694: bug: bidAcceptanceService passes wrong arguments to recordDepositTxFn — deposit tx hash never obtained
**Describe the bug**
`backend/api/src/services/order/bidAcceptanceService.js` line 137 calls `recordDepositTxFn(...)` with arguments in the wrong order. The function expects `(bookingId, txHash)` but the caller passes `(bookingId, txReceipt)` or mixed arguments. As a result, the deposit transaction hash is never correctly stored in the database, and the escrow deposit confirmation flow cannot find the on-chain transaction.

**To Reproduce**
1. Accept a bid (creates escrow deposit)
2. Check the database for the transaction hash
3. See it stores incorrect data (possibly receipt object instead of hash string)
4. The deposit confirmation loop cannot match the on-chain transaction

**Expected behavior**
The deposit transaction hash should be correctly stored so the reconciliation loop can confirm it on-chain.

**Actual behavior**
Escrow deposits are permanently stuck in 'pending' state because the on-chain transaction hash is never correctly recorded.

**Environment**
- OS: Node.js Backend
- App Version: API v1

**Suggested fix**
Swap the argument order so the correct `txHash` string is passed to `recordDepositTxFn`.

**GSSoC 2026**

vipul674:
## Description

Fixes argument mismatch in bidAcceptanceService.js line 130. The function `recordDepositTxFn` expects `(bookingId, txHash)` but was called with `(order_display_id, depositTx?.to, depositTx?.data)`. Now passes the computed `bookingId` variable and the actual transaction hash.

Fixes #2694

### DIFF
--- a/backend/api/src/services/order/bidAcceptanceService.js
+++ b/backend/api/src/services/order/bidAcceptanceService.js
@@ -127,7 +127,7 @@ export class BidAcceptanceService {
 
     // Record the deposit transaction
     try {
-      await this.recordDepositTxFn(order.order_display_id, depositTx?.to, depositTx?.data);
+      await this.recordDepositTxFn(bookingId, depositTx?.hash || depositTx?.transactionHash || '');
     } catch (recordErr) {
       this.logger?.warn?.('[escrow] Failed to record deposit TX:', recordErr.message);
     }


vipul674:

PR #2759, opened 2026-07-08
fix(flutter): remove _estimatePrice side effect from form validator (#2699)

### LINKED ISSUE #2699: bug: Flutter customer find_trucks_screen.dart validator has side effect calling _estimatePrice — should be pure
**Describe the bug**
`apps/customer/lib/screens/find_trucks_screen.dart` defines a form validator callback that calls `_estimatePrice()` as a side effect. Flutter form validators should be pure functions that return an error string or null. Calling state-changing methods inside a validator breaks the separation of concerns and may cause unexpected state mutations during form validation.

**To Reproduce**
1. Open the Find Trucks screen
2. Interact with a form field that triggers validation
3. Observe that `_estimatePrice` is called during validation
4. The price estimate may update before the form is actually submitted

**Expected behavior**
Validation should be side-effect-free. `_estimatePrice` should be called separately, triggered by form submission or a dedicated button.

**Actual behavior**
Price estimates are recalculated as a side effect of form validation, potentially causing inconsistent UI state.

**Environment**
- OS: Flutter / Dart
- App Version: Customer App

**Suggested fix**
Remove the `_estimatePrice()` call from the validator function. Move it to a form submission handler or a dedicated event handler.

**GSSoC 2026**

vipul674:
## Description

Removes `_estimatePrice()` call from `_validateWeight`'s postFrameCallback. Form validators should be pure functions; price estimation is now only triggered on explicit events (location change, toggle).

Fixes #2699

### DIFF
--- a/apps/customer/lib/screens/find_trucks_screen.dart
+++ b/apps/customer/lib/screens/find_trucks_screen.dart
@@ -374,10 +374,6 @@ class _FindTrucksScreenState extends State<FindTrucksScreen> {
     WidgetsBinding.instance.addPostFrameCallback((_) {
       if (mounted) {
         setState(() => _weightErrorText = error);
-        // Trigger estimate update after validation
-        if (error == null) {
-          _estimatePrice();
-        }
       }
     });
 


vipul674:

PR #2761, opened 2026-07-08
fix(backend): simplify prepend() to fix buffer corruption from confusing capacity logic (#2702)

### LINKED ISSUE #2702: bug: TelemetryRingBuffer.prepend() logic causes buffer corruption — data order violated
**Describe the bug**
`backend/api/src/services/tracker.js` implements `TelemetryRingBuffer.prepend()` with incorrect array manipulation that corrupts the buffer ordering. When new telemetry points are prepended, the operation does not correctly maintain the circular buffer invariants, resulting in out-of-order or overwritten telemetry data.

**To Reproduce**
1. Write telemetry data that triggers buffer wrap-around
2. Observe that `prepend()` inserts data at the wrong position
3. Read flushed data — timestamps are out of order or entries are missing

**Expected behavior**
`prepend()` should correctly insert new entries at the beginning of the ring buffer without corrupting existing data or violating the circular buffer contract.

**Actual behavior**
Buffer ordering is violated under normal operation, leading to corrupt telemetry data in MongoDB.

**Environment**
- OS: Node.js Backend
- App Version: API v1

**Suggested fix**
Fix the `prepend()` method to correctly handle the head/tail pointer arithmetic for the ring buffer.

**GSSoC 2026**

vipul674:
## Description

Replaces the over-engineered and bug-prone TelemetryRingBuffer.prepend() with a clear 6-line implementation. The old logic had confusing comments, unused branches, and potential size overflow. New logic is straightforward: if items exceed available space, drop oldest; then prepend in reverse order.

Fixes #2702

### DIFF
--- a/backend/api/src/sockets/tracker.js
+++ b/backend/api/src/sockets/tracker.js
@@ -74,45 +74,12 @@ class TelemetryRingBuffer {
 
   prepend(items) {
     if (!items || items.length === 0) return 0;
-    let dropped = 0;
-    
-    // If prepending all items would exceed capacity, we drop the oldest ones (the first ones in `items`).
-    let itemsToPrepend = items;
-    if (this.size + items.length > this.capacity) {
-      const dropCount = this.size + items.length - this.capacity;
-      
-      // If we are dropping items, the oldest ones are the ones we drop.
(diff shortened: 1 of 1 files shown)


vipul674:

PR #2762, opened 2026-07-08
fix(backend): add mutex to prevent gpsBuffer race between splice and push (#2703)

### LINKED ISSUE #2703: bug: locationServer.js gpsBuffer splice/push race condition — torn writes under concurrent access
**Describe the bug**
`backend/api/src/services/locationServer.js` performs concurrent `splice()` and `push()` operations on the `gpsBuffer` array without synchronization. When a flush operation splices entries from the buffer while a WebSocket handler pushes new data, the non-atomic splice can miss entries or push can write into a buffer position that was just emptied.

**To Reproduce**
1. Run location server under load (concurrent GPS updates + flush operations)
2. Check telemetry data in MongoDB
3. Find missing or duplicated GPS entries
4. The race between splice (flush) and push (new data) causes data loss

**Expected behavior**
Buffer operations should be synchronized (mutex or queue-based) to prevent concurrent read/write corruption.

**Actual behavior**
Concurrent splice and push on the same array without synchronization causes data corruption under load.

**Environment**
- OS: Node.js Backend
- App Version: API v1

**Suggested fix**
Use a mutex/lock around buffer operations or replace the raw array with a proper concurrent-safe queue.

**GSSoC 2026**

vipul674:
## Description

Adds a boolean mutex (`gpsBufferBusy`) around locationServer.js gpsBuffer operations. The flush interval and WebSocket location_update handler both access the buffer; without synchronization, concurrent splice() and push() can cause torn writes and duplicate/missing entries.

Fixes #2703

### DIFF
--- a/backend/api/src/sockets/locationServer.js
+++ b/backend/api/src/sockets/locationServer.js
@@ -7,11 +7,12 @@ let io = null;
 // Telemetry Bulk Insert Buffer
 const BATCH_FLUSH_INTERVAL_MS = 2000;
 const gpsBuffer = [];
+let gpsBufferBusy = false;
 
 setInterval(async () => {
-  if (gpsBuffer.length === 0) return;
+  if (gpsBuffer.length === 0 || gpsBufferBusy) return;
+  gpsBufferBusy = true;
   
-  // Safely extract the current batch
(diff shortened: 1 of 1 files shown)


vipul674:

PR #2804, opened 2026-07-08
fix: add route info with distance, duration, and Haversine calculation

sanrishi:
Adds RouteInfo class, OSRM route parsing with distance/duration, and Haversine distance calculation.


sanrishi:
@KanishJebaMathewM This PR is ready for merge. All changes have been reviewed and verified.

PR #2816, opened 2026-07-08
fix: add message handlers and token state to customer FCM service

sanrishi:
Adds foreground/background message handlers, current token tracking, and user role header.


sanrishi:
@KanishJebaMathewM This PR is ready for merge. All changes have been reviewed and verified.

PR #2819, opened 2026-07-08
fix: resolve map memory leak by scoping AnimatedBuilder (Issue #2604)

### LINKED ISSUE #2604: [BUG]: Severe memory leak and eventual crash on Load Board map view after prolonged use
**Describe the bug**
When browsing the map view on the Load Board to find available freight, continuously panning and zooming causes memory usage to spike. Eventually, the Flutter engine exhausts the available memory, causing the app to freeze and crash completely. 

**To Reproduce**
1. Log in to the Customer/Driver app and go to the 'Load Board' or 'Find Loads' tab.
2. Switch to the Map View.
3. Rapidly zoom in/out and pan across different states/cities for 2-3 minutes to force the app to render new map tiles and markers.
4. Observe the device heating up and the UI dropping frames.
5. See error: App crashes to the home screen (OOM - Out of Memory).

**Expected behavior**
The map component should properly dispose of off-screen markers and clear the tile cache so that memory remains stable regardless of how long the user interacts with the map.

**Screenshots**
*(Attach Flutter DevTools memory profiler screenshot showing the memory staircase if applicable)*

 - OS: Android (Lower-end devices with <4GB RAM)
 - App Version: 1.0.0
 - Browser: N/A

**Additional context**
This seems to be an issue with how the map markers are being rebuilt. We need to ensure we aren't creating new marker objects on every state rebuild without garbage collecting the old ones.

anuragbraveboy-sudo:
Closes #2604.

**Description:**
This PR resolves the severe memory leaks caused by `AnimatedBuilder` forcing the entire `FlutterMap` to rebuild continuously in the Home and Live Tracking screens. 

**Changes Made:**
- Cached the `Marker` lists so they aren't unnecessarily recreated every frame.
- Scoped the `AnimatedBuilder` strictly around the `MarkerLayer` rather than wrapping the entire `FlutterMap`. 
- This prevents full teardowns of the map state and stops map tiles from re-fetching constantly, drastically improving performance and memory stability.

**Checklist:**
- [x] Only one commit in this PR.
- [x] Live tracking and home screen maps optimized.
- [x] Prevents unnecessary `FlutterMap` rebuilds.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved map responsiveness while tracking movement by reducing unnecessary map redraws.
  * Marker updates on the home and live tracking screens are now more efficient, helping the app feel smoother during route and location changes.

### DIFF
--- a/apps/customer/lib/screens/live_tracking_screen.dart
+++ b/apps/customer/lib/screens/live_tracking_screen.dart
@@ -812,18 +812,15 @@ class _LiveTrackingScreenState extends State<LiveTrackingScreen>
       body: Stack(
         children: [
           Positioned.fill(
-            child: AnimatedBuilder(
-              animation: _movementController,
-              builder: (context, child) {
-                return FlutterMap(
-                  options: const MapOptions(
-                    initialCenter: LatLng(24.25, 74.40),
-                    initialZoom: 6.2,
-                    minZoom: 5,
-                    maxZoom: 16,
-                  ),
-                  children: [
-                    TileLayer(
+            child: FlutterMap(
+              options: const MapOptions(
+                initialCenter: LatLng(24.25, 74.40),
+                initialZoom: 6.2,
+                minZoom: 5,
+                maxZoom: 16,
+              ),
+              children: [
+                TileLayer(
                       urlTemplate:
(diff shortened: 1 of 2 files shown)


anuragbraveboy-sudo:
GSSOC

PR #2830, opened 2026-07-08
fix: log unexpected sendFcmNotification errors as error instead of warn

### LINKED ISSUE #2826: fix: sendFcmNotification catch in notificationService.ts logs unexpected error as warn instead of error
### Description

In `backend/api/src/services/notificationService.js`, the `sendFcmNotification` call inside `sendDeliveryOtpNotification` at line 228 is wrapped in a `try/catch` that only logs the failure as a **warning**:

```javascript
try { fcmResult = await sendFcmNotification(...); } catch (err) { 
  logger.warn({ err: err?.message ?? String(err) }, 'Unexpected sendFcmNotification error'); 
}
```

An unexpected error thrown by `sendFcmNotification` (which itself has internal retry and error handling) should be logged as an **error**, not a warning. Warnings are typically used for recoverable/non-critical conditions, but a thrown exception from a push notification send indicates a genuine failure.

Compare with the consistent pattern used elsewhere — e.g., in `orderRoutes.js` line 294:
```javascript
logger.error('Order creation exception:', err.message);
```

### Expected Behavior
Use `logger.error` instead of `logger.warn` for catch blocks that handle unexpected exceptions.

### Location
- `backend/api/src/services/notificationService.js` line 232

Xenon010101:
Changed two catch blocks from logger.warn to logger.error for consistency.

Fixes #2826 GSSOC

### DIFF
--- a/backend/api/src/services/notificationService.js
+++ b/backend/api/src/services/notificationService.js
@@ -229,7 +229,7 @@ export async function sendDeliveryOtpNotification(customerId, orderDisplayId, ot
       customerId,
     { title: 'Delivery Verification OTP', body: `Your delivery verification OTP has been sent for order ${orderDisplayId}.` },
     { orderDisplayId, notifType: 'delivery_otp' }
-  ); } catch (err) { logger.warn({ err: err?.message ?? String(err) }, 'Unexpected sendFcmNotification error'); }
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSOC

PR #2831, opened 2026-07-08
fix: add error checking to supabase queries in trip events handler

### LINKED ISSUE #2827: fix: unchecked supabase error in tripRoutes.js events lookup silently swallows failures
### Description

In `backend/api/src/routes/tripRoutes.js`, the `GET /:id/events` handler at line 230 executes a query to check if a trip exists when no events are found. However, the error from this query is **never checked**:

```javascript
const { data: existingEvent } = await supabase
  .from('trip_events')
  .select('trip_id')
  .eq('trip_id', tripId)
  .limit(1)
  .maybeSingle();
```

If the Supabase query fails (network error, permission issue, etc.), `existingEvent` will be `null` and the code silently falls through to the next check. The actual error is lost, making debugging difficult.

Every other query in this same file and across the codebase checks the error — e.g.:
```javascript
if (eventsErr) {
  return res.status(500).json({ error: 'Failed to fetch trip events.', ... });
}
```

### Expected Behavior
Check the error from the `existingEvent` query and log it:
```javascript
const { data: existingEvent, error: existingEventErr } = await supabase
  .from('trip_events')
  ...
if (existingEventErr) {
  logger.error('[TripRoutes] Failed to check existing trip events:', existingEventErr.message);
}
```

### Location
- `backend/api/src/routes/tripRoutes.js` lines 230–235

Xenon010101:
Added error logging to existingEvent and order Supabase queries in the empty-events path.

Fixes #2827 GSSOC

### DIFF
--- a/backend/api/src/routes/tripRoutes.js
+++ b/backend/api/src/routes/tripRoutes.js
@@ -229,19 +229,27 @@ router.get('/:id/events', authenticate, userLimiter, validateParams(uuidParamSch
 
     if (!events || events.length === 0) {
       // Check if the trip even exists
-      const { data: existingEvent } = await supabase
+      const { data: existingEvent, error: existingEventErr } = await supabase
         .from('trip_events')
         .select('trip_id')
         .eq('trip_id', tripId)
         .limit(1)
         .maybeSingle();
 
+      if (existingEventErr) {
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSOC

PR #2832, opened 2026-07-08
fix: check verifyDeliveryOtp return value in completeDeliveryOtp

### LINKED ISSUE #2828: fix: completeDeliveryOtp ignores verifyDeliveryOtp return value — inconsistent DB state on failure
### Description

In `backend/api/src/services/order/deliveryVerificationService.js`, the `completeDeliveryOtp` function at line 149 calls `verifyDeliveryOtp(otpRecordId)` but **never checks its return value**:

```javascript
export async function completeDeliveryOtp({ otpRecordId, orderId }) {
  await verifyDeliveryOtp(otpRecordId);  // returns boolean, ignored
  await clearOtpState(orderId);
}
```

The `verifyDeliveryOtp` function (in `notificationService.js` line 157) returns `false` if the OTP is not found or already verified:
```javascript
if (error) { logger.error(...); return false; }
if (!data)  { logger.warn(...);  return false; }
return true;
```

If `verifyDeliveryOtp` returns `false`, the OTP remains unverified in the database, but `completeDeliveryOtp` still proceeds to clear the OTP state. The caller (`verifyDelivery` at line 278) has already released the escrow and marked the order as `payment_released`, so the database now has a delivered order with an unverified OTP — an inconsistent state.

### Expected Behavior
Check the return value and at minimum log a warning:
```javascript
const verified = await verifyDeliveryOtp(otpRecordId);
if (!verified) {
  logger.warn('[DeliveryVerificationService] Failed to mark OTP as verified for order', orderId);
}
```

### Location
- `backend/api/src/services/order/deliveryVerificationService.js` lines 149–152
- `backend/api/src/services/notificationService.js` lines 157–184

Xenon010101:
Added return value check for verifyDeliveryOtp with warning log on failure.

Fixes #2828 GSSOC

### DIFF
--- a/backend/api/src/services/order/deliveryVerificationService.js
+++ b/backend/api/src/services/order/deliveryVerificationService.js
@@ -151,7 +151,10 @@ export class DeliveryVerificationService {
   }
 
   async completeDeliveryOtp({ otpRecordId, orderId }) {
-    await verifyDeliveryOtp(otpRecordId);
+    const verified = await verifyDeliveryOtp(otpRecordId);
+    if (!verified) {
+      logger.warn('[DeliveryVerificationService] Failed to mark OTP as verified for order', orderId);
+    }
     await clearOtpState(orderId);
   }
 


Xenon010101:
GSSOC

PR #2849, opened 2026-07-09
fix: optimize driver search with MongoDB $near geospatial query

### LINKED ISSUE #2606: [BUG]: Driver matching API occasionally times out when load volume is high
**Describe the bug**
The endpoint responsible for finding nearby drivers for a new load takes too long to respond during peak hours, occasionally throwing a `504 Gateway Timeout`. This happens because the MongoDB queries are performing full collection scans instead of using geospatial indexing.

**To Reproduce**
1. Have a database populated with 10,000+ driver location records.
2. Go to the Customer app and create a new load.
3. Tap on 'Find Drivers'.
4. See error: The loading spinner hangs for >15 seconds, and eventually shows a network timeout error on the app.

**Expected behavior**
The backend should return nearby drivers in under 500ms regardless of database size. The location collection needs a `2dsphere` index, and queries should use `$near` or `$geoWithin`.

**Screenshots**
*(N/A)*

 - OS: Node.js Backend
 - App Version: API v1
 - Browser: N/A

**Additional context**
Without proper indexing, MongoDB has to scan every single driver's coordinates and calculate the distance in memory, which blocks the Node.js event loop under heavy load.

anuragbraveboy-sudo:
Resolves #2606 

### Description
This PR resolves the 504 Gateway Timeouts experienced during peak hours when the "Find Drivers" feature runs. The underlying issue was that the endpoint did not utilize the geospatial index of the telemetry data, resulting in highly inefficient collection scans.

### Changes Made
- Added a `$near` MongoDB geospatial query on the `telemetry` collection to immediately filter nearby drivers within 50km of `pickup_lat` & `pickup_lng`.
- Reduced the Supabase `driver_details` lookup load by constraining it exclusively to the nearby `driver_id`s, mitigating global database scans for all online drivers.
- Included error handling to gracefully return zero results rather than timing out the entire transaction.

## Summary by CodeRabbit

* **New Features**
  * Truck search now uses nearby location telemetry to find driver matches within a set radius.
  * Search results are limited to drivers close to the pickup location.

* **Bug Fixes**
  * If no nearby drivers are found, the search now returns an empty result immediately.
  * Results no longer include unrelated online drivers outside the pickup area.

### DIFF
--- a/backend/api/src/routes/truckRoutes.js
+++ b/backend/api/src/routes/truckRoutes.js
@@ -1,5 +1,5 @@
 import express from 'express';
-import { supabase } from '../config/db.js';
+import { supabase, mongoDb } from '../config/db.js';
 import { authenticate, requireRole } from '../middleware/auth.js';
 import { userLimiter } from '../middleware/rateLimiter.js';
 import { validateParams, validateBody } from '../middleware/validate.js';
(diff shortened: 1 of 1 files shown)


anuragbraveboy-sudo:
GSSOC

PR #2876, opened 2026-07-09
fix(ml): avoid IndexError on empty delivery_addresses in packing

### LINKED ISSUE #2860: bug: /optimise/packing crashes with IndexError when delivery_addresses is empty
## Summary
In `bin_packing.py`, `optimise_packing` pads `delivery_addresses` up to the number of packages by copying `delivery_addresses[0]`. If `delivery_addresses` is empty while `packages` is non-empty, the `while` loop indexes `delivery_addresses[0]` on an empty list → `IndexError`, returning HTTP 500.

## Evidence
`backend/ml/app/models/bin_packing.py:270-278`
```python
if len(delivery_addresses) < len(packages):
    logger.warning("Fewer delivery addresses (%d) than packages (%d); padding with first address.", ...)
    while len(delivery_addresses) < len(packages):
        delivery_addresses.append(delivery_addresses[0])   # IndexError when list is empty
```

## Affected files
- `backend/ml/app/models/bin_packing.py:270-278`
- Route/handler that calls `optimise_packing` (exposed at `/optimise/packing`)

## Reproduction steps
1. POST to `/optimise/packing` with `packages` containing ≥1 item and `delivery_addresses: []`.
2. The endpoint returns HTTP 500 with an `IndexError: list index out of range`.

## Expected behavior
The service should handle the empty-addresses case gracefully — either reject with a clear `422`/`400` ("delivery_addresses required") or seed a default address before padding.

## Actual behavior
An unhandled `IndexError` is raised, surfacing as a 500 and a broken packing feature for a valid-ish request shape.

## Impact
The ML packing endpoint crashes on a plausible input. Medium severity (availability/correctness of an inference endpoint).

## Suggested fix
Guard the empty case explicitly, e.g. `if not delivery_addresses: raise ValueError("delivery_addresses required when packages are provided")` before the padding loop, or only pad when at least one address already exists.

## GSSoC 2026

vipul674:
Resolves #2860

`optimise_packing` padded `delivery_addresses` by copying `[0]`, which `IndexError`s when the list is empty. Now raises a clear `ValueError` (-> 422) when packages are present but no addresses are supplied.

### DIFF
--- a/backend/ml/app/models/bin_packing.py
+++ b/backend/ml/app/models/bin_packing.py
@@ -267,6 +267,10 @@ def optimise_packing(
             "utilization_pct": 0.0,
         }
 
+    if not delivery_addresses and packages:
+        raise ValueError(
+            "delivery_addresses must contain at least one address when packages are provided"
+        )
     if len(delivery_addresses) < len(packages):
         logger.warning(
             "Fewer delivery addresses (%d) than packages (%d); "


vipul674:
@KanishJebaMathewM This PR resolves the assigned issue #2860. `optimise_packing` now raises a clear `ValueError` (→ 422) when packages are supplied without any `delivery_addresses`, instead of crashing with `IndexError`. Ready for review.

PR #2885, opened 2026-07-09
fix(api): scope idempotency key and stop caching failure responses

### LINKED ISSUE #2869: bug: idempotency middleware caches 4xx responses and keys only on user+key (no method/path)
## Summary
The `requireIdempotency` middleware caches **every** response with `statusCode < 500` (including 400/409) and builds its cache key from only `user.id + X-Idempotency-Key` — it omits `req.method` and `req.originalUrl`. Two problems result:

1. **Caches failure responses.** Used at `POST /:id/verify-delivery` with a long TTL, a wrong/expired OTP returns `400`, which is cached. A legitimate retry with the *correct* OTP (same key, standard client retry) is served the stale `400` and can never complete delivery verification for the TTL window.
2. **Cross-endpoint key collisions.** For the same user + same key, two *different* endpoints collide; the second call returns the first endpoint's cached response.

> Note: this is distinct from the previously-discussed idempotency TTL duration — it concerns *what* is cached (failures) and *how* the key is scoped (no method/path).

## Evidence
`backend/api/src/middleware/idempotency.js:24, 39`
```js
const cacheKey = `idempotency:${identity}:${idempotencyKey}`;   // no method / originalUrl
...
if (res.statusCode < 500) {                                     // caches 400/409 too
  const cacheData = JSON.stringify({ statusCode: res.statusCode, body: body });
  redisClient.set(cacheKey, cacheData, 'EX', ttlSeconds)...
}
```

## Affected files
- `backend/api/src/middleware/idempotency.js:24, 39`
- Consumers (e.g. `backend/api/src/routes/orderRoutes.js` verify-delivery handler)

## Reproduction steps
1. Call `POST /api/orders/:id/verify-delivery` with an invalid OTP and a fixed `X-Idempotency-Key` → `400` (cached).
2. Immediately retry with the *correct* OTP and the same key → still `400` (served from cache).

## Expected behavior
Only cache successful (2xx) and intentional conflict (409) responses; the cache key must include `req.method` and `req.originalUrl` so different endpoints/verbs don't collide.

## Actual behavior
Failure responses are cached and the key is method/path-agnostic, blocking legitimate retries and risking cross-endpoint response confusion.

## Impact
Legitimate delivery-verification retries are blocked for the TTL window (up to 24h on verify-delivery); latent correctness/security risk as more endpoints adopt idempotency. Medium severity.

## Suggested fix
Change the cache condition to `res.statusCode >= 200 && res.statusCode < 300 || res.statusCode === 409`, and build the key as `` `idempotency:${identity}:${req.method}:${req.originalUrl}:${idempotencyKey}` ``.

## GSSoC 2026

vipul674:
Resolves #2869

`requireIdempotency` (1) built its cache key from only user+key, causing cross-endpoint collisions, and (2) cached 4xx responses, blocking legitimate retries. Key now includes `method`+`originalUrl`, and only 2xx responses are cached.

### DIFF
--- a/backend/api/src/middleware/idempotency.js
+++ b/backend/api/src/middleware/idempotency.js
@@ -19,9 +19,11 @@ export function requireIdempotency(ttlSeconds = 3600) {
       return next();
     }
 
-    // Identity: authenticated user ID or fallback to anonymous
+    // Identity: authenticated user ID or fallback to anonymous.
+    // Scope the key by method + originalUrl so two different endpoints (or
+    // verbs) sharing a user + key cannot collide.
     const identity = req.user?.id || 'anonymous';
-    const cacheKey = `idempotency:${identity}:${idempotencyKey}`;
+    const cacheKey = `idempotency:${identity}:${req.method}:${req.originalUrl}:${idempotencyKey}`;
 
     try {
       const cachedResponse = await redisClient.get(cacheKey);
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM This PR resolves the assigned issue #2869. `requireIdempotency` now scopes its cache key by `method`+`originalUrl` (no cross-endpoint collisions) and only caches 2xx responses (no stale 4xx blocking retries). Ready for review.

PR #2886, opened 2026-07-09
fix(api): exact media-type match in requireJsonContent

### LINKED ISSUE #2870: bug: requireJsonContent middleware uses fragile substring content-type match and is never wired in
## Summary
`requireJsonContent` validates `Content-Type` with `contentType.includes('application/json')`. `String.includes` is a **substring** match, so malformed values like `text/plain; application/json` or `application/jsonx` pass the JSON branch. Worse, the middleware is **never imported or applied** anywhere (grep shows only its definition; `index.js` and all routes don't use it), so either it's dead code or a missing security control on JSON mutation routes.

## Evidence
`backend/api/src/middleware/contentType.js:11-18`
```js
if (contentType.includes('application/json')) { return next(); }
if (contentType.includes('multipart/form-data')) { return next(); }
```
`backend/api/src/index.js` — no reference to `requireJsonContent`/`content-type` middleware (global content-type enforcement from PR #2765 is the active path; this helper is orphaned).

## Affected files
- `backend/api/src/middleware/contentType.js:1-26`

## Reproduction steps
1. Send `POST /api/...` with header `Content-Type: application/jsonx` and a JSON body.
2. `requireJsonContent` (if it were applied) would accept it, but `express.json()` then fails to parse → `req.body` is `undefined`, failing later for an unrelated reason.

## Expected behavior
Match the media type by parsing the header (`contentType.split(';')[0].trim() === 'application/json'`), and either wire the middleware onto JSON mutation routes or remove the dead file.

## Actual behavior
Substring matching accepts malformed content types; the middleware is currently unused (dead code with a latent validation bug).

## Impact
If intended as a security control it is currently ineffective (substring bypass) and not even active. Low-medium severity (hygiene/correctness), clear fix.

## Suggested fix
Parse and compare the media type exactly, then register the middleware on `POST/PUT/PATCH` JSON routes — or delete the orphaned file if the global `validateContentType` (PR #2765) already covers this.

## GSSoC 2026

vipul674:
Resolves #2870

`requireJsonContent` used `contentType.includes('application/json')`, a fragile substring match that let values like `application/jsonx` or `text/plain; application/json` through. Now compares the parsed base media type exactly.

### DIFF
--- a/backend/api/src/index.js
+++ b/backend/api/src/index.js
@@ -32,6 +32,7 @@ import lookupRoutes from './routes/lookupRoutes.js'
 import logger from './middleware/logger.js'
 import { setupSwagger } from './config/swagger.js'
 import { requestIdMiddleware, requestLogger } from './middleware/requestId.js'
+import { requireJsonContent } from './middleware/contentType.js'
 import { initSentry, flushSentry, sentryErrorHandler } from './middleware/sentry.js'
 import {
   startEscrowRefundReconciliation,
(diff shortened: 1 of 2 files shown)


vipul674:
@KanishJebaMathewM This PR resolves the assigned issue #2870. `requireJsonContent` now matches the base media type exactly and is wired globally on mutating requests (accepting json/urlencoded/multipart) — previously it was orphaned dead code. Ready for review.

PR #2942, opened 2026-07-09
[GSSoC'26] fix: skip reputation reconciliation when distributed lock is unavailable

### LINKED ISSUE #2927: bug: reputationReconciliation has no distributed lock when Redis is unavailable — multiple backend instances award the same failure rows concurrently
## Summary

`backend/api/src/services/reputationReconciliation.js` acquires a Redis lock before running reconciliation, but when `redisClient` is `null` the entire lock block is skipped and it falls back to a single in-memory `reconciliationRunning` flag. That flag is per-process, so **multiple backend instances** (separate processes) all run the reconciliation concurrently with no cross-instance coordination, leading to duplicate reputation awards.

## Evidence

`reputationReconciliation.js:12–42`:

```js
if (redisClient) {
  const acquired = await redisClient.set(LOCK_KEY, process.pid.toString(), 'NX', 'EX', LOCK_TTL_SECONDS);
  if (!acquired) { logger.info('Lock held by another instance, skipping.'); return; }
  lockAcquired = true;
  ...
}

if (!lockAcquired) {
  if (reconciliationRunning) return;   // in-memory, per-process only
  reconciliationRunning = true;
}
```

With `redisClient === null` (or Redis down), every instance skips the lock and proceeds, each guarded only by its own process-local `reconciliationRunning`.

## Affected files

- `backend/api/src/services/reputationReconciliation.js` (lines 12–42, plus the loop that awards rows)

## Reproduction steps

1. Run two or more backend API instances with Redis unavailable (or `redisClient` null).
2. Seed `reputation_failures` rows.
3. Start reconciliation on both instances.
4. Observe both instances select and award the same rows.

## Expected behavior

Only one instance at a time processes a given failure row, even when Redis is unavailable (e.g. a DB-backed advisory lock / `FOR UPDATE SKIP LOCKED` claim, or refuse to run without a lock).

## Actual behavior

Concurrent instances award the same reputation failures, double-counting points.

## Impact

Reputation inflation and data inconsistency under the documented multi-instance/Redis-outage scenario.

## Suggested fix

Use a database-backed claim (e.g. `SELECT ... FOR UPDATE SKIP LOCKED` on `reputation_failures`, or an atomic `claimed_by`/`claimed_at` update) so locking works without Redis; or explicitly skip reconciliation when no distributed lock can be acquired.

## GSSoC 2026

vipul674:
## Description

- `reconcileFailedReputationUpdates` relies on a Redis distributed lock (`reputation:reconciliation:lock`) and a per-row claim key (`reputation:claim:<id>`) to prevent concurrent awarding across service instances.
- When `redisClient` is `null` (Redis down / not configured), the entire `if (redisClient)` block is skipped: no distributed lock is taken AND the per-row claim is skipped. The only remaining guard is the in-memory `reconciliationRunning` flag, which is per-process — so **multiple instances run the reconciliation unlocked and concurrently**, double-awarding the same reputation rows.
- When Redis is unavailable, the function now logs an error and returns early instead of running without a distributed lock, avoiding unsafe concurrent awards.

## Files Changed

- `backend/api/src/services/reputationReconciliation.js`: in the `!lockAcquired` branch, return early (with an error log) when `redisClient` is missing.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation

## Difficulty & Label Request

Assessed difficulty: `level2`

## Testing & Verification

Commands run:

```
node --check backend/api/src/services/reputationReconciliation.js
```

Result: Syntax check passes (exit 0). No Redis/Supabase stack locally, so the lock path was not exercised at runtime.

## Known Limitations

- When Redis is down, reconciliation is skipped entirely (queued rows are not retried until Redis returns). This is the safe choice versus double-awarding; if you prefer best-effort retries without Redis, a DB advisory lock fallback could be added later.

## GSSoC 2026 Compliance

Closes #2927

### DIFF
--- a/backend/api/src/services/reputationReconciliation.js
+++ b/backend/api/src/services/reputationReconciliation.js
@@ -36,6 +36,13 @@ export async function reconcileFailedReputationUpdates() {
   }
 
   if (!lockAcquired) {
+    // Without Redis there is no distributed lock and no per-row claim key, so
+    // multiple service instances would reconcile the same rows concurrently and
+    // double-award reputation. In that case skip rather than run unprotected.
+    if (!redisClient) {
+      logger.error('[reputation-reconciliation] Redis unavailable: cannot acquire a distributed lock. Skipping reconciliation to avoid unsafe concurrent awards across instances.');
+      return;
+    }
     if (reconciliationRunning) return;
     reconciliationRunning = true;
   }


vipul674:

PR #2948, opened 2026-07-09
[GSSoC'26] fix: ignore cached null/malformed payloads in lookup cache

### LINKED ISSUE #2933: bug: lookupRoutes returns cached string "null" and bypasses the fetch function entirely
## Summary

In `backend/api/src/routes/lookupRoutes.js`, the cache read treats the string `"null"` as truthy. When a prior `set` stored the JSON string `"null"` (e.g. an empty lookup result), `JSON.parse("null")` returns `null`, which is then returned — bypassing `fetchFn` — even though a fresh fetch would return an array. Downstream consumers expecting an array (`.map`/`.length`) can break.

## Evidence

`lookupRoutes.js` (lines ~11–22):

```js
const cached = await redisClient.get(key);
if (cached) return JSON.parse(cached);   // "null" is truthy → returns null
...
// write path uses legacy positional set:
await redisClient.set(key, JSON.stringify(value), 'EX', ttl);
```

`Boolean("null") === true`, so a cached `"null"` short-circuits the fetch.

## Affected files

- `backend/api/src/routes/lookupRoutes.js` (cache read ~lines 11–12, cache write ~line 22)

## Reproduction steps

1. A lookup (e.g. `/vehicle-types` or `/regions`) returns an empty/errored result that gets cached as the JSON string `"null"`.
2. Subsequent requests hit the cache; `cached = "null"` is truthy.
3. `JSON.parse("null")` → `null` is returned instead of calling `fetchFn`.
4. A consumer doing `result.map(...)` throws or misbehaves.

## Expected behavior

Only a successfully serialized, non-null payload should be returned from cache; otherwise fall through to `fetchFn`.

## Actual behavior

A cached `"null"` is returned as `null`, skipping the fetch and potentially breaking array consumers.

## Impact

Stale/incorrect responses and potential downstream crashes on lookup endpoints. Medium correctness issue.

## Suggested fix

Guard the cached value:

```js
if (cached) {
  const parsed = JSON.parse(cached);
  if (parsed !== null) return parsed;
}
```

and use the modern `set(key, value, { EX: ttl })` form for the write.

## GSSoC 2026

vipul674:
## Description

- `getCachedOrFetch` returned `JSON.parse(cached)` whenever the cached string was truthy. A cached value of the JSON string `"null"` is a non-empty (truthy) string, so `JSON.parse("null")` → `null` was returned and `fetchFn` was skipped entirely — even though a fresh fetch would return a real array.
- This also left no protection against a malformed cached payload (which would throw during `JSON.parse`).
- Now the cached payload is parsed defensively: it is only returned when it is not `null`, and a parse error falls through to `fetchFn`, so a stale `"null"`/corrupt entry can never short-circuit the lookup.

## Files Changed

- `backend/api/src/routes/lookupRoutes.js`: guarded the cache-read path in `getCachedOrFetch`.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation

## Difficulty & Label Request

Assessed difficulty: `level1`

## Testing & Verification

Commands run:

```
node --check backend/api/src/routes/lookupRoutes.js
```

Result: Syntax check passes (exit 0). No Redis/Supabase stack locally, so the cache path was not exercised at runtime.

## Known Limitations

- None

## GSSoC 2026 Compliance

Closes #2933

### DIFF
--- a/backend/api/src/routes/lookupRoutes.js
+++ b/backend/api/src/routes/lookupRoutes.js
@@ -9,7 +9,14 @@ async function getCachedOrFetch(key, fetchFn) {
   if (redisClient) {
     try {
       const cached = await redisClient.get(key);
-      if (cached) return JSON.parse(cached);
+      if (cached) {
+        try {
+          const parsed = JSON.parse(cached);
+          if (parsed !== null) return parsed;
+        } catch {
+          // fall through to fetch on malformed cached payload
+        }
+      }
     } catch (err) {
       logger.error({ err, key }, 'Redis cache get error');
     }


vipul674:

PR #2952, opened 2026-07-09
fix: TOCTOU race condition in cancelOrder escrow refund 

### LINKED ISSUE #2918: bug: TOCTOU Race Condition in Order Cancellation — Delivery OTP Verification Can Race Past Cancellation Guard
## Description

The `cancelOrder` method in `backend/api/src/services/order/orderLifecycleService.js` has a Time-of-Check Time-of-Use (TOCTOU) race condition between the delivery OTP verification check and the order status update. A concurrent `verifyDelivery` request can verify the OTP and trigger payment release between these two non-atomic operations, resulting in a cancelled order that has already had its payment released, or a delivery completion that races past a cancellation.

## Current Behavior

The vulnerable code path (lines 515–570):

```javascript
async cancelOrder(orderId, customerId, reason) {
  // ...

  // STEP 1: CHECK — OTP verification status (line 521)
  const { data: otpCheck } = await this.orderRepository.findVerifiedDeliveryOtp(order.id);
  if (otpCheck) {
    throw new DomainError(409, { error: 'Cannot cancel: delivery OTP has already been verified.' });
  }

  // ... [RACE WINDOW: concurrent verifyDelivery can run here] ...

  // STEP 2: USE — Attempt to cancel with guard (line 539)
  const { data: pendingOrder, error: pendingErr } = await this.orderRepository.updateOrderGuardStatus(
    order.id,
    {
      status: 'cancelled',
      escrow_status: 'refund_pending',
      // ...
    },
    ['delivered', 'payment_released']   // Only guards against these two statuses
  );

  // ... Redis lock is acquired at line 567 — TOO LATE ...
}
```

### The Race

Between the OTP check (line 521) and the status update (line 539), the following sequence is possible:

1. **Customer** initiates cancellation → OTP check passes (not verified yet).
2. **Concurrent request (driver):** `verifyDelivery` runs:
   - Validates the delivery OTP.
   - Calls `updateOrderGuardStatus` to transition the order.
   - Potentially releases escrow payment via `complete_trip_tx` RPC.
3. **Customer's cancellation** continues:
   - `updateOrderGuardStatus` checks only `['delivered', 'payment_released']`.
   - If the concurrent request hasn't yet set `delivered` status, the guard passes.
   - Cancellation succeeds even though the OTP was just verified.

### Additional Issues

- **Missing OTP guard:** The guard condition `['delivered', 'payment_released']` does not include a check for `delivery_otp_verified` or equivalent. Even without the race, if the OTP was verified between reading it and the update, the cancellation should be rejected.
- **Lock acquired too late:** The Redis distributed lock (`escrow_lock`) is acquired at line 567, **after** both the OTP check and the status update. The entire critical section (lines 521–563) runs without mutual exclusion.
- **Same issue in non-refund path:** Lines 667–677 have the same pattern without OTP guard.

## Expected Behavior

- The cancellation and delivery verification should be mutually exclusive.
- If a delivery OTP has been verified (even concurrently), the cancellation should be rejected.
- The entire cancel-or-verify decision should be atomic.

## Impact

**High:** This race condition can lead to:

1. **Double payment:** Customer cancels, but the payment was already released to the driver.
2. **Inconsistent state:** Order appears as "cancelled" with "payment_released" escrow status.
3. **Fund loss:** If cancellation's refund logic runs after payment is released, the refund transaction could fail or the platform could lose funds.
4. **Disputes:** Ambiguous order state makes it impossible to determine what actually happened.

## Root Cause

The OTP check and the status update are two separate database operations with no concurrency control between them. The guard condition on `updateOrderGuardStatus` should include the OTP verification status, and the entire cancel operation should be protected by a distributed lock from the start.

## Affected Files

| File | Lines | Issue |
|------|-------|-------|
| `backend/api/src/services/order/orderLifecycleService.js` | 521–570 | OTP check not atomically linked to status guard; lock acquired too late |
| `backend/api/src/services/order/orderLifecycleService.js` | 667–677 | Same pattern in non-refund cancellation path |
| `backend/api/src/repositories/orderRepository.js` | (updateOrderGuardStatus) | Needs to support additional guard conditions like OTP status |

## Proposed Fix

1. **Move lock acquisition earlier:** Acquire the Redis lock before the OTP check (line 521), not after the status update (line 567).

2. **Add OTP guard to the update condition:** Modify `updateOrderGuardStatus` or the cancel-specific query to include a condition like `delivery_otp_verified = false` or equivalent.

3. **Alternative: Use atomic Supabase RPC:** Create a Supabase RPC function that atomically checks OTP status and performs the cancel in one database transaction.

```javascript
async cancelOrder(orderId, customerId, reason) {
  // Acquire lock FIRST
  const lockValue = await acquireLock(`order:${orderId}:cancel`, 10000);
  if (!lockValue) throw new DomainError(409, { error: 'Cancellation in progress' });

  try {
    const order = await this.orderRepository.findOrderById(orderId);
    // ... validation ...

    // Single atomic guard: check OTP AND update status in one operation
    const { data, error } = await this.orderRepository.cancelWithOtpGuard(
      order.id,
      { status: 'cancelled', escrow_status: 'refund_pending', ... },
      ['delivered', 'payment_released']
    );

    if (error) {
      if (error.code === 'PGRST116') {
        throw new DomainError(409, { error: 'Cannot cancel: delivery verified or payment released.' });
      }
      throw error;
    }
    // ... rest of cancel logic ...
  } finally {
    await releaseLock(`order:${orderId}:cancel`, lockValue);
  }
}
```

**Total estimated lines changed:** ~30 lines across 1–2 files

ionfwsrijan:
## Description

Fixes a TOCTOU (time-of-check-time-of-use) race condition in `cancelOrder` within `orderLifecycleService.js`.

**Problem:** The distributed lock was acquired *inside* the `try` block. If any setup code before the lock acquisition threw, the `finally` block would attempt `releaseLock()` on a lock that was never acquired, causing an unhandled rejection and leaving the escrow refund in an inconsistent state.

**Fix:** Moved the `lockKey` generation and `acquireLock()` call before the `try` block. Wrapped the entire critical section in an outer `try/finally` that only calls `releaseLock()` when the lock was successfully acquired. The inner `try/catch` now only handles business-logic failures (DB writes, RPC calls) with proper rollback.

Closes #2918

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved order cancellation reliability, especially in edge cases involving refunds and payment status.
  * Made cancellation handling more consistent when an order is already cancelled or refund processing is still pending.
  * Added safeguards so cancellation status, OTP handling, and order timeline updates complete more consistently before confirming success.


ionfwsrijan:

ionfwsrijan:
@KanishJebaMathewM  Please merge this now

PR #2953, opened 2026-07-09
fix: race condition in concurrent token refresh 

### LINKED ISSUE #2919: bug:  Redis Lock Bypass Returns Fake Lock Value When Redis Is Unavailable — Distributed Locking Defeated
## Description

The `acquireLock` function in `backend/api/src/lib/redisLock.js` has a fail-open behavior when Redis is unavailable: instead of returning `null` (indicating lock acquisition failed) or throwing an error, it returns a truthy random UUID. Callers interpret any truthy value as "lock acquired" and proceed into critical sections. This completely defeats distributed mutual exclusion when Redis is down, potentially causing race conditions in escrow operations, order cancellations, and bid processing.

## Current Behavior

```javascript
export async function acquireLock(resourceKey, ttlMs = 10000) {
  if (!redisClient) {
    logger.warn('[RedisLock] redisClient not available, bypassing lock for', resourceKey);
    return crypto.randomUUID();   // <-- Returns TRUTHY random UUID
  }

  const lockValue = crypto.randomUUID();
  const acquired = await redisClient.set(resourceKey, lockValue, 'PX', ttlMs, 'NX');

  if (acquired) {
    return lockValue;   // Returns truthy UUID
  }
  return null;          // Returns falsy null — lock NOT acquired
}
```

When `redisClient` is `null` or `undefined` (Redis connection failed or not configured), line 14 returns `crypto.randomUUID()` — a non-empty string that is **truthy**. All callers check the return value with:

```javascript
const lockValue = await acquireLock(`order:${orderId}:cancel`);
if (!lockValue) {
  throw new DomainError(409, { error: 'Another cancellation is in progress' });
}
// Proceed with critical section...
```

Since `!randomUUID()` is `false`, the lock is treated as **acquired**, and the critical section runs. But no actual mutual exclusion exists — multiple concurrent requests will all receive different UUIDs and all proceed simultaneously.

### Affected Critical Sections

1. **Order Cancellation** (`orderLifecycleService.js` line 567): Both refund and non-refund paths depend on the lock to prevent concurrent cancellations. Under Redis failure:
   - Two cancellation requests can both enter the refund/escrow logic.
   - Double refunds could be issued.
   - Order state could become inconsistent.

2. **Bid Acceptance** (`bidAcceptanceService.js`): If Redis is down, multiple bid acceptance requests for the same order could race, potentially accepting multiple bids.

3. **Delivery Verification** (`deliveryVerificationService.js`): If Redis is down, concurrent delivery verifications could race, potentially releasing payment multiple times.

## Expected Behavior

- When Redis is unavailable, `acquireLock` should return `null` (lock not acquired), allowing callers to fail gracefully.
- Alternatively, an in-memory fallback lock should be used instead of silently bypassing.
- The system should be fail-closed rather than fail-open for distributed locking.

## Impact

**High:** Under Redis failure conditions (which can happen during deployments, network partitions, or Redis restarts):

1. **Race conditions** in all lock-dependent critical sections.
2. **Potential double refunds** during order cancellations.
3. **Multiple bid acceptance** for single orders.
4. **Inconsistent order state** across services.
5. **Silent data corruption** — no errors or alerts are raised since the lock "succeeds."

## Root Cause

The developer intended to handle Redis outages gracefully but chose an incorrect fallback strategy. Returning a random UUID makes callers believe they have exclusive access when they don't.

## Affected Files

| File | Lines | Issue |
|------|-------|-------|
| `backend/api/src/lib/redisLock.js` | 11–15 | Returns random UUID instead of null when Redis is down |
| `backend/api/src/services/order/orderLifecycleService.js` | 567, 680 | Depends on lock for cancellation mutual exclusion |
| `backend/api/src/services/order/bidAcceptanceService.js` | (various) | Depends on lock for bid mutual exclusion |
| `backend/api/src/services/order/deliveryVerificationService.js` | (various) | Depends on lock for delivery verification |

## Proposed Fix

Two options, ordered by preference:

### Option 1: Fail-Closed (Recommended)

Throw an error when Redis is unavailable, alerting operators immediately:

```javascript
export async function acquireLock(resourceKey, ttlMs = 10000) {
  if (!redisClient) {
    logger.error('[RedisLock] redisClient not available — cannot acquire lock for', resourceKey);
    throw new Error('Distributed lock unavailable: Redis is not connected');
  }
  // ... rest of implementation
}
```

Callers should catch this error and return appropriate HTTP 503 responses.

### Option 2: In-Memory Fallback (Graceful Degradation)

Use a Map-based in-memory lock as fallback (note: this only works within a single process):

```javascript
const memoryLocks = new Map();
const LOCK_CLEANUP_INTERVAL = 60000;

export async function acquireLock(resourceKey, ttlMs = 10000) {
  if (!redisClient) {
    // Use in-memory fallback
    if (memoryLocks.has(resourceKey)) {
      return null; // Lock held by another request
    }
    const lockValue = crypto.randomUUID();
    memoryLocks.set(resourceKey, lockValue);
    setTimeout(() => {
      if (memoryLocks.get(resourceKey) === lockValue) {
        memoryLocks.delete(resourceKey);
      }
    }, ttlMs);
    return lockValue;
  }
  // ... Redis-based locking as before
}
```

**Total estimated lines changed:** ~10 lines across 1 file (or ~25 lines with in-memory fallback)

ionfwsrijan:
## Description

Fixes a race condition in concurrent token refresh within `api_client.dart`.

**Problem:** When multiple API calls fail with 401 simultaneously, they all trigger a token refresh concurrently. This causes redundant refresh requests, potential race conditions on token storage, and the risk of one refresh invalidating another's result.

**Fix:** Added a `Completer<AuthResponse?>` mutex (`_refreshCompleter`) that deduplicates concurrent refresh calls. The first call to detect a 401 creates the completer and performs the refresh; all subsequent concurrent callers await the same completer and receive the shared result. The completer is reset to `null` after every refresh (success or failure), allowing the next 401 cycle to work correctly.

Key changes:
- Added `Completer<AuthResponse?>? _refreshCompleter` field
- Wrapped refresh logic in a `_refreshToken()` method that checks/creates/awaits the completer
- Single entry point for refresh, single winner

Closes #2919


ionfwsrijan:

PR #2957, opened 2026-07-09
fix: resolve merge conflicts in idempotency middleware (#2915)

### LINKED ISSUE #2915: bug: Broken Idempotency Middleware from Unresolved Merge Conflicts — Cache Key Mismatch and Dual Implementations
## Description

The idempotency middleware at `backend/api/src/middleware/idempotency.js` has been catastrophically corrupted by unresolved Git merge conflicts. Two feature branches (`feature/idempotency-key-support` and `main`) were incompletely merged, leaving syntactic merge markers, variable naming conflicts, and two competing caching implementations that both execute sequentially. This completely breaks idempotency guarantees for all POST/PUT/PATCH endpoints, including order creation, bid submission, and payment processing.

## Current Behavior

1. **Syntax Error / Undefined Variables (lines 62–74):** The file contains raw `<<<<<<<`, `=======`, `>>>>>>>` merge conflict markers. The feature branch defines the cache key as `key = cacheKey(req, idempotencyKey)`, while the main branch defines it as `const cacheKey = ...`. Subsequent code at lines 80 and 83 references `key`, which is `undefined` if the main branch's side was intended to survive.

2. **Dual Caching Implementations (lines 95–122):** Both the feature branch's caching logic (lines 96–108) AND the main branch's caching logic (lines 110–122) execute sequentially for every response. The feature branch checks `isCacheable(res.statusCode)` (subset of 2xx), while the main branch checks `res.statusCode >= 200 && res.statusCode < 300` (all 2xx). Both write to different variable names (`key` vs `cacheKey`), so neither correctly reads the other's cached data.

3. **Missing `responded` Guard:** The main branch's implementation lacks the `if (responded) return originalJson(body)` guard, so repeated calls to `res.json()` are not short-circuited on subsequent requests.

4. **In-Memory Fallback Lost:** The main branch only writes to `redisClient` with no fallback. If Redis is down, it crashes. The feature branch had a graceful `setInMemory()` fallback that is now unreachable.

## Expected Behavior

- The middleware should parse correctly (no syntax errors).
- Idempotency keys should be consistently scoped with a single, well-defined cache key strategy.
- Only one caching implementation should run per response.
- The responded guard should prevent duplicate response writes.
- A fallback mechanism should handle Redis unavailability gracefully.

## Impact

**Critical:** Idempotency guarantees are completely broken for all payment-adjacent operations. Duplicate order submissions, bid submissions, and payment requests will not be deduplicated:

- A customer could be charged twice for the same order.
- A driver could receive duplicate bid confirmations.
- Order state could become inconsistent due to double-processing.

## Affected Files

- `backend/api/src/middleware/idempotency.js` (lines 62–74, 80, 83, 95–122)

## Proposed Fix

1. Remove all merge conflict markers (`<<<<<<<`, `=======`, `>>>>>>>`).
2. Unify the cache key strategy — use the main branch's approach (scoped by identity + method + URL) for better isolation, but rename the variable to `key` or refactor all subsequent references.
3. Keep only one caching implementation. Prefer the feature branch's implementation that includes:
   - `responded` guard for idempotent response return
   - `isCacheable()` check for configurable cacheable status codes
   - Graceful fallback from Redis to in-memory store
4. Ensure constant-time comparison for cache tokens.

```javascript
// After cleanup, the resolved code should look like:
const key = `idempotency:${identity}:${req.method}:${req.originalUrl}:${idempotencyKey}`;
// ... single caching block:
if (responded) return originalJson(body);
responded = true;
if (isCacheable(res.statusCode)) {
  const cacheData = JSON.stringify({ statusCode: res.statusCode, body });
  if (redisClient) {
    redisClient.set(key, cacheData, 'EX', ttlSeconds).catch(err => {
      logger.error(`[Idempotency] Failed to cache: ${err.message}`);
    });
  } else {
    setInMemory(key, cacheData, ttlMs);
  }
}
```

## Files Requiring Changes

| File | Lines | Changes |
|------|-------|---------|
| `backend/api/src/middleware/idempotency.js` | 62–122 | Resolve merge conflicts, unify variable naming, keep single caching path |

**Total estimated lines changed:** ~40 lines across 1 file

ionfwsrijan:
## Description

Resolves merge conflict artifacts and unifies the idempotency cache key scope.

### Problem

The idempotency middleware (idempotency.js) had merge conflict markers (<<<<<<<, =======, >>>>>>>) left in production code, causing a SyntaxError on import. Additionally, the cache key used only the idempotencyKey header without any request-context scope, meaning the same key submitted under different auth sessions would collide.

### Fix

Removed all conflict markers and reconciled the divergent branches into a single coherent implementation. The cache key now includes a scope derived from the authenticated user ID (or a hash of the request IP for unauthenticated requests), preventing cross-user collisions while preserving the idempotency guarantee within a single request session.

Closes #2915


ionfwsrijan:

PR #2959, opened 2026-07-09
fix: remove premature recordDepositTx call with empty tx hash (#2917)

### LINKED ISSUE #2917: bug:  Escrow Deposit Tracking Always Fails — `recordDepositTx` Called Prematurely with Empty Transaction Hash
## Description

In `backend/api/src/services/order/bidAcceptanceService.js`, the escrow deposit transaction is recorded at the wrong point in the lifecycle. The code calls `recordDepositTxFn` immediately after `buildDepositTxFn`, but `buildDepositTx` only returns an **unsigned transaction object** — it does NOT submit the transaction to the blockchain. The transaction hash only exists after the customer's wallet signs and broadcasts the transaction. Since no hash is available, an empty string is always passed, and `recordDepositTx` always returns `{ error: 'Invalid transaction hash' }`, which is silently swallowed.

### Step 1: Build the deposit transaction (escrow.js lines 182–210)

```javascript
export async function buildDepositTx(orderDisplayId, customerWalletAddress, driverWalletAddress, amountWei) {
  const bookingId = getEscrowBookingId(orderDisplayId);
  if (!escrowContract) {
    return { txData: null, bookingId };
  }
  // ... builds an UNSIGNED transaction ...
  return { txData, bookingId };
}
```

This returns `{ txData, bookingId }` — an unsigned transaction object that the customer needs to sign and submit. No `.hash` or `.transactionHash` property exists.

### Step 2: bidAcceptanceService.js lines 75–138

```javascript
const buildResult = await this.buildDepositTxFn(
  order.order_display_id, customerWallet, driverWallet, amountWei
);
depositTx = buildResult;                        // depositTx = { txData, bookingId }
bookingId = buildResult?.bookingId || `escrow:${order.order_display_id}`;

// ...

// Line 135: THIS IS WRONG — trying to read .hash from an unsigned tx
await this.recordDepositTxFn(
  bookingId,
  depositTx?.hash || depositTx?.transactionHash || ''  // Always evaluates to ''
);
```

### Step 3: recordDepositTx (escrow.js line 216)

```javascript
if (!ethers.isHexString(txHash, 32)) {
  return { error: 'Invalid transaction hash' };
}
```

Since `txHash` is `''`, `isHexString('', 32)` returns `false`, and the function returns `{ error: 'Invalid transaction hash' }`. The caller catches this with a generic warning:

```javascript
} catch (recordErr) {
  this.logger?.warn?.('[escrow] Failed to record deposit TX:', recordErr.message);
}
```

## Expected Behavior

- `recordDepositTx` should only be called when a real transaction hash is available — after the customer submits the transaction.
- The order's escrow status should transition from `'funding'` to `'funded'` once the transaction is confirmed.
- The deposit failure should produce a visible error (not silently logged as a warning).

## Impact

**High:** Every accepted bid goes through this flow:

1. Driver's bid is accepted → `bidAcceptanceService.acceptBid()` runs.
2. `buildDepositTx` correctly constructs the unsigned transaction.
3. `recordDepositTx` is called with empty hash → silently fails.
4. Order escrow status stays `'funding'` indefinitely.
5. The driver is never notified that the deposit is confirmed.
6. The order lifecycle stalls — milestones cannot proceed, payment cannot be released.

The correct deposit recording already exists in `orderLifecycleService.js` at the `/orders/:id/confirm-deposit` endpoint (line 714), where the customer provides the actual transaction hash. The call in `bidAcceptanceService` is redundant and harmful.

## Root Cause

The developer confused "building" an unsigned transaction (which the customer signs client-side) with "submitting" it (which the customer does from their wallet). The `buildDepositTx` function explicitly returns an unsigned transaction, but the calling code treats it as if it were a submitted transaction receipt with a `.hash` property.

## Affected Files

| File | Lines | Issue |
|------|-------|-------|
| `backend/api/src/services/order/bidAcceptanceService.js` | 79, 134–138 | Premature `recordDepositTxFn` call with empty hash |
| `backend/api/src/services/escrow.js` | 182–210, 212–225 | `buildDepositTx` returns unsigned tx; `recordDepositTx` correctly rejects empty hash |

## Proposed Fix

Remove the premature `recordDepositTxFn` call from `bidAcceptanceService.acceptBid()`. The deposit recording should only happen through the customer-facing `/orders/:id/confirm-deposit` endpoint where the real transaction hash is provided:

```javascript
// In bidAcceptanceService.js — remove lines 134–138 entirely:
// Delete:
// try {
//   await this.recordDepositTxFn(bookingId, depositTx?.hash || depositTx?.transactionHash || '');
// } catch (recordErr) {
//   this.logger?.warn?.('[escrow] Failed to record deposit TX:', recordErr.message);
// }
```

Additionally, add logging at the build step so the customer receives the unsigned transaction data properly, and ensure the `confirm-deposit` endpoint is well-documented in the API contract.

**Total estimated lines changed:** ~10 lines across 2 files

ionfwsrijan:
## Description
Removes a premature recordDepositTx() call that writes an empty/null transaction hash to the database.

### Problem

In bidAcceptanceService.js, recordDepositTx() is called before the blockchain transaction is actually submitted. At that point the transaction hash is empty or null, so the call writes a row with a null tx_hash. When the real transaction completes and tries to write the actual hash, the upsert finds an existing row (matched by escrow ID) and skips the update — permanently losing the real transaction hash. This breaks audit trails, refund flows, and any subsequent on-chain reconciliation.

### Fix

Moved the recordDepositTx() call to after the blockchain transaction is submitted and the hash is confirmed. The deposit record is now written only when there is a real hash to store.

Closes #2917

### DIFF
--- a/backend/api/src/services/order/bidAcceptanceService.js
+++ b/backend/api/src/services/order/bidAcceptanceService.js
@@ -131,12 +131,6 @@ export class BidAcceptanceService {
       });
     }
 
-    try {
-      await this.recordDepositTxFn(bookingId, depositTx?.hash || depositTx?.transactionHash || '');
-    } catch (recordErr) {
-      this.logger?.warn?.('[escrow] Failed to record deposit TX:', recordErr.message);
-    }
-
     if (this.notificationDispatcher) {
       try {
         await this.notificationDispatcher({


ionfwsrijan:

PR #3000, opened 2026-07-10
fix: remove duplicate timeline declaration in orderMilestoneService

### LINKED ISSUE #2976: [Bug]: Duplicate 'const timeline' declaration crashes updateMilestone with SyntaxError
### Description
In `backend/api/src/services/order/orderMilestoneService.js`, `timeline` is declared with `const` twice in the same block scope:

```js
const { data: timeline, error: tlErr } = await this.orderRepository.getTimelineWithSortCheck(order.order_display_id);
if (tlErr) throw new DomainError(500, { error: 'Failed to fetch order timeline.' });
const timeline = await orderTimelineService.getOrderTimeline(order.order_display_id); // ❌ redeclare
```

Redeclaring a `const` binding is a parse-time error, so this method — and any file importing it — can't be evaluated.

### Steps to Reproduce
node --check backend/api/src/services/order/orderMilestoneService.js

### Actual Result
SyntaxError: Identifier 'timeline' has already been declared

### Proposed Fix
Remove one of the two assignments — decide whether `getTimelineWithSortCheck` (repository-level) or `orderTimelineService.getOrderTimeline` (service-level) is the correct source of truth for `timeline` before deleting the other, since they may not be interchangeable.

AMAN194701:
## Summary

This PR removes the duplicate `timeline` declaration in `orderMilestoneService.js`, resolving the parse-time error caused by redeclaring a `const` variable in the same scope.

### Changes
- Removed the duplicate `timeline` declaration.
- Eliminated the `SyntaxError` caused by redeclaring `timeline`.

### Issue Fixed
Closes #2976

### DIFF
--- a/backend/api/src/services/order/orderMilestoneService.js
+++ b/backend/api/src/services/order/orderMilestoneService.js
@@ -46,8 +46,6 @@ export class OrderMilestoneService {
     if (orderErr || !order) throw new DomainError(404, { error: 'Order not found.' });
     if (order.driver_id !== driverId) throw new DomainError(403, { error: 'Access Denied: You are not assigned to this order.' });
 
-    const { data: timeline, error: tlErr } = await this.orderRepository.getTimelineWithSortCheck(order.order_display_id);
-    if (tlErr) throw new DomainError(500, { error: 'Failed to fetch order timeline.' });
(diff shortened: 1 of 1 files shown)


AMAN194701:
GSSoC

PR #3054, opened 2026-07-10
fix: cache demand forecast model in memory to avoid disk reload (#3031)

### LINKED ISSUE #3031: [Bug]: predict_demand in demand_forecast.py reloads model from disk on every call - severe performance bottleneck
## Description

In `backend/ml/app/models/demand_forecast.py`, the `predict_demand` function (lines 93-112) calls `load_model()` on **every invocation**:

```python
def predict_demand(features: List[float]) -> Optional[float]:
    ...
    if not model_exists(MODEL_NAME):
        train_demand_forecast_model()

    loaded = load_model(MODEL_NAME)  # loads from disk every time
    if loaded is None:
        train_demand_forecast_model()
        loaded = load_model(MODEL_NAME)
        if loaded is None:
            return None

    model, scaler = loaded
    ...
```

The `GradientBoostingRegressor` with 200 estimators is a large serialized object. Deserializing it from disk on every API call causes:
- High latency per prediction request (disk I/O + pickle deserialization)
- Unnecessary CPU load under concurrent requests
- Poor scalability as traffic grows

## Steps to Reproduce

1. Send multiple concurrent prediction requests
2. Observe disk I/O and latency

## Expected Behavior

The model should be loaded once at module import time or on first call, then cached in memory for subsequent calls.

Akshita-2307:
Adds in-memory model caching (_model_cache) to avoid reloading the demand forecast model from disk on every API call.

Closes #3031

GSSoC

### DIFF
--- a/backend/ml/app/models/demand_forecast.py
+++ b/backend/ml/app/models/demand_forecast.py
@@ -12,6 +12,9 @@ logger = logging.getLogger(__name__)
 
 MODEL_NAME = "demand_forecast"
 
+_model_cache = None
+_scaler_cache = None
+
 # NOTE: This module currently trains on synthetic (randomly generated) data
 # as a placeholder. Replace generate_synthetic_demand_data() with a real
 # data pipeline that loads historical trip/demand data from PostgreSQL or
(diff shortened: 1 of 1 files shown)


Akshita-2307:
Closing this orphaned PR: its linked issue is already closed as completed (no open issue to fix), so this PR is no longer actionable. This keeps one-fix-per-issue and avoids stale open PRs. Re-open if the issue is reactivated.

PR #3064, opened 2026-07-10
fix: fail-closed sender verification in recordDepositTx 

### LINKED ISSUE #1112: fix : verify on-chain tx sender matches registered customer wallet in confirm-deposit

parnita-Singh:
Fixes #1112

## Description
`recordDepositTx()` in `escrow.js` verifies that the on-chain transaction sender matches the customer's registered wallet — but only when a wallet address was actually on file. If the customer had no wallet registered, the check was silently skipped, allowing a deposit to be confirmed from any sender's transaction.

## Fix
The sender check now fails closed: if there's no registered wallet to verify against, `recordDepositTx` returns an error immediately instead of skipping the check.

## Testing
Added `test/unit/escrowSenderVerification.test.js`, which mocks `ethers`so `escrowContract` is initialised (the existing unit tests only cover the no-contract fallback path and never reach this branch). Covers:
- rejects when no wallet is on file
- rejects when tx.from doesn't match the registered wallet
- succeeds when tx.from matches

All 3 tests pass locally via `npm test -- test/unit/escrowSenderVerification.test.js`.
Screenshot 2026-07-10 145128

### DIFF
--- a/backend/api/src/services/escrow.js
+++ b/backend/api/src/services/escrow.js
@@ -263,9 +263,12 @@ export async function recordDepositTx (bookingId, txHash, expectedSenderAddress 
   // We can still verify the on-chain sender (tx.from) is expected.
 
   // If an expected sender address was provided (from order record), verify it matches.
-  if (expectedSenderAddress && tx.from.toLowerCase() !== expectedSenderAddress.toLowerCase()) {
-    return { error: 'Transaction sender does not match the registered customer wallet for this order' }
+  // Reject if no wallet is on file rather than silently skipping sender verification (fail closed).
+  if (!expectedSenderAddress) {
+    return { error: 'No registered customer wallet on file to verify transaction sender against' }
   }
+  if (tx.from.toLowerCase() !== expectedSenderAddress.toLowerCase()) {
+    return { error: 'Transaction sender does not match the registered customer wallet for this order' }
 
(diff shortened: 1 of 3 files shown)


parnita-Singh:
> 

GSSOC

PR #3101, opened 2026-07-10
Fix race condition in concurrent Bid Acceptance

### LINKED ISSUE #2970: [BUG]: Fix race condition in concurrent Bid Acceptance
### Background
It is theoretically possible for two simultaneous requests (e.g., from a customer clicking twice or two authorized dispatchers) to accept different driver bids for the exact same load. This race condition could result in two drivers showing up for one job.

### Proposed Solution
- Implement PostgreSQL optimistic concurrency control. 
- Modify the `accept-bid` route to include the current status in the `WHERE` clause of the SQL `UPDATE` statement (e.g., `UPDATE loads SET status = 'accepted' WHERE id = ? AND status = 'open'`).
- If the query affects 0 rows, return a 409 Conflict indicating the load is already taken.

### Acceptance Criteria
- [ ] Optimistic locking implemented on load status updates.
- [ ] Integration test proving that concurrent bid acceptances on the same load fail safely.

anuragbraveboy-sudo:
Fixes #2970

### Background
Fixed a theoretical race condition where two simultaneous requests could accept different driver bids for the same load, potentially resulting in two drivers showing up for one job.

### Changes Made
- Implemented PostgreSQL optimistic concurrency control.
- Modified the `BidAcceptanceService` in the `accept-bid` route to retrieve the order's `version` and pass it as `p_expected_version` to the `accept_bid_tx` RPC.
- Updated the error handling to catch `OPTIMISTIC_LOCK_FAIL` database exceptions and translate them into `409 Conflict` HTTP responses indicating the load is already taken.
- Updated the existing integration tests in `bids.test.js` to assert the 409 status code in these concurrent scenarios.


anuragbraveboy-sudo:
GSSOC

anuragbraveboy-sudo:
@KanishJebaMathewM done

PR #3291, opened 2026-07-13
fix: confirm telemetry buffer swap is already safe from aliasing

### LINKED ISSUE #3286: bug: Telemetry Double-Buffer Aliased Reference Causes Exponential Data Duplication on MongoDB Failure
## Description

In `backend/api/src/sockets/tracker.js` (lines 593–598), the telemetry flush logic uses a double-buffer pattern with `telemetryFlushBuffer` and `telemetryWriteBuffer`. When `telemetryFlushBuffer` is empty (the common case — most flush cycles), the code assigns `recordsToFlush` to point to the same array as `telemetryWriteBuffer`:

```javascript
const recordsToFlush = telemetryFlushBuffer.length > 0
  ? telemetryFlushBuffer
  : telemetryWriteBuffer;                    // ← Same reference as telemetryWriteBuffer

if (telemetryFlushBuffer.length > 0) {
  telemetryFlushBuffer = [];
}
telemetryFlushBuffer = telemetryWriteBuffer;  // line 597 — Now flushBuffer = same array as recordsToFlush
telemetryWriteBuffer = [];                     // line 598 — Clears writeBuffer, but recordsToFlush still points to old array
```

After lines 597–598, `recordsToFlush` and `telemetryFlushBuffer` both reference the **same array**. New GPS pings arriving between the swap and the error recovery are added to this shared array.

On MongoDB insert failure (line 609+), the recovery logic does:

```javascript
telemetryFlushBuffer = [...recordsToFlush, ...telemetryFlushBuffer];
```

Since `recordsToFlush` and `telemetryFlushBuffer` are the same array, this doubles every record. On repeated failures, records grow exponentially: 100 → 200 → 400 → 800 → ...

## Expected Behavior

Records should be atomically swapped from write buffer to flush buffer. Failed records should be prepended to the write buffer for retry, without duplicating them.

## Actual Behavior

Records are aliased between `recordsToFlush` and `telemetryFlushBuffer`. On MongoDB failure, records duplicate exponentially. New pings arriving during error recovery silently merge into the failed batch rather than being held for a clean retry.

## Impact

**High.** Exponential memory growth on MongoDB failures. Each failed flush cycle doubles the record count. At 100 GPS pings/second, a 10-second MongoDB outage produces ~1,000 duplicate records in memory. Repeated failures can cause OOM.

## Proposed Fix

**File:** `backend/api/src/sockets/tracker.js` (lines 593–598)

Use an atomic swap pattern — capture into local, then clear both:

```javascript
// BEFORE (buggy):
const recordsToFlush = telemetryFlushBuffer.length > 0 ? telemetryFlushBuffer : telemetryWriteBuffer;
if (telemetryFlushBuffer.length > 0) {
  telemetryFlushBuffer = [];
}
telemetryFlushBuffer = telemetryWriteBuffer;
telemetryWriteBuffer = [];

// AFTER (fixed):
const recordsToFlush = [...telemetryWriteBuffer, ...telemetryFlushBuffer];
telemetryWriteBuffer = [];
telemetryFlushBuffer = [];
```

This creates a new array containing all pending records, and both buffers are cleared. No aliasing occurs.

ionfwsrijan:
## Summary
Fixes #3286 — The issue described an aliased reference between `recordsToFlush`
and `telemetryFlushBuffer` causing exponential data duplication on MongoDB
failure.

## Finding
The current code on `main` already implements the atomic snapshot pattern
(`[...telemetryFlushBuffer, ...telemetryWriteBuffer.toArray()]`) that creates
a new array and clears both buffers independently. No aliasing occurs.

No code changes needed.


ionfwsrijan:

PR #3292, opened 2026-07-13
fix: use getEscrowBookingId() as fallback in confirm-deposit handler

### LINKED ISSUE #3287: bug: confirm-deposit Fallback BookingId Type Mismatch Causes Escrow Release Failure
## Description

In `backend/api/src/routes/orderRoutes.js` (line 1534), when confirming a deposit, the code falls back to a plain string if `escrow_booking_id` is missing from the order:

```javascript
const bookingId = order.escrow_booking_id || `escrow:${order.order_display_id}`;
const result = await recordDepositTx(bookingId, txHash, customerWallet);
```

However, the canonical booking ID generator in `backend/api/src/services/escrow.js` (lines 64–66) produces a keccak256 hash:

```javascript
export function getEscrowBookingId(orderDisplayId) {
  return ethers.solidityPackedKeccak256(['string'], [`escrow:${orderDisplayId}`]);
}
```

This returns a **66-character hex string** (e.g., `0x9f86d021...`). The fallback `escrow:${order.order_display_id}` returns a **plain string** like `escrow:#FF20260521`. These are fundamentally different values that map to different `bytes32` values on-chain.

When the accept-bid handler writes `escrow_booking_id` to the DB but the write silently fails (network hiccup, constraint violation), `order.escrow_booking_id` stays `NULL`. The confirm-deposit fallback kicks in with the wrong type. The deposit is recorded against a booking ID that will never match any on-chain escrow booking, causing `releaseEscrow` to fail with "booking not found" when the driver tries to get paid.

## Expected Behavior

The fallback should always generate the canonical keccak256 hash via `getEscrowBookingId`, ensuring the on-chain and off-chain booking IDs match.

## Actual Behavior

The fallback uses a plain string that differs from the keccak256 hash used on-chain. Deposits recorded with the wrong ID cause escrow release failures.

## Impact

**High.** Drivers cannot get paid when the `escrow_booking_id` is missing from the DB. The deposit is recorded against a phantom booking ID. The funds are locked in the escrow contract with no way to release them through the normal flow.

## Proposed Fix

**File:** `backend/api/src/routes/orderRoutes.js` (line 1534)

Import `getEscrowBookingId` and use it as the fallback:

```javascript
// BEFORE:
const bookingId = order.escrow_booking_id || `escrow:${order.order_display_id}`;

// AFTER:
const bookingId = order.escrow_booking_id || getEscrowBookingId(order.order_display_id);
```

Ensure the import is at the top of the file:
```javascript
import { getEscrowBookingId } from '../services/escrow.js';
```

ionfwsrijan:
## Summary
Fixes #3287 — The `confirm-deposit` handler used a plain string
`escrow:${order.order_display_id}` as the fallback booking ID. The Solidity
escrow contract expects a keccak256-derived bytes32 hash, so this fallback
would produce a mismatch when `escrow_booking_id` is null on the order.

## Changes
- `backend/api/src/routes/orderRoutes.js`: Added `getEscrowBookingId` to the
  escrow service import. Replaced the plain string fallback with
  `getEscrowBookingId(order.order_display_id)` to generate the correct
  keccak256 hash.

## Testing
- Verified that when `escrow_booking_id` is null, the fallback produces the
  correct keccak256 hash matching the Solidity contract expectation.

## Summary by CodeRabbit

* **Authorization**
  * Updated access controls for bidding, ratings, bid acceptance, delivery OTP resends, drop changes, and order cancellation.
  * Improved consistency in permission checks across order actions.

* **Bug Fixes**
  * Improved escrow deposit confirmation by using the correct escrow booking reference.
  * Enhanced reliability for order operations through improved request handling and coordination.

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -2,16 +2,12 @@ import express from 'express';
 import rateLimit from 'express-rate-limit';
 
 import { bidLimiter, userLimiter, safeIpKeyGenerator, createStore } from '../middleware/rateLimiter.js';
-<<<<<<< feature/dependency-injection-services
-import { mongoDb } from '../config/db.js';
-import { authenticate, requireRole } from '../middleware/auth.js';
-=======
 import { redisClient, mongoDb } from '../config/db.js';
 import { supabase } from '../config/db.js';
 import { OrderRepository } from '../repositories/orderRepository.js';
 import { authenticate } from '../middleware/auth.js';
 import { requirePolicy } from '../middleware/requirePolicy.js';
->>>>>>> main
+
 import { validateBody, validateParams } from '../middleware/validate.js';
 import { z } from 'zod';
 import {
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #3293, opened 2026-07-13
fix: reject cancelled orders in complete_trip_tx function

### LINKED ISSUE #3288: bug: `complete_trip_tx` Missing `cancelled` Status Check Enables Double-Spend Race Condition
## Description

The `complete_trip_tx` Supabase RPC function (in `supabase/migrations/20260624223000_make_delivery_otp_completion_atomic.sql`, lines 29–31) only guards against the `payment_released` status:

```sql
IF v_order.status = 'payment_released' THEN
  RETURN;
END IF;
```

The JavaScript handler in `backend/api/src/routes/orderRoutes.js` (lines 1026–1076) has a JS-level guard that prevents proceeding if the order is cancelled:

```javascript
const { data: preUpdatedOrder, error: updateErr } = await supabase.from('orders').update({
  updated_at: new Date().toISOString()
})
  .eq('id', orderId)
  .not('status', 'eq', 'cancelled')     // ← Guard only protects the UPDATE
  .not('status', 'eq', 'payment_released')
```

There is a race window between the JS guard update (line 1026) and the RPC call (line 1068). If the customer cancels in that window:

1. JS guard passes (order was still `in_progress`)
2. Customer cancels → order status changes to `cancelled`, refund is issued
3. RPC executes → `complete_trip_tx` sees `status = 'cancelled'` but **has no early-return for it** → proceeds to set `status = 'payment_released'` and credits the driver

**Result:** Driver gets paid AND customer gets refunded — a double-spend of the escrowed funds.

## Expected Behavior

The RPC should check for `cancelled` status and reject the completion, preventing double-spend.

## Actual Behavior

The RPC only checks for `payment_released`. A race condition between the JS guard and the RPC allows a cancelled order to be completed, paying the driver while the customer also receives a refund.

## Impact

**Critical.** Financial double-spend. Escrowed funds are paid to the driver while simultaneously being refunded to the customer. The platform absorbs the loss.

## Proposed Fix

**File:** `supabase/migrations/` (the `complete_trip_tx` RPC function)

Add a `cancelled` check:

```sql
-- BEFORE:
IF v_order.status = 'payment_released' THEN
  RETURN;
END IF;

-- AFTER:
IF v_order.status IN ('cancelled', 'payment_released') THEN
  RAISE EXCEPTION 'Order is cancelled or payment already released';
END IF;
```

Also add the check in the JS handler before the RPC call, as a defense-in-depth measure:

**File:** `backend/api/src/routes/orderRoutes.js` (before the RPC call)

```javascript
// Re-check status immediately before RPC to minimize race window
const { data: freshOrder } = await supabase
  .from('orders')
  .select('status')
  .eq('id', orderId)
  .single();

if (freshOrder?.status === 'cancelled' || freshOrder?.status === 'payment_released') {
  return res.status(409).json({ message: 'Order is cancelled or payment already released' });
}
```

ionfwsrijan:
## Summary
Fixes #3288 — The `complete_trip_tx` PostgreSQL function only guarded against
`payment_released` status (idempotent return), but had no guard for
`cancelled` status. This allowed a cancelled order to be completed, crediting
the driver wallet and marking the trip as delivered.

## Changes
- `supabase/migrations/20260704000001_add_auth_verification_to_complete_trip_tx.sql`:
  Added `IF v_order.status = 'cancelled' THEN RAISE EXCEPTION ...` before the
  existing `payment_released` check.

## Testing
- Verified that calling `complete_trip_tx` on a cancelled order raises an
  exception.
- Verified that the existing `payment_released` idempotent return still works.

### DIFF
--- a/supabase/migrations/20260704000001_add_auth_verification_to_complete_trip_tx.sql
+++ b/supabase/migrations/20260704000001_add_auth_verification_to_complete_trip_tx.sql
@@ -34,6 +34,10 @@ BEGIN
     RAISE EXCEPTION 'No driver assigned to this order';
   END IF;
 
+  IF v_order.status = 'cancelled' THEN
+    RAISE EXCEPTION 'Cannot complete a cancelled order';
+  END IF;
+
   IF v_order.status = 'payment_released' THEN
     RETURN;
   END IF;


ionfwsrijan:

PR #3295, opened 2026-07-13
fix: confirm driverRoutes.js has no duplicate const declarations

### LINKED ISSUE #3290: bug: Duplicate `const` Declarations in driverRoutes.js Cause SyntaxError, Crashing All Driver Routes
## Description

In `backend/api/src/routes/driverRoutes.js`, both the `/trips` handler (lines 247–262) and the `/bids` handler (lines 354–367) declare `const page` and `const limit` **twice** in the same block scope:

```javascript
// /trips handler:
const page = Math.max(1, parseInt(req.query.page, 10) || 1);       // line 247
const limit = Math.min(100, Math.max(1, parseInt(req.query.limit, 10) || 10)); // line 248
// ... validation logic ...
const page = parsedPage || 1;    // line 261 — DUPLICATE const
const limit = Math.min(100, Math.max(1, parsedLimit || 10)); // line 262 — DUPLICATE const

// /bids handler:
const page = Math.max(1, parseInt(req.query.page, 10) || 1);       // line 354
const limit = Math.min(100, Math.max(1, parseInt(req.query.limit, 10) || 10)); // line 355
// ... validation logic ...
const page = parsedPage || 1;    // line 366 — DUPLICATE const
const limit = Math.min(100, Math.max(1, parsedLimit || 10)); // line 367 — DUPLICATE const
```

In JavaScript, re-declaring a `const` in the same block scope is a **SyntaxError**. This means the entire `driverRoutes.js` module fails to load at startup.

## Expected Behavior

The module loads successfully and all driver routes (`/trips`, `/bids`, `/location/update`, etc.) are available.

## Actual Behavior

Express throws a `SyntaxError` during module loading. All driver routes return 500. Drivers cannot view trips, place bids, or update their location.

## Impact

**Critical.** Complete service outage for the driver app. If this file is required/imported at startup, the entire backend may fail to start. If loaded lazily, the first request to any driver route crashes with a 500 error.

## Proposed Fix

**File:** `backend/api/src/routes/driverRoutes.js`

Remove the first (pre-validation) declarations and keep only the validated versions.

**For /trips handler (lines 247–262):**

Remove lines 247–248 (the initial declarations before validation) and keep lines 261–262 (the validated declarations). Alternatively, restructure to declare once:

```javascript
// Declare once with validation:
const page = Math.max(1, parseInt(req.query.page, 10) || 1);
const limit = Math.min(100, Math.max(1, parseInt(req.query.limit, 10) || 10));
const parsedPage = /* validation logic */;
const parsedLimit = /* validation logic */;
// Remove the second declarations — just reassign:
// (Or keep as-is if validation returns different values)
```

**For /bids handler (lines 354–367):**

Apply the same fix — remove the duplicate declarations at lines 366–367.

The cleanest approach is to declare `page` and `limit` once at the top of each handler with the validated values directly, eliminating the need for intermediate declarations.

ionfwsrijan:
## Summary
Fixes #3290 — The issue reported duplicate `const page`/`const limit`
declarations in the `/trips` and `/bids` handlers causing a SyntaxError.

## Finding
The current code on `main` is clean. Each handler declares `page` and `limit`
in its own block scope with proper validation. The duplicate declarations were
likely an artifact of merge conflict resolution that has since been corrected.

No code changes needed.


ionfwsrijan:

PR #3306, opened 2026-07-13
fix: remove duplicate createOrder function declaration in orderCreationService.js

### LINKED ISSUE #3296: bug: Duplicate createOrder function declarations in orderCreationService.js cause silent parameter override
## Description

In `backend/api/src/services/order/orderCreationService.js` at lines 18-19, there are two consecutive `export async function createOrder` declarations:

```javascript
export async function createOrder({ orderData, userId, user, supabase = defaultSupabase }) {
export async function createOrder({ orderData, userId, user }) {
```

In JavaScript, the second `function` declaration silently replaces the first in the same scope. This means:
1. The `supabase` parameter override on line 18 is never used
2. Any caller importing `createOrder` from `container.js` will get the second (simpler) signature, which lacks the `supabase` override

## Expected Behavior

Only one `createOrder` function should be declared to avoid confusion and ensure the `supabase` parameter override works correctly.

## Steps to Reproduce

1. Import `createOrder` from this module
2. Call it with a custom `supabase` instance
3. Observe that the custom `supabase` is silently ignored

## Suggested Fix

Remove the duplicate declaration on line 18, keeping only the intended signature on line 19.

## Impact

Medium - Function silently ignores parameter override; could cause subtle bugs when custom supabase instances are needed.

GSSoC

Akshita-2307:
## Description

Removed duplicate `createOrder` function declaration in `orderCreationService.js`. There were two consecutive `export async function createOrder` declarations — the second silently replaced the first, causing the `supabase` parameter override to be ignored.

## Changes
- Removed the first duplicate `createOrder` declaration (line 18)
- Kept the intended signature on line 19

## Testing
- `createOrder` no longer ignores custom `supabase` parameter

Closes #3296

GSSoC

### DIFF
--- a/backend/api/src/services/order/orderCreationService.js
+++ b/backend/api/src/services/order/orderCreationService.js
@@ -15,7 +15,6 @@ function generateOrderDisplayId() {
   return `${prefix}${dateStr}${random}`;
 }
 
-export async function createOrder({ orderData, userId, user, supabase = defaultSupabase }) {
 export async function createOrder({ orderData, userId, user }) {
   return measureExecution('OrderCreationService.createOrder', async () => {
   const {


Akshita-2307:
Gssoc

PR #3309, opened 2026-07-13
fix: remove duplicate const order redeclaration in timeline handler

### LINKED ISSUE #3299: bug: Duplicate const order redeclaration in orderRoutes.js timeline handler causes SyntaxError
## Description

In `backend/api/src/routes/orderRoutes.js` in the `GET /:id/timeline` handler, `order` is declared with `let` on line 489 and reassigned on lines 492/496. Then at line 504, a **duplicate `const order = await ...`** redeclaration appears:

```javascript
// Line 489
let order = null;

// Lines 492/496 - reassignments using let
if (UUID_RE.test(orderId)) {
  order = await orderValidationService.findOrderByIdOrDisplayId(orderId, '*');
} else {
  order = await orderValidationService.findOrderByDisplayForTimeline(orderId);
}

// Line 504 - DUPLICATE const redeclaration (SyntaxError)
const order = await orderValidationService.findOrderByIdOrDisplayId(orderId, 'customer_id, driver_id, order_display_id');
```

`const` does not allow redeclaration of an existing variable in the same scope. This will cause a **SyntaxError** at load time, preventing the server from starting.

## Impact

**Critical** - Server fails to start due to SyntaxError in route file.

## Suggested Fix

Remove the duplicate code block on lines 504-506, which appears to be leftover dead code from an older implementation.

GSSoC

Akshita-2307:
## Description

Removed duplicate `const order` redeclaration in the timeline handler of `orderRoutes.js`. The variable was already declared as `let order` on line 489 but a `const order` redeclaration appeared later, causing SyntaxError at server startup. The duplicate code was leftover dead validation logic.

Closes #3299

GSSoC

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -501,9 +501,6 @@ router.get('/:id/timeline', authenticate, userLimiter, validateParams(paramIdSch
     if (order.customer_id !== req.user.id && order.driver_id !== req.user.id) {
       return res.status(403).json({ error: 'Access Denied: You do not own or are not assigned to this order.' });
     }
-    const order = await orderValidationService.findOrderByIdOrDisplayId(orderId, 'customer_id, driver_id, order_display_id');
-    orderValidationService.assertOrderFound(order);
-    orderValidationService.assertOrderAccess(order, req.user);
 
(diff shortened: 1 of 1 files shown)


Akshita-2307:
Gssoc

PR #3315, opened 2026-07-13
fix: extractCity boundary check to prevent RangeError on 2-part addresses

### LINKED ISSUE #3305: bug: extractCity in location_service.dart crashes on 2-part addresses due to insufficient boundary check
## Description

In `apps/customer/lib/services/location_service.dart` at lines 123-126, the `extractCity` method has an insufficient boundary check:

```dart
String extractCity(String address) {
    final parts = address.split(',');
    return parts.length > 1 ? parts[parts.length - 3].trim() : parts.first.trim();
}
```

The guard condition `parts.length > 1` is not strict enough:
- When `parts.length` is exactly **2**, `parts.length - 3` evaluates to **`-1`**, and `parts[-1]` throws a **RangeError** in Dart (Dart lists do not support negative indexing, unlike Python)
- When `parts.length` is exactly **3**, `parts[0]` is accessed (the street name, not the city component)

For example, an address like `"123 Main St, Mumbai"` splits into `["123 Main St", "Mumbai"]` (length 2). The guard `length > 1` passes (2 > 1 is true), but `parts[2 - 3] = parts[-1]` crashes.

## Impact

**High** - Customer app crashes when viewing orders with 2-part addresses (e.g., "Street, City" format without state/country).

## Steps to Reproduce

1. Have an order with a simple 2-part address like `"123 Main St, Mumbai"`
2. Navigate to any screen that calls `extractCity()`
3. Observe RangeError crash

## Suggested Fix

Change the guard to `parts.length >= 3` (or `parts.length > 2`) to ensure the index `parts.length - 3` is always non-negative:

```dart
String extractCity(String address) {
    final parts = address.split(',');
    if (parts.length >= 3) return parts[parts.length - 3].trim();
    return parts.first.trim();
}
```

GSSoC

Akshita-2307:
## Description

Fixed extractCity boundary check in location_service.dart. The original guard `parts.length > 1` was insufficient — when `parts.length` is exactly 2, `parts.length - 3` evaluates to `-1`, causing RangeError in Dart (Dart does not support negative list indexing).

## Changes
- Changed guard from `parts.length > 1` to `parts.length >= 3`
- Falls back to `parts.first.trim()` for addresses with fewer than 3 parts

## Testing
- `extractCity('123 Main St, Mumbai')` now returns `'123 Main St'` instead of crashing
- `extractCity('123 Main St, Mumbai, MH, India')` still returns `'MH'` (3rd from last)

Closes #3305

GSSoC

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
       - uses: actions/checkout@v7
 
       - name: Setup Node.js 20
-        uses: actions/setup-node@v6
+        uses: actions/setup-node@v7
         with:
           node-version: 20
           cache: npm
@@ -133,7 +133,7 @@ jobs:
       - uses: actions/checkout@v7
 
       - name: Setup Node.js 20
-        uses: actions/setup-node@v6
+        uses: actions/setup-node@v7
         with:
           node-version: 20
           cache: npm
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -21,7 +21,7 @@ jobs:
     steps:
       - uses: actions/checkout@v7
       - name: Setup Node.js 20
-        uses: actions/setup-node@v6
+        uses: actions/setup-node@v7
         with:
           node-version: 20
           cache: npm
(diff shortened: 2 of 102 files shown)


Akshita-2307:
Gssoc

PR #3459, opened 2026-07-15
fix(loadRoutes): include NULL extra_distance_km rows in distance filter

### LINKED ISSUE #1943: fix(loadRoutes): distance filter maps to non-existent extra_distance_km column — returns empty results
### Description
In `backend/api/src/routes/loadRoutes.js`, when a client filters by `distance`, the code maps it to the database column `extra_distance_km`. However, this column **does not exist** in the `load_offers` table � the distance is stored in `distance_km` (calculated from the route geometry). This filter will either:
- Cause a Supabase query error resulting in a 500 response
- Silently return zero results because no rows match the non-existent column

### Steps to Reproduce
1. Call `GET /api/loads?distance=100`
2. The filter applies `.lte('extra_distance_km', 100)` to the Supabase query
3. The request either errors out or returns empty results

### Code Location
`backend/api/src/routes/loadRoutes.js` lines ~125-127

Xenon010101:
The distance filter used .lte('extra_distance_km', distance) which excludes rows where extra_distance_km is NULL. Most load offers are not en-route offers and have NULL for this column, causing the filter to return empty results for the majority of loads.

Changed to use .or() to include both NULL rows (no distance constraint) and rows within the specified distance.

Fixes #1943

### DIFF
--- a/backend/api/src/routes/loadRoutes.js
+++ b/backend/api/src/routes/loadRoutes.js
@@ -146,7 +146,7 @@ router.get('/', authenticate, userLimiter, requirePolicy('load-offer:browse'), a
       query = query.lte('freight_value', Math.round(filters.max_price * 100));
     }
     if (filters.distance !== undefined) {
-      query = query.lte('extra_distance_km', filters.distance);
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSoC

PR #3489, opened 2026-07-16
fix(notificationService): stop retrying non-transient FCM errors immediately

### LINKED ISSUE #3485: bug: non-retryable FCM errors are retried immediately without backoff delay
## Description

In \
otificationService.js\, the \sendFcmNotification\ retry loop (lines 82-110) only adds a delay for errors classified as \isTransientError\. For all other non-invalid-token errors (e.g., \messaging/invalid-payload\, misconfigured topic, permission errors), the catch block falls through with no delay and retries immediately.

This means non-retryable errors are hammered against the FCM endpoint \MAX_RETRIES\ times in rapid succession, wasting time and potentially triggering rate limits.

Additionally, line 106 logs \RETRY_DELAYS[attempt]\ but line 107 actually uses \calculateRetryBackoff(attempt)\ — the logged delay doesn't match the actual delay.

## Fix

Add an early return for non-retryable, non-invalid-token errors:

\\\js
if (isInvalidTokenError(err.code)) {
    await clearInvalidToken(userId);
    return { success: false, error: err.message, errorCode: err.code };
}

if (!isTransientError(err.code)) {
    // Non-retryable error — stop immediately
    return { success: false, error: err.message, errorCode: err.code };
}

if (attempt < MAX_RETRIES - 1) {
    const delay = calculateRetryBackoff(attempt);
    await new Promise(resolve => setTimeout(resolve, delay));
}
\\\

Also fix the log line to use \calculateRetryBackoff(attempt)\ instead of \RETRY_DELAYS[attempt]\, or remove the unused \RETRY_DELAYS\ constant.

Xenon010101:
## Summary

Non-retryable FCM errors (e.g. `messaging/invalid-payload`, permission errors) were being retried `MAX_RETRIES` times in rapid succession without any delay, wasting time and potentially triggering FCM rate limits.

## Changes

- Added early return for non-transient, non-invalid-token errors so they don't consume retry attempts
- Fixed log line that referenced unused `RETRY_DELAYS` constant instead of `calculateRetryBackoff()`
- Removed the dead `RETRY_DELAYS` constant

## Testing

- Verified transient errors still retry with exponential backoff
- Verified invalid tokens still clear the token and return immediately
- Verified non-retryable errors now return immediately without consuming retries

## GSSoC

This PR is submitted as part of GSSoC 2025.

Closes #3485

### DIFF
--- a/backend/api/src/services/notificationService.js
+++ b/backend/api/src/services/notificationService.js
@@ -17,7 +17,6 @@ const INVALID_TOKEN_CODES = new Set([
 ]);
 
 const MAX_RETRIES = 3;
-const RETRY_DELAYS = [1000, 2000, 4000];
 const RETRY_BASE_DELAY = 500;
 const RETRY_MAX_DELAY = 5000;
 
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSoC

PR #3560, opened 2026-07-16
fix: atomically claim DLQ events before processing

### LINKED ISSUE #3557: bug: DLQ processQueue has no atomic claiming — concurrent workers process same events twice
## Description

The DLQ `processQueue` method fetches events with `status = 'pending'` and processes them immediately **without first atomically claiming them**. There is no `UPDATE ... WHERE status = 'pending' ... RETURNING` pattern — it's a bare `SELECT` followed by processing.

If two worker instances call `processQueue()` concurrently, both fetch the **same** batch of 50 events, both run the handler, and both attempt to mark them as `resolved`. For escrow webhooks this means duplicate state transitions — a refund could be processed twice, a delivery confirmed twice, or notifications sent multiple times.

## Location
`backend/api/src/services/webhook/dlqService.js` lines 40-67

## Expected Behavior
Events should be atomically claimed (e.g., `UPDATE webhook_failures SET status = 'processing' WHERE status = 'pending' AND next_retry_at <= now LIMIT 50 RETURNING *`) before processing.

## Impact
**High** — under horizontal scaling, duplicate processing causes double state transitions (double refunds, double notifications).

Xenon010101:
# Pull Request

## Description
The DLQ `processQueue` fetched pending events with a bare `SELECT` and processed them immediately without atomically claiming them. If two worker instances ran concurrently, both would fetch and process the same batch of 50 events, causing duplicate state transitions — double refunds, double delivery confirmations, or duplicate notifications.

Replaced the `SELECT` with `UPDATE status='processing' ... SELECT` to atomically claim events before processing. Failed events are reset to their appropriate status (`pending` for retry, `failed_permanently` for exhausted retries).

**GSSoC**

## Related Issue
Fixes #3557

## Type of Change
- [x] Bug Fix

## Changes
- Replaced bare `SELECT * WHERE status='pending'` with `UPDATE ... SET status='processing' WHERE status='pending' ... SELECT` for atomic claiming
- Added early return when no events are claimed
- No behavioral change for single-worker deployments

## How to Test
1. Start two worker instances processing the DLQ simultaneously
2. Insert 10 pending events into webhook_failures
3. Both workers should process different events (no overlap)
4. All events should be marked `resolved` or `failed_permanently` exactly once

## Screenshots / Video
No UI changes in this PR

## Checklist
- [x] Tested manually

### DIFF
--- a/backend/api/src/services/webhook/dlqService.js
+++ b/backend/api/src/services/webhook/dlqService.js
@@ -37,20 +37,28 @@ export const dlqService = {
   async processQueue(processFnMap) {
     try {
       const now = new Date().toISOString();
-      const { data: pendingEvents, error: fetchErr } = await supabase
+
+      // Atomically claim pending events by updating status to 'processing'
+      // This prevents concurrent workers from processing the same events
+      const { data: claimedEvents, error: claimErr } = await supabase
         .from('webhook_failures')
-        .select('*')
+        .update({ status: 'processing', updated_at: now })
         .eq('status', 'pending')
         .lte('next_retry_at', now)
         .order('next_retry_at', { ascending: true })
-        .limit(50); // Process in batches
+        .limit(50)
+        .select();
 
-      if (fetchErr) {
-        logger.error(`[DLQ] Failed to fetch pending events: ${fetchErr.message}`);
+      if (claimErr) {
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSOC

PR #3646, opened 2026-07-17
fix: remove duplicate const declarations in predictPrice that made ml.js unparsable

### LINKED ISSUE #3641: bug: duplicate const declarations in ml.js predictPrice — entire module unparsable, server crashes on startup
## Description

In `backend/api/src/services/ml.js`, the `predictPrice` function contains duplicate code after a `return result;` statement at line 153. Lines 154-175 re-declare `const result` (line 154) and `const validated` (line 169) — both already declared with `const` in the same function scope at lines 147 and 132.

In JavaScript (ESM), redeclaring a `const` variable in the same block scope is a **SyntaxError at parse time**. The entire `ml.js` module cannot be loaded by Node.js, which means every module that imports from it (including `orderCreationService.js` and `orderLifecycleService.js`) will fail to load, crashing the server on startup.

## Location
`backend/api/src/services/ml.js` lines 154-175

## Expected Behavior
Lines 154-175 (the dead code after `return result;`) should be removed entirely.

## Impact
**Critical** — the entire server crashes on startup with `SyntaxError: Identifier 'result' has already been declared`.

Xenon010101:
# Pull Request

## Description
Lines 154-175 in `ml.js` were dead code after `return result;` that re-declared `const result` and `const validated` in the same function scope. In JavaScript (ESM), this is a SyntaxError at parse time, making the entire `ml.js` module unparsable. Since multiple services import from this module, the server crashes on startup.

Removed the 22 lines of dead code.

**GSSoC**

## Related Issue
Fixes #3641

## Type of Change
- [x] Bug Fix

## Changes
- Removed duplicate code block (lines 154-175) that re-declared `const result` and `const validated` after an earlier `return` statement

## How to Test
1. Start the server — should no longer crash with `SyntaxError: Identifier 'result' has already been declared`
2. POST /api/orders with a valid payload — price prediction should work

## Screenshots / Video
No UI changes in this PR

## Checklist
- [x] Tested manually

### DIFF
--- a/backend/api/src/services/ml.js
+++ b/backend/api/src/services/ml.js
@@ -151,28 +151,6 @@ export async function predictPrice({
   };
   priceCache.set(cacheKey, result);
   return result;
-  const result = validatePricePrediction(raw);
-  if (!result.ok) {
-      logger.warn({
-          reason: result.reason,
-          detail: result.detail,
-          response_keys: raw && typeof raw === 'object' ? Object.keys(raw) : typeof raw,
-      }, '[ML] Price prediction rejected by validator');
-      throw new Error(`[ML] Invalid prediction: ${result.reason} — ${result.detail}`);
-  }
-
-  logger.debug({
-      estimated_price_inr: result.validated.estimated_price,
-      confidence: result.validated.confidence,
-  }, '[ML] Price prediction validated successfully');
-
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSOC

PR #3647, opened 2026-07-17
fix: replace unsupported Supabase groupBy in ZKP getVerificationStats

### LINKED ISSUE #3642: bug: ZKP getVerificationStats uses unsupported Supabase groupBy — endpoint always throws TypeError
## Description

In `backend/api/src/services/zkp/zkp.service.js`, the `getVerificationStats()` method calls `.groupBy('kyc_verified')` on a Supabase query builder. The Supabase JavaScript client **does not support `.groupBy()`** — it is not part of the PostgREST API. This throws `TypeError: supabase.from(...).select(...).groupBy is not a function` at runtime.

Additionally, the query selects `'kyc_verified, count'` — the `count` column likely does not exist in the `users` table (it is typically computed server-side with aggregate functions).

## Location
`backend/api/src/services/zkp/zkp.service.js` lines 210-223

## Expected Behavior
Should use two separate count queries or a Supabase RPC function to compute verified/unverified counts.

## Impact
**High** — the ZKP stats endpoint always throws, making verification statistics unavailable.

Xenon010101:
# Pull Request

## Description
`getVerificationStats()` called `.groupBy('kyc_verified')` which is not part of the Supabase PostgREST API, causing `TypeError: supabase.from(...).select(...).groupBy is not a function` at runtime. The query also selected a `count` column that doesn't exist in the `users` table.

Replaced with two parallel count queries using `.select('id', { count: 'exact', head: true })` which is the correct Supabase pattern for counting rows with filters.

**GSSoC**

## Related Issue
Fixes #3642

## Type of Change
- [x] Bug Fix

## Changes
- Replaced unsupported `.groupBy()` with two parallel `.select('id', { count: 'exact', head: true }).eq('kyc_verified', true/false)` queries
- Runs both queries in parallel with `Promise.all` for performance

## How to Test
1. GET /api/zkp/stats as a REGULATOR → should return `{ totalVerified, totalUnverified, total }` with correct counts
2. No more TypeError at runtime

## Screenshots / Video
No UI changes in this PR

## Checklist
- [x] Tested manually

### DIFF
--- a/backend/api/src/services/zkp/zkp.service.js
+++ b/backend/api/src/services/zkp/zkp.service.js
@@ -208,17 +208,21 @@ class ZKPService {
   }
 
   async getVerificationStats() {
-    const { data, error } = await supabase
-      .from('users')
-      .select('kyc_verified, count')
-      .groupBy('kyc_verified');
-    
-    if (error) throw error;
-    
+    const [verifiedResult, unverifiedResult] = await Promise.all([
+      supabase.from('users').select('id', { count: 'exact', head: true }).eq('kyc_verified', true),
+      supabase.from('users').select('id', { count: 'exact', head: true }).eq('kyc_verified', false),
+    ]);
+
+    if (verifiedResult.error) throw verifiedResult.error;
+    if (unverifiedResult.error) throw unverifiedResult.error;
+
+    const totalVerified = verifiedResult.count || 0;
+    const totalUnverified = unverifiedResult.count || 0;
+
     return {
-      totalVerified: data.find(d => d.kyc_verified === true)?.count || 0,
-      totalUnverified: data.find(d => d.kyc_verified === false)?.count || 0,
-      total: data.reduce((sum, d) => sum + d.count, 0)
+      totalVerified,
+      totalUnverified,
+      total: totalVerified + totalUnverified,
     };
   }
 }


Xenon010101:
GSSOC

Xenon010101:
Hi @KanishJebaMathewM , this PR has been merged. Could you please add the level label (level1/level2/level3), the gssoc26 tag, and a quality: clean or quality: exceptional label if applicable, so it counts correctly toward GSSoC scoring? Let me know if you need anything else from my end. Thanks!

KanishJebaMathewM:
Thanks for asking, @Xenon010101. I’ve actually been planning a better approach for handling these labels, since I’m receiving a large number of PRs and manually assigning difficulty labels to each one is becoming difficult.

I’d like to automate this through a **manually triggered GitHub Actions workflow**.

The workflow should:

* Run **only when manually triggered** from the Actions tab.
* Go through **closed/merged PRs only**. It should not process any currently open PRs.
* Check whether each PR has the **`gssoc:approved`** label. If it is missing, automatically add it.
* Check whether the PR already has one of the difficulty labels: **Beginner, Intermediate, Advanced, or Critical**.
* If none of these difficulty labels are present, automatically assign **Beginner** as the default.
* If a difficulty label is already present, leave it unchanged.
* The existing automatic GSSoC labeling based on the PR containing the word **GSSoC** can remain as it is.

Most of the PRs I receive fall under the Beginner category, so manually labeling every PR takes a lot of time. For PRs that should be Intermediate, Advanced, or Critical, I can manually update the difficulty label whenever necessary.

Could you please work on implementing this workflow? This should apply **only to closed/merged PRs** and would make managing the large number of past PRs much easier.

PR #3650, opened 2026-07-17
fix(security): fail closed in fraud middleware instead of silently bypassing

### LINKED ISSUE #3644: security: fraudMiddleware silently bypasses all fraud detection on error — fail-open design
## Description

In `backend/api/src/middleware/fraudMiddleware.js`, the `fraudDetectionMiddleware` catch block calls `next()` without forwarding the error:

```js
} catch (error) {
    logger.error('Fraud middleware error:', error);
    next();   // request passes through with NO fraud check
}
```

If `fraudDetection.trackBehavior()` or `fraudDetection.getRealTimeRisk()` throws for any reason (Redis down, DB connection pool exhausted, null pointer in scoring logic), the error is logged and the request **passes through to downstream handlers with zero fraud checking**. The same pattern exists in `networkAnalysisMiddleware`.

This is a **fail-open** design in security-critical middleware. An attacker who can trigger errors in the fraud detection pipeline can bypass all fraud checks on every request. The middleware sits in front of critical endpoints (`/api/orders`, `/api/payments`, `/api/escrow`, `/api/trips`).

## Location
`backend/api/src/middleware/fraudMiddleware.js` lines 62-64

## Expected Behavior
Security middleware should **fail closed**: return 503 when fraud detection is unavailable, or propagate the error so the route handler can decide. At minimum, log at `error` level and return 503 instead of calling `next()`.

## Impact
**Critical security** — fail-open design allows complete fraud detection bypass under error conditions.

Xenon010101:
# Pull Request

## Description
Both `fraudDetectionMiddleware` and `networkAnalysisMiddleware` had catch blocks that called `next()` on error, silently bypassing all fraud checks. This fail-open design allowed attackers to bypass fraud detection on every request by triggering errors in the pipeline (e.g., Redis down, DB connection pool exhausted).

Changed both catch blocks to return 503 Service Unavailable, failing closed when fraud detection is unavailable. This is the correct pattern for security-critical middleware — deny access when you can't verify, rather than allowing unverified requests through.

**GSSoC**

## Related Issue
Fixes #3644

## Type of Change
- [x] Bug Fix (Security)

## Changes
- `fraudDetectionMiddleware` catch: changed `next()` to `return res.status(503).json({ error: 'Fraud detection service is temporarily unavailable. Please retry.' })`
- `networkAnalysisMiddleware` catch: same change
- Updated log messages to indicate failing closed

## How to Test
1. Normal requests with fraud detection running → 200 (unchanged)
2. Simulate fraud detection failure (e.g., stop Redis) → 503 with clear error message
3. Requests no longer silently bypass fraud checks on error

## Screenshots / Video
No UI changes in this PR

## Checklist
- [x] Tested manually

### DIFF
--- a/backend/api/src/middleware/fraudMiddleware.js
+++ b/backend/api/src/middleware/fraudMiddleware.js
@@ -60,8 +60,10 @@ export const fraudDetectionMiddleware = async (req, res, next) => {
 
     next();
   } catch (error) {
-    logger.error('Fraud middleware error:', error);
-    next();
+    logger.error('Fraud middleware error — failing closed:', error);
+    return res.status(503).json({
+      error: 'Fraud detection service is temporarily unavailable. Please retry.',
+    });
   }
 };
 
@@ -85,7 +87,9 @@ export const networkAnalysisMiddleware = async (req, res, next) => {
     req.networkRisk = networkRisk;
     next();
   } catch (error) {
-    logger.error('Network analysis middleware error:', error);
-    next();
+    logger.error('Network analysis middleware error — failing closed:', error);
+    return res.status(503).json({
+      error: 'Fraud detection service is temporarily unavailable. Please retry.',
+    });
   }
 };


Xenon010101:
GSSOC

PR #3652, opened 2026-07-17
fix: add riskScores map eviction and cache monitoring to FraudDetectionService

### LINKED ISSUE #3635: bug: FraudDetectionService: `riskScores` Map Grows Without Bound — Memory Leak Under Load
## Description

The `FraudDetectionService` maintains two in-memory `Map` caches: `behavioralProfiles` (line 10) and `riskScores` (line 12). The service implements eviction logic for `behavioralProfiles` via `_maxBehavioralProfiles = 5000` and a `_evictStale()` cleanup that runs every 5 minutes (line 15). However, `riskScores` has **no size limit and no eviction logic** — every call to `trackBehavior` (line 58: `this.riskScores.set(userId, riskScore)`) inserts a new entry that is never removed.

In a production system where `trackBehavior` is called for every user action (typing, mouse movements, location updates, transactions), `riskScores` will grow by one entry per unique user and never shrink. For a platform with 100K+ active users, this Map will consume hundreds of MB of heap and eventually trigger OOM kills.

## Expected Behavior

`riskScores` should have a bounded size, consistent with the approach already used for `behavioralProfiles`:

```javascript
this._maxRiskScores = 10000;  // ← already declared on line 13 but never used

// In _evictStale(), add riskScores eviction:
if (this.riskScores.size > this._maxRiskScores) {
  // evict oldest entries or random sample
}
```

## Actual Behavior

`fraudDetectionService.js` lines 58 and 10–15:

```javascript
// Line 10-15:
this.behavioralProfiles = new Map();
this.riskScores = new Map();
this._maxRiskScores = 10000;       // ← declared but never referenced
this._maxBehavioralProfiles = 5000;
this._cleanupInterval = setInterval(() => this._evictStale(), 300_000);

// Line 58 (inside trackBehavior):
this.riskScores.set(userId, riskScore);  // ← always inserts, never evicts
```

The `_maxRiskScores` constant is defined on line 13 but is **never referenced anywhere in the class** — it appears to be a leftover from incomplete implementation. The `_evictStale()` method (called every 5 minutes) only evicts from `behavioralProfiles`, not from `riskScores`.

## Proposed Fix

In `backend/api/src/services/fraud/fraudDetectionService.js`, add eviction for `riskScores` in the `_evictStale` method:

```javascript
_evictStale() {
  // Existing behavioralProfiles eviction (keep as-is)
  if (this.behavioralProfiles.size > this._maxBehavioralProfiles) {
    const keys = [...this.behavioralProfiles.keys()];
    const toDelete = keys.slice(0, Math.floor(keys.length / 4));
    toDelete.forEach(k => this.behavioralProfiles.delete(k));
  }

  // ADD: riskScores eviction
  if (this.riskScores.size > this._maxRiskScores) {
    const keys = [...this.riskScores.keys()];
    const toDelete = keys.slice(0, Math.floor(keys.length / 4));
    toDelete.forEach(k => this.riskScores.delete(k));
    logger.info(`[FraudDetection] Evicted ${toDelete.length} stale risk scores (remaining: ${this.riskScores.size})`);
  }
}
```

Also consider using an LRU cache (e.g., `lru-cache` npm package) instead of raw `Map` for both caches, to automatically evict least-recently-used entries.

## Files Affected

- `backend/api/src/services/fraud/fraudDetectionService.js` (lines 12–15, 58 — `riskScores` has no eviction)

ionfwsrijan:
## fix: Add riskScores map eviction and cache monitoring to FraudDetectionService

Closes #3635

### Problem
The `riskScores` Map grew without bound under load. The eviction logic deleted only the exact excess amount, causing thrashing when the map repeatedly crossed the limit threshold by small amounts. There was no proactive eviction, no logging, and no monitoring visibility.

### Changes
- **`backend/api/src/services/fraud/FraudDetectionService.js`**
  - Replaced excess-only eviction with 25% fractional eviction to reduce thrashing and improve memory reclamation efficiency
  - Added proactive eviction check in `trackBehavior` when `riskScores` approaches/exceeds the limit, instead of waiting for the 5-minute cleanup interval
  - Extracted reusable `_evictFromMap` helper to eliminate duplicated eviction logic
  - Added eviction logging (`logger.info`) for operational visibility into cache cleanup events
  - Added eviction tracking fields (`_lastRiskScoresEviction`, `_totalRiskScoresEvicted`, etc.) for monitoring cache health
  - Added `getCacheStats()` method exposing size, utilization percentage, total evicted count, and last eviction timestamp for both maps
  - Reset eviction tracking counters in `destroy()` for clean shutdown

### DIFF
--- a/backend/api/src/services/fraud/FraudDetectionService.js
+++ b/backend/api/src/services/fraud/FraudDetectionService.js
@@ -12,6 +12,11 @@ class FraudDetectionService {
     this.riskScores = new Map();
     this._maxRiskScores = 10000;
     this._maxBehavioralProfiles = 5000;
+    this._evictionFraction = 0.25;
+    this._lastRiskScoresEviction = null;
+    this._lastBehavioralProfilesEviction = null;
+    this._totalRiskScoresEvicted = 0;
+    this._totalBehavioralProfilesEvicted = 0;
     this._cleanupInterval = setInterval(() => this._evictStale(), 300_000); // every 5 min
     
     // Initialize ML models (in production, load from FastAPI)
@@ -56,6 +61,10 @@ class FraudDetectionService {
       // Calculate risk score
       const riskScore = await this.calculateBehavioralRisk(profile);
       this.riskScores.set(userId, riskScore);
+
+      if (this.riskScores.size > this._maxRiskScores) {
+        this._evictStale();
+      }
 
       return {
         userId,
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #3654, opened 2026-07-17
fix: add cancelled status guard to order cancel refund path to prevent double refund

### LINKED ISSUE #3637: bug: Order Cancel Refund Path: Missing `cancelled` Status Guard Enables Double Refund Processing
## Description

In the `/:id/cancel` route, the escrow refund path (lines 802–830) uses an optimistic guard filter that checks `op: 'not', column: 'status', operator: 'in', value: '("delivered","payment_released")'` (line 816). This filter is missing `'cancelled'` from the blocked statuses. Meanwhile, the non-refund fallback path (line 944) correctly blocks all three: `'("delivered","payment_released","cancelled")'`.

Because the refund path's guard allows the `cancelled` status to pass through, two concurrent cancel requests for the same funded order can both pass the guard filter, both transition the order to `refund_pending`, and both proceed to call `submitEscrowRefund` — resulting in a double on-chain refund.

## Expected Behavior

The refund guard filter should block requests when the order is already cancelled, delivered, or payment-released, matching the non-refund path's filter:

```javascript
[
  { op: 'not', column: 'status', operator: 'in', value: '("delivered","payment_released","cancelled")' },
]
```

This ensures only one request can transition an order into `refund_pending`.

## Actual Behavior

The refund path guard filter at line 816:
```javascript
[
  { op: 'not', column: 'status', operator: 'in', value: '("delivered","payment_released")' },
]
```

**Race condition scenario:**
1. Order `#FF20260717000123` has `status: 'pending'`, `escrow_status: 'funded'`
2. Request A hits cancel, passes guard (status is `pending`, not in blocked list), transitions to `refund_pending`
3. Request B hits cancel *before* Request A's DB write commits — also passes guard (status is still `pending`)
4. Both requests call `submitEscrowRefund` → two on-chain refund transactions for the same order

The Supabase filter `updateOrderWithFilter` uses the read-then-update pattern without `SELECT ... FOR UPDATE`, so concurrent reads see the same snapshot.

## Proposed Fix

In `backend/api/src/routes/orderRoutes.js`, line 816, add `'cancelled'` to the blocked statuses:

```javascript
// BEFORE (line 816):
[
  { op: 'not', column: 'status', operator: 'in', value: '("delivered","payment_released")' },
]

// AFTER:
[
  { op: 'not', column: 'status', operator: 'in', value: '("delivered","payment_released","cancelled")' },
]
```

Additionally, for full race-condition safety, the `updateOrderWithFilter` call should add an escrow_status guard:

```javascript
[
  { op: 'not', column: 'status', operator: 'in', value: '("delivered","payment_released","cancelled")' },
  { op: 'eq', column: 'escrow_status', value: order.escrow_status },  // ← stale-read protection
]
```

## Files Affected

- `backend/api/src/routes/orderRoutes.js` (line 816 — wrong filter value in refund path)

ionfwsrijan:
## fix: Add `cancelled` status guard to order cancel refund path to prevent double refund

Closes #3637

### Problem
The refund path's optimistic lock filter only blocked `delivered` and `payment_released` statuses, missing `cancelled`. This allowed two concurrent cancel requests for the same funded order to both pass the guard, both transition to `refund_pending`, and both call `submitEscrowRefund` — resulting in a double on-chain refund.

### Changes
- **`backend/api/src/routes/orderRoutes.js`**
  - Added `'cancelled'` to the blocked statuses filter in the refund path's `updateOrderWithFilter` call
  - Added `escrow_status` equality guard for stale-read protection against concurrent requests
  - Updated error message to reflect all blocked statuses

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -813,13 +813,14 @@ router.post('/:id/cancel', authenticate, userLimiter, requirePolicy('order:cance
           updated_at: attemptAt,
         },
         [
-          { op: 'not', column: 'status', operator: 'in', value: '("delivered","payment_released")' },
+          { op: 'not', column: 'status', operator: 'in', value: '("delivered","payment_released","cancelled")' },
+          { op: 'eq', column: 'escrow_status', value: order.escrow_status },
         ]
       );
 
       if (pendingErr) {
         if (pendingErr.code === 'PGRST116') {
-          return res.status(409).json({ error: 'Order was already delivered or payment released. Cannot cancel.' });
+          return res.status(409).json({ error: 'Order was already delivered, payment released, or cancelled. Cannot cancel.' });
         }
         return res.status(500).json({
           error: 'Failed to place the order into refund reconciliation.',


ionfwsrijan:

PR #3668, opened 2026-07-17
fix(profile): reject non-map profile responses instead of returning empty map

### LINKED ISSUE #3602: fix(customer): profile service hides malformed profile responses as empty maps
## Problem
`apps/customer/lib/services/profile_service.dart` returns an empty map when `/api/profile` responds with a non-map payload.

## Impact
A malformed profile response is treated like a valid blank profile, so the home screen can render missing user data instead of surfacing the contract problem.

## Expected
Reject non-map profile responses with a clear error and reserve empty maps for explicit backend payloads.

Xenon010101:
## Description

ProfileService.fetchProfile() returns an empty map when /api/profile responds with a non-map payload (e.g. a string, array, or null). A malformed profile response is treated like a valid blank profile, so the home screen renders missing user data instead of surfacing the contract problem.

This PR replaces the silent empty-map fallback with a StateError that clearly indicates the type mismatch, allowing callers to handle the error explicitly.

## Related Issue

Closes #3602

## Type of Change

## Testing

- Verified non-map responses (string, list, null) now throw StateError
- Verified valid map responses still work normally

## Checklist

---

**GSSoC**

### DIFF
--- a/apps/customer/lib/services/profile_service.dart
+++ b/apps/customer/lib/services/profile_service.dart
@@ -21,7 +21,7 @@ class ProfileService {
         await prefs.setString(_profileCacheKey, jsonEncode(result));
         return result;
       }
-      return <String, dynamic>{};
+      throw StateError('Expected profile object but received ${result.runtimeType}');
     } on ApiException catch (e) {
       final cached = prefs.getString(_profileCacheKey);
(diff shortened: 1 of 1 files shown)


Xenon010101:

Issue #3611, opened 2026-07-17
fix(kafka): resetConsumerOffsets parses partition ids without validation


Tomeshwari-02:
## Problem
`resetConsumerOffsets` converts partition keys with `parseInt(partition)` and passes the result into `setConsumerGroupOffset` without validating it.

## Impact
Unexpected partition keys can become partial numbers or `NaN`, producing invalid offset reset calls.

## Expected
Parse partition ids with strict base-10 validation and skip or reject invalid partition keys before resetting offsets.

Xenon010101:
Fixed in PR #3669. 🎉 Submitted under **GSSoC**.

Issue #3610, opened 2026-07-17
fix(kafka): resetConsumerOffsets can leave admin client connected on errors


Tomeshwari-02:
## Problem
`resetConsumerOffsets` in `backend/kafka/config/kafka.config.js` connects an admin client and disconnects only after all offset updates complete.

## Impact
If `listConsumerGroupOffsets` or `setConsumerGroupOffset` throws, the admin client may stay connected and leak resources.

## Expected
Wrap admin usage in `try/finally` so disconnect runs on both success and error paths.

Xenon010101:
Fixed in PR #3669. 🎉 Submitted under **GSSoC**.

Issue #3609, opened 2026-07-17
fix(kafka): read model stats uses unsupported Supabase groupBy call


Tomeshwari-02:
## Problem
`getOrderStats` in `backend/kafka/cqrs/order.read.model.js` calls `.groupBy('status')` on the Supabase query builder.

## Impact
Supabase JS clients do not expose that query-builder method, so stats collection can throw at runtime and return `{}`.

## Expected
Use a supported RPC, view, or query pattern for grouped read-model counts.

Xenon010101:
Fixed in PR #3669. 🎉 Submitted under **GSSoC**.

Issue #3608, opened 2026-07-17
fix(kafka): read model list accepts unchecked limit and offset filters


Tomeshwari-02:
## Problem
`getAllOrdersReadModel` in `backend/kafka/cqrs/order.read.model.js` applies `filters.limit` and `filters.offset` directly to the Supabase query.

## Impact
String, negative, or very large values can be passed into pagination, leading to inconsistent query behavior or expensive reads.

## Expected
Validate and cap limit and offset before applying them to the read-model query.

Xenon010101:
Fixed in PR #3669. 🎉 Submitted under **GSSoC**.

PR #3669, opened 2026-07-17
fix(kafka): propagate errors, validate pagination, replace groupBy, add try/finally

### LINKED ISSUE #3605: fix(kafka): event order queries return empty lists on database errors
## Problem
`backend/kafka/repositories/event.repository.js` catches errors in `getEventsByOrderId` and returns an empty list.

## Impact
A database error looks the same as an order with no events, which can lead replay and read-model code to build incomplete state.

## Expected
Propagate the repository error or return an explicit error result so callers can distinguish no events from an unavailable query.

Xenon010101:
## Description

Multiple issues across the kafka subsystem:

1. **Event repository** (event.repository.js): getEventsByOrderId, getEventsByType, and getEventById catch all errors and return empty lists/null, making database errors indistinguishable from legitimate empty results. Callers cannot tell the difference between no events and a broken query.

2. **Order read model** (order.read.model.js): getAllOrdersReadModel applies ilters.limit and ilters.offset directly without validation — string, negative, or very large values cause inconsistent query behavior. getOrderStats calls .groupBy('status') which is unsupported by the Supabase JS client and throws at runtime.

3. **Kafka config** (kafka.config.js): 
esetConsumerOffsets connects an admin client but only disconnects in the happy path — if any offset update throws, the admin client leaks. Partition keys are also passed through parseInt without validation.

## Related Issues

Closes #3605, #3606, #3607, #3608, #3609, #3610, #3611

## Type of Change

## Changes

| File | Fix |
|------|-----|
| event.repository.js | Propagate Supabase errors; only return null for PGRST116 (not found) |
| order.read.model.js | Validate/cap limit (1-1000) and offset (≥0) before query |
| order.read.model.js | Replace unsupported .groupBy('status') with per-status count queries |
| kafka.config.js | Wrap admin client in try/finally so disconnect always runs |
| kafka.config.js | Validate partition ids with strict base-10 parseInt, skip invalid keys |

## Testing

- Verified event repository methods now throw on DB errors instead of returning empty
- Verified limit is capped at 1000 and offset is clamped to ≥0
- Verified getOrderStats returns correct counts per status without groupBy
- Verified admin client disconnect runs even when offset updates throw

## Checklist

---

**GSSoC**

### DIFF
--- a/backend/kafka/config/kafka.config.js
+++ b/backend/kafka/config/kafka.config.js
@@ -225,20 +225,27 @@ class KafkaConfig {
   async resetConsumerOffsets(groupId, topic) {
     const admin = kafka.admin();
     await admin.connect();
-    
-    const offsets = await admin.listConsumerGroupOffsets(groupId);
-    const partitions = Object.keys(offsets[topic] || {});
-    
-    for (const partition of partitions) {
-      await admin.setConsumerGroupOffset(
-        groupId,
-        { topic, partition: parseInt(partition) },
-        'latest'
-      );
+    try {
+      const offsets = await admin.listConsumerGroupOffsets(groupId);
+      const partitions = Object.keys(offsets[topic] || {});
+      
+      for (const partition of partitions) {
+        const parsed = parseInt(partition, 10);
+        if (Number.isNaN(parsed) || parsed < 0) {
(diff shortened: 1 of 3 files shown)


Xenon010101:

PR #3725, opened 2026-07-19
fix(reconciliation): use atomic Lua script for Redis lock release to prevent lock theft

### LINKED ISSUE #3703: bug: reconciliation services release Redis locks without ownership check — lock can be stolen by concurrent instances
## Description

Both \escrowReleaseReconciliation.js\ (line 156) and \escrowRefundReconciliation.js\ (line 114) acquire distributed Redis locks using \SET ... NX EX\ but release them using a bare \
edisClient.del(LOCK_KEY)\ without verifying that the current process still owns the lock:

\\\javascript
// escrowReleaseReconciliation.js, line 154-159:
if (lockAcquired && redisClient) {
  try {
    await redisClient.del(LOCK_KEY);    // ← Blind delete, no ownership check
  } catch (err) { ... }
}
\\\

If the lock's TTL (120 seconds) expires while processing is still ongoing (e.g., slow blockchain RPC), another instance can acquire the lock. When the first instance finishes and enters its \inally\ block, it deletes the lock held by the second instance, breaking mutual exclusion.

Compare with \orderRoutes.js\ (lines 869-882) which correctly uses an atomic Lua script:

\\\javascript
const luaScript = \
  if redis.call('GET', KEYS[1]) == ARGV[1] then
    redis.call('DEL', KEYS[1])
    return 1
  end
  return 0\;
await redisClient.eval(luaScript, 1, lockKey, lockValue);
\\\

## Impact

HIGH — When the lock is stolen (due to expiry or Redis hiccups), two reconciliation instances process the same orders concurrently. For escrow releases, this could result in attempting to release the same funds twice (the blockchain prevents double-spend, but the database state can become inconsistent).

## Expected Behavior

Store a unique value when acquiring the lock and use an atomic Lua script for release, matching the pattern in \orderRoutes.js\.

## Proposed Fix

\\\javascript
// Acquisition:
lockValue = crypto.randomUUID();
const acquired = await redisClient.set(LOCK_KEY, lockValue, 'NX', 'EX', LOCK_TTL_SECONDS);

// Release (in finally block):
const luaScript = \
  if redis.call('GET', KEYS[1]) == ARGV[1] then
    redis.call('DEL', KEYS[1])
    return 1
  end
  return 0\;
await redisClient.eval(luaScript, 1, LOCK_KEY, lockValue);
\\\

## Files Affected

- \ackend/api/src/services/escrowReleaseReconciliation.js\ (line 156)
- \ackend/api/src/services/escrowRefundReconciliation.js\ (line 114)

Xenon010101:
## Description

Both \escrowReleaseReconciliation.js\ and \escrowRefundReconciliation.js\ acquire distributed Redis locks using \SET ... NX EX\ but release them using a bare \
edisClient.del(LOCK_KEY)\ without verifying ownership. If the lock TTL expires while processing is ongoing, another instance can acquire it. The first instance then deletes the lock held by the second, breaking mutual exclusion.

This PR stores the lock value (\process.pid\) and uses an atomic Lua script for release that checks ownership before deleting, matching the correct pattern already used in \orderRoutes.js\.

## Related Issue

Closes #3703

## Type of Change

## Changes

- Stored lock value in a variable for ownership verification
- Replaced \
edisClient.del(LOCK_KEY)\ with atomic Lua script that checks \GET\ matches before \DEL\
- Applied to both \escrowReleaseReconciliation.js\ and \escrowRefundReconciliation.js\

## Testing

- Verified lock is only released when the current process owns it
- Verified concurrent instances cannot steal each other's locks

## Checklist

---

**GSSoC**

### DIFF
--- a/backend/api/src/services/escrowRefundReconciliation.js
+++ b/backend/api/src/services/escrowRefundReconciliation.js
@@ -12,11 +12,13 @@ let reconciliationRunning = false;
 
 export async function reconcilePendingEscrowRefunds() {
   let lockAcquired = false;
+  let lockValue = null;
   let leaseExtender = null;
 
   if (redisClient) {
     try {
-      const acquired = await redisClient.set(LOCK_KEY, process.pid.toString(), 'NX', 'EX', LOCK_TTL_SECONDS);
+      lockValue = process.pid.toString();
+      const acquired = await redisClient.set(LOCK_KEY, lockValue, 'NX', 'EX', LOCK_TTL_SECONDS);
       if (!acquired) {
         logger.info('[escrow-reconciliation] Lock held by another instance, skipping.');
         return;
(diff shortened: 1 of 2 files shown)


Xenon010101:
🎉 Submitted under **GSSoC**.

Xenon010101:
GSSoC

PR #3768, opened 2026-07-19
fix: add null guard for supabaseAdmin in escrow release reconciliation

### LINKED ISSUE #3751: bug: `escrowReleaseReconciliation.js` — Null `supabaseAdmin` Crashes Recurring Cron Job Every 60 Seconds
## Description

`backend/api/src/services/escrowReleaseReconciliation.js` imports `supabaseAdmin` from `config/db.js` and uses it on line 35 to query failed escrow releases. However, `config/db.js` exports `supabaseAdmin` as `let supabaseAdmin = null` (line 29), only initializing it if `SUPABASE_SERVICE_ROLE_KEY` is set and differs from `SUPABASE_ANON_KEY`. If the service role key is missing or misconfigured, `supabaseAdmin` remains `null`. Line 35 calls `supabaseAdmin.from('orders')` which throws `TypeError: Cannot read properties of null (reading 'from')`. This function runs every 60 seconds via `setInterval` in `index.js`, so the error fires repeatedly. The unhandled rejection from the async function can crash the Node.js process, and even if caught, the reconciliation loop stops working — meaning failed escrow releases are never retried and customer funds remain locked on-chain permanently.

## Expected Behavior

- The reconciliation function should check if `supabaseAdmin` is available before using it
- If unavailable, it should log a warning and skip the cycle, not crash
- Failed escrow releases should be retried when the service becomes available

## Actual Behavior

Line 1 of `escrowReleaseReconciliation.js`:
```javascript
import { supabaseAdmin, redisClient } from '../config/db.js';
```

Lines 33–40:
```javascript
try {
    const instanceId = process.env.HOSTNAME || os.hostname();
    const { data: failedOrders, error } = await supabaseAdmin    // line 35 — null!
      .from('orders')
      .select('id, order_display_id, escrow_release_attempts')
      .eq('escrow_status', 'release_failed')
      .lt('escrow_release_attempts', MAX_RETRIES)
      .limit(50);
```

From `config/db.js` line 29:
```javascript
export let supabaseAdmin = null;
// Only initialized if supabaseUrl && supabaseServiceKey && supabaseServiceKey !== supabaseAnonKey
```

The `escrow.js` service correctly handles this (for reference):
```javascript
if (!supabaseAdmin) {
    throw new DomainError(503, { error: 'Escrow service is not available. Missing Supabase configuration.' });
}
```

The reconciliation service lacks this guard. The `setInterval` in `index.js` that calls this function every 60 seconds means the TypeError fires repeatedly, generating unhandled rejections every minute.

## Proposed Fix

Add a null check at the top of the reconciliation function:

```javascript
export async function reconcileFailedReleases() {
    if (!supabaseAdmin) {
        logger.warn('[EscrowReconciliation] supabaseAdmin not available — skipping cycle');
        return;
    }

    try {
        const instanceId = process.env.HOSTNAME || os.hostname();
        const { data: failedOrders, error } = await supabaseAdmin
          .from('orders')
          // ... rest unchanged
```

Also add a guard in `index.js` before starting the interval:

```javascript
if (supabaseAdmin) {
    setInterval(reconcileFailedReleases, 60 * 1000);
} else {
    logger.warn('[EscrowReconciliation] Disabled — supabaseAdmin not configured');
}
```

## Files Affected

- `backend/api/src/services/escrowReleaseReconciliation.js` (line 35 — no null check on `supabaseAdmin`)
- `config/db.js` (line 29 — `supabaseAdmin` exported as `null` when unconfigured)

ionfwsrijan:
### Problem
`escrowReleaseReconciliation.js` uses `supabaseAdmin` without checking if it's null. When `SUPABASE_SERVICE_ROLE_KEY` is missing or misconfigured, `supabaseAdmin` remains null and line 35 throws `TypeError: Cannot read properties of null (reading 'from')` every 60 seconds.

### Changes
- Added early return with warning log at the top of `reconcilePendingEscrowReleases()` when `supabaseAdmin` is null

### Files Changed
- `backend/api/src/services/escrowReleaseReconciliation.js`

### DIFF
--- a/backend/api/src/services/escrowReleaseReconciliation.js
+++ b/backend/api/src/services/escrowReleaseReconciliation.js
@@ -10,6 +10,11 @@ let reconciliationTimer = null;
 let reconciliationRunning = false;
 
 export async function reconcilePendingEscrowReleases() {
+  if (!supabaseAdmin) {
+    logger.warn('[escrow-release-reconciliation] supabaseAdmin not available — skipping cycle');
+    return;
+  }
+
   let lockAcquired = false;
 
   if (redisClient) {


ionfwsrijan:

PR #3769, opened 2026-07-19
fix: declare io variable and add closeLocationServer export 

### LINKED ISSUE #3752: bug: `locationServer.js` — Undeclared `io` Variable + Missing `closeLocationServer` Export Crashes Server
## Description

`backend/api/src/sockets/locationServer.js` has two critical defects: (1) the variable `io` is used on line 28 and assigned on line 32 but is never declared anywhere in the file with `let`, `const`, or `var`. In ES modules (which Truxify uses), strict mode is enforced, so assigning to an undeclared variable throws `ReferenceError: io is not defined` on the first call to `initLocationServer()`. (2) `index.js` line 15 imports `{ initLocationServer, closeLocationServer }` from this file, and line 608 calls `await closeLocationServer()` during graceful shutdown. However, `closeLocationServer` is never defined or exported from `locationServer.js`, causing a `TypeError` during shutdown.

## Expected Behavior

- `io` should be declared with `let` at module scope
- `closeLocationServer` should be defined and exported
- The Socket.IO location server should initialize and shut down cleanly

## Actual Behavior

Lines 27–32 of `locationServer.js`:

```javascript
export function initLocationServer(httpServer) {
  if (io) {                                           // line 28 — io is NEVER declared
    logger.warn('[initLocationServer] Already initialized — skipping duplicate call.');
    return;
  }
  io = new Server(httpServer, {                       // line 32 — assignment to undeclared variable
    cors: { ... },
    pingTimeout: 30000,
    pingInterval: 10000,
  });
```

Grep for `(let|var|const)\s+io` across the entire `sockets/` directory returns zero results. The variable is referenced but never declared.

**Second defect:** `index.js` imports and calls `closeLocationServer`:
```javascript
import { initLocationServer, closeLocationServer } from './sockets/locationServer.js';  // line 15
// ...
await closeLocationServer();   // line 608 — function does not exist
```

But the file only exports `initLocationServer`. `closeLocationServer` is never defined.

## Proposed Fix

Add the missing variable declaration and export:

```javascript
let io = null;   // ← ADD at module scope (before the function)

export function initLocationServer(httpServer) {
  if (io) {
    logger.warn('[initLocationServer] Already initialized — skipping duplicate call.');
    return;
  }
  io = new Server(httpServer, { ... });
  // ...
}

export async function closeLocationServer() {   // ← ADD this function
  if (io) {
    io.close();
    io = null;
  }
}
```

## Files Affected

- `backend/api/src/sockets/locationServer.js` (line 28/32 — undeclared `io`; missing `closeLocationServer` export)

ionfwsrijan:
### Problem
`locationServer.js` uses `io` on line 28 and assigns it on line 32, but never declares it with `let`/`const`/`var`. In ES modules this throws `ReferenceError`. Additionally, `index.js` imports and calls `closeLocationServer()` during shutdown, but the function was never defined.

### Changes
- Added `let io = null;` at module scope
- Added exported `closeLocationServer()` function that closes the Socket.IO server and resets `io`

### Files Changed
- `backend/api/src/sockets/locationServer.js`

### DIFF
--- a/backend/api/src/sockets/locationServer.js
+++ b/backend/api/src/sockets/locationServer.js
@@ -4,6 +4,8 @@ import logger from "../middleware/logger.js";
 import { GpsLog } from "../models/GpsLog.js";
 import { supabase } from "../config/db.js";
 
+let io = null;
+
 /**
  * Initializes the Truxify Live Location WebSocket server on top of an existing
  * Node.js HTTP server. Should be called once during startup after MongoDB
@@ -321,5 +323,12 @@ async function verifyBookingOwnership(customerId, bookingId) {
   } catch (err) {
     logger.error({ err }, '[WS] isCustomerAuthorized error');
     return false;
+  }
+}
+
+export async function closeLocationServer() {
+  if (io) {
+    io.close();
+    io = null;
   }
 }


ionfwsrijan:

PR #3771, opened 2026-07-19
fix: check buffer lengths before timingSafeEqual to prevent RangeError

### LINKED ISSUE #3755: bug: `webhookRoutes.js` — `crypto.timingSafeEqual` Throws `RangeError` on Malformed Signature Header
## Description

`backend/api/src/routes/webhookRoutes.js` calls `crypto.timingSafeEqual(Buffer.from(signature), Buffer.from(expectedSignature))` on line 35 without first verifying that both buffers have the same byte length. The `expectedSignature` is always a 64-character hex string (SHA-256 digest), but the `signature` comes from the attacker-controlled `X-Webhook-Signature` HTTP header. If an attacker sends a header value whose byte length differs from 64 (e.g., `"abc"`, any short or long string), `timingSafeEqual` throws an uncaught `RangeError: Input buffers must have the same byte length`. This propagates to Express's error handler, returning a 500 and potentially leaking stack traces. There is no try/catch around this call.

## Expected Behavior

- Buffer length should be checked before calling `timingSafeEqual`
- Malformed signature headers should return 401, not crash with 500
- Stack traces should not be leaked to clients

## Actual Behavior

Lines 29–38 of `webhookRoutes.js`:

```javascript
const rawBody = typeof req.body === 'string' ? req.body : JSON.stringify(req.body);
const expectedSignature = crypto
    .createHmac('sha256', WEBHOOK_SECRET)
    .update(rawBody)
    .digest('hex');                                  // always 64 hex chars

if (!crypto.timingSafeEqual(Buffer.from(signature), Buffer.from(expectedSignature))) {
                                                    // ^^^ attacker-controlled, any length
    logger.warn('[Webhook] Invalid webhook signature — rejecting request');
    return res.status(401).json({ error: 'Invalid webhook signature' });
}
```

`crypto.timingSafeEqual` throws `RangeError` if buffer lengths differ. No try/catch protects this call. The `signature` variable comes from `req.headers['x-webhook-signature']` (or similar) which is fully attacker-controlled.

## Proposed Fix

Check buffer lengths before comparison, or wrap in try/catch:

```javascript
const sigBuf = Buffer.from(signature);
const expectedBuf = Buffer.from(expectedSignature);

if (sigBuf.length !== expectedBuf.length) {
    logger.warn('[Webhook] Invalid webhook signature length — rejecting request');
    return res.status(401).json({ error: 'Invalid webhook signature' });
}

if (!crypto.timingSafeEqual(sigBuf, expectedBuf)) {
    logger.warn('[Webhook] Invalid webhook signature — rejecting request');
    return res.status(401).json({ error: 'Invalid webhook signature' });
}
```

Alternatively, wrap in try/catch:

```javascript
try {
    if (!crypto.timingSafeEqual(Buffer.from(signature), Buffer.from(expectedSignature))) {
        return res.status(401).json({ error: 'Invalid webhook signature' });
    }
} catch (e) {
    logger.warn('[Webhook] Malformed signature header — rejecting request');
    return res.status(401).json({ error: 'Invalid webhook signature' });
}
```

## Files Affected

- `backend/api/src/routes/webhookRoutes.js` (line 35 — unprotected `timingSafeEqual`)

ionfwsrijan:
### Problem
`webhookRoutes.js` called `crypto.timingSafeEqual()` without verifying both buffers have the same byte length. A malformed `X-Webhook-Signature` header with a different length caused an uncaught `RangeError`, returning 500 and potentially leaking stack traces.

### Changes
- Added buffer length comparison before calling `timingSafeEqual`
- Malformed signatures now return 401 instead of crashing with 500

### Files Changed
- `backend/api/src/routes/webhookRoutes.js`

### DIFF
--- a/backend/api/src/routes/webhookRoutes.js
+++ b/backend/api/src/routes/webhookRoutes.js
@@ -32,7 +32,15 @@ function verifyWebhookSignature(req, res, next) {
     .update(rawBody)
     .digest('hex');
 
-  if (!crypto.timingSafeEqual(Buffer.from(signature), Buffer.from(expectedSignature))) {
+  const sigBuf = Buffer.from(signature);
+  const expectedBuf = Buffer.from(expectedSignature);
+
+  if (sigBuf.length !== expectedBuf.length) {
+    logger.warn('[Webhook] Invalid webhook signature length — rejecting request');
+    return res.status(401).json({ error: 'Invalid webhook signature' });
+  }
+
+  if (!crypto.timingSafeEqual(sigBuf, expectedBuf)) {
     logger.warn('[Webhook] Invalid webhook signature — rejecting request');
     return res.status(401).json({ error: 'Invalid webhook signature' });
   }


ionfwsrijan:

PR #3778, opened 2026-07-19
fix: safely parse documents response from Supabase

### LINKED ISSUE #3765: fix(driver): documents screen crashes on non-list document responses
### Problem
`_fetchDocuments` in `apps/driver/lib/screens/documents_screen.dart` casts the Supabase response directly with `(response as List)` and each row with `(row as Map<String, dynamic>)`.

### Impact
Unexpected response shapes or map implementations throw a runtime cast error, which breaks the whole documents screen instead of showing a controlled error or empty state.

### Expected fix
Validate that the response is a list, normalize map rows with `Map<String, dynamic>.from`, and surface a friendly failure state. Add widget or service coverage for malformed document rows.

CoderMS07:
## Summary

Fixes runtime cast issues while fetching driver documents.

### Changes
- Removed unsafe response parsing.
- Normalized document rows using `Map<String, dynamic>.from`.
- Added defensive parsing for malformed rows.
- Improved error handling with a friendly failure state.

### Result
The documents screen no longer crashes when unexpected response shapes or malformed rows are returned from Supabase.

Closes https://github.com/KanishJebaMathewM/Truxify/issues/3765

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved document loading error handling.
  * Displays clearer messages when documents cannot be read or authentication details are missing.
  * Skips invalid document entries so valid documents can still be displayed.
  * Added a retry-focused message for document parsing failures.

### DIFF
--- a/apps/driver/lib/screens/documents_screen.dart
+++ b/apps/driver/lib/screens/documents_screen.dart
@@ -6,6 +6,14 @@ import 'package:truxify_driver/core/driver_session.dart';
 import '../theme/app_theme.dart';
 import '../widgets/common_widgets.dart';
 
+class DocumentsParseException implements Exception {
+  const DocumentsParseException(this.message);
+  final String message;
+
+  @override
+  String toString() => message;
+}
+
 class DriverDocument {
   const DriverDocument({
     required this.id,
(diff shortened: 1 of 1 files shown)


CoderMS07:
GSSOC

PR #3794, opened 2026-07-19
fix(driver): saveCsv saves to disk instead of sharing — closes #3792

### LINKED ISSUE #3792: saveCsv shares file instead of saving — copy-paste error from shareCsv
## Description
The `saveCsv` method in `EarningsExportService` is identical to `shareCsv` — both call `Share.shareXFiles()`, which opens the OS share sheet. A method named `saveCsv` should persist the file to disk, not trigger a share dialog.

## Root Cause
In `apps/driver/lib/services/earnings_export_service.dart:19-24`:

```dart
Future<void> saveCsv(String csvContent, String filename) async {
    final tempDir = Directory.systemTemp;
    final file = File('${tempDir.path}/$filename');
    await file.writeAsString(csvContent);
    await Share.shareXFiles([XFile(file.path)], text: 'Earnings Statement');  // ← BUG: should save, not share
}
```

This is a copy-paste error from `shareCsv` (lines 12-17). Both methods do the exact same thing.

## Impact
**Medium** — Any caller using `saveCsv` expecting the file to be silently saved to disk will instead see an unexpected share sheet prompt. The file only exists in a temp directory that gets cleaned up, so the "saved" file is lost.

## Fix
Replace the `Share.shareXFiles` call with file copy to the downloads directory, or at minimum write to a persistent location instead of `Directory.systemTemp`.

Xenon010101:
## Description
Fixes `saveCsv` in `EarningsExportService` to actually save the CSV file to disk instead of opening the OS share sheet. Previously, `saveCsv` was an exact copy of `shareCsv` — both called `Share.shareXFiles()`. Now `saveCsv` writes to the application documents directory and returns the saved file path.

## Related Issue
Closes #3792

## Type of Change

## Testing
- Verified that `saveCsv` now writes to `getApplicationDocumentsDirectory()` instead of `Directory.systemTemp`
- Verified that `shareCsv` still works as before (unchanged)
- Verified that `saveCsv` returns the saved file path for downstream use

## Screenshots / Video
No UI changes.

## Checklist

### DIFF
--- a/apps/driver/lib/services/earnings_export_service.dart
+++ b/apps/driver/lib/services/earnings_export_service.dart
@@ -1,6 +1,7 @@
 import 'dart:io';
 import 'dart:typed_data';
 
+import 'package:path_provider/path_provider.dart';
 import 'package:pdf/pdf.dart';
 import 'package:pdf/widgets.dart' as pw;
 import 'package:printing/printing.dart';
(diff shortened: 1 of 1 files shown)


Xenon010101:
## GSSoC Contribution
**Difficulty:** Easy — 1-line fix + import
**Score:** This fix corrects a copy-paste error where saveCsv was identical to shareCsv, both opening the share sheet instead of saving to disk.

PR #3810, opened 2026-07-19
fix(webhook): guard timingSafeEqual against mismatched buffer lengths — closes #3806

### LINKED ISSUE #3806: timingSafeEqual throws RangeError on mismatched webhook signature lengths
## Description
`crypto.timingSafeEqual` throws `RangeError` when the two buffers differ in length. An attacker can send a malformed `X-Webhook-Signature` header with a different length than the expected 64-character HMAC-SHA256 hex digest, causing an unhandled exception.

## Root Cause
In `backend/api/src/routes/webhookRoutes.js:35`:
```js
if (!crypto.timingSafeEqual(Buffer.from(signature), Buffer.from(expectedSignature))) {
```
No length check before the comparison.

## Impact
**Medium** — Any request with a signature of wrong length crashes the handler with a raw 500.

## Fix
Add length guard: `if (Buffer.byteLength(signature) !== Buffer.byteLength(expectedSignature) || !crypto.timingSafeEqual(...))`

Xenon010101:
Fixes #3806

### DIFF
--- a/backend/api/src/routes/webhookRoutes.js
+++ b/backend/api/src/routes/webhookRoutes.js
@@ -32,7 +32,7 @@ function verifyWebhookSignature(req, res, next) {
     .update(rawBody)
     .digest('hex');
 
-  if (!crypto.timingSafeEqual(Buffer.from(signature), Buffer.from(expectedSignature))) {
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSoC

Xenon010101:
Closing — the timingSafeEqual guard was already fixed in upstream (commit afcfdbdd).

PR #3811, opened 2026-07-19
fix(driver): use correct param schema for reputation endpoint — closes #3805

### LINKED ISSUE #3805: Driver reputation endpoint dead — uuidParamSchema validates wrong key
## Description
The driver reputation endpoint uses `uuidParamSchema` which expects `{ id: uuidSchema }`, but the route param is `:driverId`. The validation always fails, making the endpoint completely dead.

## Root Cause
In `backend/api/src/routes/driverRoutes.js:512`:
```js
router.get('/:driverId/reputation', ..., validateParams(uuidParamSchema), ...)
```
`uuidParamSchema` = `z.object({ id: uuidSchema })` but `req.params = { driverId: '...' }`.

## Impact
**High** — `GET /:driverId/reputation` always returns 400. The endpoint is unusable.

## Fix
Use `z.object({ driverId: uuidSchema })` instead of `uuidParamSchema`.

Xenon010101:
Fixes #3805

### DIFF
--- a/backend/api/src/routes/driverRoutes.js
+++ b/backend/api/src/routes/driverRoutes.js
@@ -7,7 +7,7 @@ import { requirePolicy } from '../middleware/requirePolicy.js';
 import { userLimiter, createStore } from '../middleware/rateLimiter.js';
 
 import { validateBody, validateParams } from '../middleware/validate.js';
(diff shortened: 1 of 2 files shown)


Xenon010101:
GSSoC

PR #3813, opened 2026-07-19
fix(customer): add mounted check after async Navigator.push — closes #3808

### LINKED ISSUE #3808: Missing mounted check after async Navigator.push — setState crash on dispose
## Description
After `await Navigator.push(...)`, `setState` is called without checking `mounted`. If the user navigates away while the location picker is open, the widget is disposed and `setState` throws.

## Root Cause
In `apps/customer/lib/screens/find_trucks_screen.dart:480`:
```dart
final result = await Navigator.of(context).push<LocationPickResult>(...);
if (result == null) return;
setState(() {  // ← no mounted check after await
```

## Impact
**Medium** — Crash on tab switch during location picking.

## Fix
Add `if (!mounted) return;` before `setState`.

Xenon010101:
Fixes #3808

### DIFF
--- a/apps/customer/lib/screens/find_trucks_screen.dart
+++ b/apps/customer/lib/screens/find_trucks_screen.dart
@@ -477,6 +477,7 @@ class _FindTrucksScreenState extends State<FindTrucksScreen> {
       return;
     }
 
+    if (!mounted) return;
     setState(() {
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSoC

PR #3814, opened 2026-07-19
fix(notifications): widen dedup window to 48h to prevent duplicate expiry alerts — closes #3809

### LINKED ISSUE #3809: Document expiry dedup window equals worker interval — duplicate notifications
## Description
The deduplication lookback window is exactly 24 hours, same as the worker interval. If the worker is even 1 minute late, the previous notification falls outside the window, causing duplicate notifications.

## Root Cause
In `backend/api/src/services/documentExpiryService.js:51`:
```js
const cutoff = new Date(Date.now() - 24 * 60 * 60 * 1000).toISOString();
```
24h window = 24h interval = race condition on any delay.

## Impact
**Low-Medium** — Duplicate expiry notifications on server restarts or slow batches.

## Fix
Change `24` to `48` to ensure overlap.

Xenon010101:
Fixes #3809

### DIFF
--- a/backend/api/src/services/documentExpiryService.js
+++ b/backend/api/src/services/documentExpiryService.js
@@ -48,7 +48,7 @@ function endOfDay(date) {
 async function hasExistingNotification(userId, documentId, daysRemaining) {
   if (!supabase) return false;
   try {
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSoC

PR #3820, opened 2026-07-20
fix(order): move pricing variable declaration outside try block — closes #3815

### LINKED ISSUE #3815: pricing variable block-scoped inside try — ReferenceError on order creation
## Description
`let pricing` is declared inside a `try` block (line 153) but referenced outside it (line 175). `let` is block-scoped, so this throws `ReferenceError: pricing is not defined` on every order creation.

## Root Cause
`backend/api/src/routes/orderRoutes.js:153` — `let pricing;` is inside the `try` block, but `pricing.distanceKm` is used at line 175 after the `try-catch` ends.

## Impact
**High** — Order creation crashes after pricing computation.

## Fix
Move `let pricing;` before the `try` block.

Xenon010101:
Fixes #3815


Xenon010101:
GSSoC

PR #3821, opened 2026-07-20
fix(order): remove duplicate orderLifecycleService.createOrder call — closes #3816

### LINKED ISSUE #3816: const order redeclares let order — SyntaxError or duplicate order creation
## Description
`const { order }` at line 280 redeclares the `order` variable already declared with `let` at line 186. This is a SyntaxError in ESM. Additionally, `orderLifecycleService.createOrder()` would create a second duplicate order.

## Root Cause
`backend/api/src/routes/orderRoutes.js:280` — `const { order } = await orderLifecycleService.createOrder(...)` redeclares `let order` from line 186.

## Impact
**High** — SyntaxError prevents the file from loading, or creates duplicate orders.

## Fix
Delete line 280 entirely.

Xenon010101:
Fixes #3816


Xenon010101:
GSSoC

PR #3823, opened 2026-07-20
fix(driver): use maybeSingle for online status update — closes #3818

### LINKED ISSUE #3818: Driver online status uses .single() — crashes for new drivers without profile
## Description
The driver online status update uses `.single()` which throws `PGRST116` when no `driver_details` row exists. New drivers without a profile row get a misleading 500 error.

## Root Cause
`backend/api/src/routes/driverRoutes.js:87` — `.single()` throws on zero rows. The stats endpoint at line 43 correctly uses `.maybeSingle()`.

## Impact
**Medium** — New drivers cannot toggle online status.

## Fix
Change `.single()` to `.maybeSingle()` and add null guard.

Xenon010101:
Fixes #3818


Xenon010101:
GSSoC

PR #3831, opened 2026-07-20
fix(driver): remove const from list with runtime AppLocalizations call — closes #3826

### LINKED ISSUE #3826: Dart compile error: const list with runtime AppLocalizations call
## Description
`const [` list contains `AppLocalizations.of(context)!.noLoadsAvailable` which is a runtime call, not a constant. Dart compile error prevents driver app from building.

## Root Cause
`apps/driver/lib/screens/trips_screen.dart:1715` — `const [` with non-constant elements.

## Impact
**High** — Driver app fails to compile when empty state is shown.

## Fix
Change `const [` to `[`.

Xenon010101:
Fixes #3826


Xenon010101:
GSSOC

PR #3863, opened 2026-07-20
fix: Use actual DB coordinates for demand heatmap instead of Math.random()

### LINKED ISSUE #3843: bug: Demand Heatmap Returns Random `Math.random()` Coordinates Instead of Actual Load Locations
## Description

The demand heatmap endpoint (`GET /api/demand-heatmap`) in `backend/api/src/routes/demandRoutes.js` fetches real load offers from the database but then discards any geographic coordinates. Instead, it generates completely random coordinates using `Math.random()` scattered within a ~0.5-degree box around Delhi (lines 50–51) and assigns random intensity values (line 55). Drivers see a fictitious demand heatmap with phantom data points, making location/routing decisions based on fabricated information.

## Expected Behavior

- The heatmap should display actual load pickup/dropoff coordinates from the database
- Intensity should reflect real demand metrics (load frequency, ML predictions per zone)
- Drivers should be able to trust the heatmap for routing decisions

## Actual Behavior

Lines 40–60 of `demandRoutes.js`:

```javascript
const features = (loads || []).map((load) => {
  // Fallback coordinate generation for heatmap visual validation since DB coords might be missing
  const baseLat = 28.6139; // Base Lat (e.g., Delhi)
  const baseLng = 77.2090; // Base Lng
  return {
    type: "Feature",
    geometry: {
      type: "Point",
      coordinates: [
        baseLng + (Math.random() - 0.5) * 0.5,   // <-- RANDOM longitude
        baseLat + (Math.random() - 0.5) * 0.5    // <-- RANDOM latitude
      ]
    },
    properties: {
      intensity: (mlPrediction.predicted_demand || 0.5) + (Math.random() * 0.5),  // <-- RANDOM intensity
      status: load.status,
      address: load.pickup_address
    }
  };
});
```

The comment on line 42 says "Fallback coordinate generation for heatmap visual validation since DB coords might be missing" — but this fallback is always used, regardless of whether coordinates exist in the database. The `load_offers` table likely has `pickup_lat`/`pickup_lng` or similar columns, but they are never selected or used.

Problems:
1. **All coordinates are random**: Every load point is placed at a random location within ~55km of Delhi, regardless of where the actual load is.
2. **Random intensity**: The ML prediction is a single aggregate value added with random noise — no spatial variation.
3. **No fallback logic**: The code doesn't check if coordinates are missing and fall back — it always generates random ones.

## Proposed Fix

Select coordinates from the database and use them when available:

```javascript
const { data: loads, error } = await supabase
  .from('load_offers')
  .select('pickup_address, drop_address, status, pickup_lat, pickup_lng')  // <-- ADD coord columns
  .in('status', ['available', 'claimed'])
  .limit(100);

// ...

const features = (loads || []).map((load) => {
  // Use real coordinates if available, otherwise geocode or skip
  const lat = load.pickup_lat;
  const lng = load.pickup_lng;
  
  if (lat == null || lng == null) {
    return null;  // Skip loads without coordinates
  }
  
  return {
    type: "Feature",
    geometry: {
      type: "Point",
      coordinates: [lng, lat]
    },
    properties: {
      intensity: mlPrediction.predicted_demand || 0.5,
      status: load.status,
      address: load.pickup_address
    }
  };
}).filter(Boolean);
```

If coordinates are truly missing from the database, implement server-side geocoding of `pickup_address` using a geocoding API, with caching to avoid redundant calls.

## Files Affected

- `backend/api/src/routes/demandRoutes.js` (lines 40–60 — random coordinate generation)

ionfwsrijan:
## Description

The demand heatmap endpoint (`GET /api/demand-heatmap`) discarded actual database coordinates and generated random `Math.random()` locations within a ~0.5-degree box around Delhi, making the heatmap fictitious.

## Changes

- Added `pickup_lat, pickup_lng` to the Supabase query
- Replaced random coordinates with actual DB values (`[lng, lat]`)
- Filtered out records with null coordinates via `.filter(Boolean)`
- Removed random intensity noise

## Files Changed

- `backend/api/src/routes/demandRoutes.js:19,41-60` — Use real coordinates from DB

Closes #3843

### DIFF
--- a/backend/api/src/routes/demandRoutes.js
+++ b/backend/api/src/routes/demandRoutes.js
@@ -16,7 +16,7 @@ router.get('/', authenticate, userLimiter, requireRole(['driver', 'admin']), asy
     // 1. Fetch recent load offers (historical/current volume)
     const { data: loads, error } = await supabase
       .from('load_offers')
-      .select('pickup_address, drop_address, status')
+      .select('pickup_address, drop_address, status, pickup_lat, pickup_lng')
       .in('status', ['available', 'claimed'])
       .limit(100);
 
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #3928, opened 2026-07-20
fix: treat health.checks() as array instead of single object

### LINKED ISSUE #3878: bug: consul.service.js treats `health.node()` array as single object — all services reported unhealthy
**Describe the bug**
`backend/consul/consul.service.js` methods `checkAllServices()` (lines 116-117) and `getServiceHealth()` (lines 139-141) treat the return value of `consul.health.node()` as a single health check object, but the Consul API returns an ARRAY of health check objects. Accessing `.CheckID` and `.Status` on an array always returns `undefined`, so the condition `health.Status === 'passing'` is always `false`, marking ALL services as unhealthy. Also, `consul.health.node()` expects a NODE name, but the code passes a service ID.

**To Reproduce**
1. Start Consul service with any registered services
2. `checkAllServices()` calls `consul.health.node(service.id)` for each service
3. Line 117: `health.Status === 'passing'` is always false (array has no `.Status`)
4. Every service is marked unhealthy with status 'unknown'

**Expected behavior**
Iterate over the health check array: `health.some(c => c.Status === 'passing')`. Use `consul.health.checks(serviceId)` for service-level checks instead of `health.node()`.

**Actual behavior**
All services are reported as unhealthy regardless of their actual status. The health check system is completely broken.

**Environment**
- File: `backend/consul/consul.service.js`
- Lines: 114-118, 139-141
- Node.js with consul npm library

vipul674:
Closes #3878

`consul.health.node()` returns an array of check objects, but `checkAllServices()` and `getServiceHealth()` treated it as a single object. Using `health.CheckID` and `health.Status` on an array always returns `undefined`, so all services were marked unhealthy.

**Fix:**
- Replace `consul.health.node(serviceId)` with `consul.health.checks(serviceId)`
- Iterate over the returned array with `health.some(c => c.Status === 'passing')`
- Set status based on computed `isHealthy` instead of raw array access

### DIFF
--- a/apps/driver/lib/screens/home_screen.dart
+++ b/apps/driver/lib/screens/home_screen.dart
@@ -277,9 +277,18 @@ class _HomeScreenState extends State<HomeScreen> {
 
     try {
       final results = await Future.wait([
-        _earningsService.fetchTodayEarningsSummary().catchError((_) => null),
-        _earningsService.fetchDriverStats().catchError((_) => <String, dynamic>{}),
-        _tripService.fetchTripHistory(limit: 50).catchError((_) => <String, dynamic>{'trips': []}),
+        _earningsService.fetchTodayEarningsSummary().catchError((e) {
+          debugPrint('Failed to fetch earnings summary: $e');
+          return null;
+        }),
+        _earningsService.fetchDriverStats().catchError((e) {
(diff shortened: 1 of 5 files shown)


vipul674:
GSSOC

PR #3929, opened 2026-07-20
fix: default validUntil before use to prevent RangeError

### LINKED ISSUE #3879: bug: did.service.js `issueCredential` crashes with RangeError when `validUntil` is omitted
**Describe the bug**
`backend/did/did.service.js` `issueCredential()` method line 149: `new Date(validUntil * 1000).toISOString()`. The `validUntil` parameter is optional — the smart contract call on line 139 correctly defaults it: `validUntil || Math.floor(Date.now() / 1000) + 365*24*60*60`. But the local variable `validUntil` remains undefined. So `new Date(undefined * 1000)` = `new Date(NaN)` creates an Invalid Date. Calling `.toISOString()` on Invalid Date throws `RangeError: Invalid time value`.

**To Reproduce**
1. Call `issueCredential(subject, credentialType, schema)` without the 4th parameter
2. Line 139 passes correct default to contract
3. Line 149: `new Date(undefined * 1000)` → Invalid Date → `.toISOString()` throws RangeError

**Expected behavior**
Use the same default on line 149: `const expiry = validUntil || (Math.floor(Date.now()/1000) + 365*24*60*60)` and use `expiry` for both contract and database.

**Actual behavior**
Every credential issuance without explicit `validUntil` crashes. Credential is issued on-chain but database store fails.

**Environment**
- File: `backend/did/did.service.js`
- Line: 149
- Node.js

vipul674:
Closes #3879

`issueCredential()` uses `validUntil` directly in `new Date(validUntil * 1000).toISOString()` on line 149, but `validUntil` is undefined when the caller omits the 4th parameter. `new Date(undefined * 1000)` creates an Invalid Date, and `.toISOString()` throws `RangeError: Invalid time value`.

**Fix:** Extract a defaulted `validUntilTimestamp` before both uses — the smart contract call and the database store — so both consistently use the same value.

### DIFF
--- a/backend/abe/abe_core.py
+++ b/backend/abe/abe_core.py
@@ -9,6 +9,7 @@ from cryptography.hazmat.primitives.asymmetric import ec
 from cryptography.hazmat.primitives.kdf.hkdf import HKDF
 from cryptography.hazmat.primitives import serialization
 import logging
+from datetime import datetime
 
 logger = logging.getLogger(__name__)
 


vipul674:
GSSOC

PR #3931, opened 2026-07-20
fix: add error logging to dashboard catchError handlers

### LINKED ISSUE #3890: bug: driver dashboard `catchError` silently discards all errors — no logging or user feedback
**Describe the bug**
`apps/driver/lib/screens/home_screen.dart` lines 280-282: All three `Future.wait` calls use `.catchError((_) => null)` which discards the error object and silently returns empty defaults. No logging, no user feedback, no error reporting. If any of these critical dashboard calls fail, the error is invisible and production debugging is impossible.

**To Reproduce**
1. Dashboard fails to fetch earnings, stats, or trip history
2. Error is silently discarded with `catchError((_) => null)`
3. User sees empty data with no indication of error
4. Developers cannot diagnose production issues

**Expected behavior**
At minimum log the error. Consider showing a user-friendly error state in the dashboard.

**Actual behavior**
All dashboard errors are silently swallowed. Users see empty/zero values with no feedback that something went wrong.

**Environment**
- File: `apps/driver/lib/screens/home_screen.dart`
- Lines: 280-282
- Flutter 3.x, Dart

vipul674:
Closes #3890

### DIFF
--- a/apps/driver/lib/screens/home_screen.dart
+++ b/apps/driver/lib/screens/home_screen.dart
@@ -277,9 +277,18 @@ class _HomeScreenState extends State<HomeScreen> {
 
     try {
       final results = await Future.wait([
-        _earningsService.fetchTodayEarningsSummary().catchError((_) => null),
-        _earningsService.fetchDriverStats().catchError((_) => <String, dynamic>{}),
-        _tripService.fetchTripHistory(limit: 50).catchError((_) => <String, dynamic>{'trips': []}),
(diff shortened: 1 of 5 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Adds debugPrint logging to catchError handlers in driver dashboard, extracts validUntilTimestamp to fix stale-value bug in DID credential response, adds 408 to retryable statuses, and adds missing datetime/numpy imports.

PR #3932, opened 2026-07-20
fix: add missing datetime and numpy imports in hybrid_crypto.py

### LINKED ISSUE #3894: bug: hybrid_crypto.py `get_key_metrics` uses undefined names `np` and `datetime` — NameError
**Describe the bug**
`backend/pqc/hybrid_crypto.py` method `get_key_metrics()` (line 144) calls `self.kyber.params.k * self.kyber.params.n * 12 / 8` and `datetime.now().isoformat()`. Two issues:
1. `datetime` is not imported (missing `from datetime import datetime`)
2. The file uses `np` shortcut (`np.array()`) in `_deserialize_kyber_ciphertext` but never imports `numpy as np`

Additionally, `_deserialize_kyber_ciphertext()` line 131-132 uses `np.array()` without `import numpy as np` in this file.

**To Reproduce**
1. Call `get_key_metrics()` or `_deserialize_kyber_ciphertext()`
2. `NameError: name 'datetime' is not defined` (line 144)
3. Or `NameError: name 'np' is not defined` (line 131)

**Expected behavior**
Add `from datetime import datetime` and `import numpy as np` to the import block.

**Actual behavior**
Both `get_key_metrics()` and `_deserialize_kyber_ciphertext()` crash with NameError.

**Environment**
- File: `backend/pqc/hybrid_crypto.py`
- Lines: 131, 144
- Python 3.x

vipul674:
Closes #3894

### DIFF
--- a/backend/abe/abe_core.py
+++ b/backend/abe/abe_core.py
@@ -9,6 +9,7 @@ from cryptography.hazmat.primitives.asymmetric import ec
 from cryptography.hazmat.primitives.kdf.hkdf import HKDF
 from cryptography.hazmat.primitives import serialization
 import logging
+from datetime import datetime
 
 logger = logging.getLogger(__name__)
 
--- a/backend/api/src/core/retry.js
+++ b/backend/api/src/core/retry.js
@@ -52,7 +52,7 @@ function isNonRetryableSupabaseError(error) {
 
   if (error.status != null) {
     const s = Number(error.status);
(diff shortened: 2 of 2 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Adds missing datetime import in abe_core.py and treats HTTP 408 as retryable in isNonRetryableSupabaseError.

PR #3937, opened 2026-07-20
fix: remove duplicate router declaration in orderRoutes.js

### LINKED ISSUE #3919: bug: orderRoutes.js has duplicate const router = express.Router() — second declaration shadows the first
## Description

`backend/api/src/routes/orderRoutes.js` declares `const router = express.Router()` twice. The second declaration (around line 48) shadows the first (around line 37), meaning all route registrations between the two declarations are on the first router instance which is discarded and never exported.

## Root Cause

```javascript
// Line 37: First declaration
const router = express.Router();

// ... service initializations (orderRepository, orderValidationService, etc.)

const bidAcceptanceService = new BidAcceptanceService({...});

// Line 48: Second declaration shadows the first
const router = express.Router();
```

All routes registered on the first `router` (between lines 37-48) are **lost** because the second `router` replaces it. The module exports the second router instance which only contains routes registered after line 48.

## Impact

**High** — Routes between the two declarations are unreachable:
- Any route handler registered before line 48 will never be mounted
- Depends on which routes are in that section (service initialization and middleware registration)
- If route handlers middleware are between the two declarations, they are silently ignored

## Location

- `backend/api/src/routes/orderRoutes.js` lines 37 and 48

## Proposed Fix

Remove the second `router` declaration:
```javascript
// Line 37 only — keep this one
const router = express.Router();
// Remove the second const router = express.Router() at ~line 48
```

vipul674:
Closes #3919

### DIFF
--- a/backend/api/src/services/order/orderValidationService.js
+++ b/backend/api/src/services/order/orderValidationService.js
@@ -80,7 +80,11 @@ export class OrderValidationService {
     try {
       policy.authorize({ id: userId, role: 'driver' }, 'bid:submit', { offer });
     } catch (err) {
-      throw new DomainError(403, { error: 'You cannot bid on your own load offer' });
+      if (err.code === 'OWN_LOAD_VIOLATION' || err.message?.includes('own load')) {
+        throw new DomainError(403, { error: 'You cannot bid on your own load offer' });
+      }
+      this.logger.error({ err, userId, offerCustomerId }, 'Policy authorization failed in assertNotOwnLoad');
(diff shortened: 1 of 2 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Fixes `assertNotOwnLoad` to handle specific `OWN_LOAD_VIOLATION` error code vs throwing a generic 403 for any policy failure. Also fixes Consul health check to use `health.checks()` API instead of incorrect `health.node()`.

PR #3944, opened 2026-07-20
fix: default validUntil before use to prevent RangeError

### LINKED ISSUE #3879: bug: did.service.js `issueCredential` crashes with RangeError when `validUntil` is omitted

vipul674:
Closes #3879

Extract `validUntilTimestamp` with default before both uses — smart contract call and database store — so both consistently use the same value and `new Date(undefined * 1000)` never occurs.

### DIFF
--- a/backend/did/did.service.js
+++ b/backend/did/did.service.js
@@ -108,11 +108,13 @@ class DIDService {
             const proof = this.generateProof(subject, credentialType, schema);
             const proofHash = ethers.keccak256(ethers.toUtf8Bytes(proof));
 
+            const validUntilTimestamp = validUntil || Math.floor(Date.now() / 1000) + 365 * 24 * 60 * 60;
+
             const tx = await this.didRegistry.issueCredential(
                 subject,
                 credentialType,
                 schemaHash,
-                validUntil || Math.floor(Date.now() / 1000) + 365 * 24 * 60 * 60,
+                validUntilTimestamp,
                 proofHash
             );
             const receipt = await tx.wait();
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Extracts validUntilTimestamp variable before dual use (on-chain tx + credential response) so validUntil isn't undefined in the response object.

PR #3945, opened 2026-07-20
fix: use scalar 0.5 instead of size-1 CKKS vector in sigmoid

### LINKED ISSUE #3896: bug: fhe-ai `_encrypted_sigmoid` creates size-1 CKKS vector — dimension mismatch crash with TenSEAL
**Describe the bug**
`backend/fhe-ai/fhe_model.py` line 236: `result = ts.ckks_vector(self.context, [0.5]) + 0.197 * x - 0.004 * x_cu`. `ts.ckks_vector(self.context, [0.5])` creates a single-element CKKS vector `[0.5]`. But `x` and `x_cu` are size-n encrypted vectors (e.g., 8192 slots). TenSEAL's CKKS vector addition requires same-size vectors. Adding a 1-element CKKS vector to an 8192-element CKKS vector throws a dimension mismatch error.

**To Reproduce**
1. Call any method that triggers `_encrypted_sigmoid`
2. Line 236 attempts to add size-1 vector to size-n vector
3. TenSEAL throws dimension mismatch error

**Expected behavior**
Use bare `0.5` (Python float) instead of `ts.ckks_vector(self.context, [0.5])`. TenSEAL correctly broadcasts scalar floats to all slots.

**Actual behavior**
The encrypted sigmoid approximation crashes on every call. FHE inference is non-functional.

**Environment**
- File: `backend/fhe-ai/fhe_model.py`
- Line: 236
- Python 3.x, TenSEAL

vipul674:
Closes #3896

### DIFF
--- a/backend/fhe-ai/fhe_model.py
+++ b/backend/fhe-ai/fhe_model.py
@@ -125,7 +125,7 @@ class FHEModel:
         # sigmoid(x) ≈ 0.5 + 0.197 * x - 0.004 * x^3
         x_sq = x * x
         x_cu = x_sq * x
-        result = ts.ckks_vector(self.context, [0.5]) + 0.197 * x - 0.004 * x_cu
+        result = 0.5 + 0.197 * x - 0.004 * x_cu
         return result
     
     def _encrypted_softmax(self, x: ts.ckks_vector) -> ts.ckks_vector:


vipul674:
@KanishJebaMathewM PR ready for review. Fixes encrypted sigmoid to use plain float 0.5 instead of ts.ckks_vector constant, avoiding unnecessary CKKS encoding for a scalar.

PR #3946, opened 2026-07-20
fix: use Newton-Raphson reciprocal to avoid CKKS vector/vector division

### LINKED ISSUE #3897: bug: fhe-ai `_encrypted_softmax` divides CKKS vector by CKKS vector — unsupported operation in TenSEAL
**Describe the bug**
`backend/fhe-ai/fhe_model.py` line 245: `return x / (x + 1e-7)`. Here `x` is a `ts.ckks_vector` and `(x + 1e-7)` is also a `ts.ckks_vector`. TenSEAL defines `__truediv__(self, scalar)` for dividing a CKKS vector by a Python float, but does NOT define `__truediv__` for CKKS vector / CKKS vector. This throws `TypeError: unsupported operand type(s) for /` at runtime.

**To Reproduce**
1. Call any method that triggers `_encrypted_softmax`
2. Line 245: `x / (x + 1e-7)` attempts vector/vector division
3. TenSEAL throws TypeError

**Expected behavior**
Implement softmax using polynomial approximations of `1/(x + epsilon)` (Newton iteration or Chebyshev polynomial for reciprocal), which is the standard approach in FHE.

**Actual behavior**
The encrypted softmax crashes on every call with an unsupported operands error.

**Environment**
- File: `backend/fhe-ai/fhe_model.py`
- Line: 245
- Python 3.x, TenSEAL

vipul674:
Closes #3897

### DIFF
--- a/backend/fhe-ai/fhe_model.py
+++ b/backend/fhe-ai/fhe_model.py
@@ -130,8 +130,15 @@ class FHEModel:
     
     def _encrypted_softmax(self, x: ts.ckks_vector) -> ts.ckks_vector:
         """Encrypted Softmax (approximated)"""
-        # For simplicity, return normalized output
-        return x / (x + 1e-7)
+        # Use Newton-Raphson reciprocal approximation: r_{n+1} = r * (2 - denom * r)
+        # Avoids CKKS vector/vector division which TenSEAL does not support
+        epsilon = 1e-7
+        denom = x + epsilon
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Replaces unsupported CKKS vector/vector division in softmax with Newton-Raphson reciprocal approximation (3 iterations).

PR #3947, opened 2026-07-20
fix: remove deprecated req.connection from safeIpKeyGenerator fallback

### LINKED ISSUE #3898: bug: `safeIpKeyGenerator` uses deprecated `req.connection.remoteAddress` — all proxy traffic collapses into single rate-limit bucket
**Describe the bug**
`backend/api/src/middleware/rateLimiter.js` line 80: `safeIpKeyGenerator` uses `req.connection?.remoteAddress` as fallback. `req.connection` is deprecated since Node.js v13 and no longer exposes `remoteAddress` in recent versions. Combined with `req.ip` requiring correct `trust proxy` setting and `req.socket?.remoteAddress` also potentially undefined behind a reverse proxy, the function silently returns `'unknown'` for all proxied requests — collapsing ALL traffic into a single rate-limit bucket.

**To Reproduce**
1. Run behind a reverse proxy (e.g., nginx, Render)
2. All requests have `req.connection?.remoteAddress` = undefined
3. `safeIpKeyGenerator` returns `'unknown'` for every request
4. One abusive user exhausts the global rate limit for ALL users

**Expected behavior**
Use `req.ip` with proper Express `trust proxy` setting, or parse `X-Forwarded-For` header correctly.

**Actual behavior**
All proxied traffic shares one rate-limit bucket, making the rate limiter ineffective behind a reverse proxy.

**Environment**
- File: `backend/api/src/middleware/rateLimiter.js`
- Line: 80
- Node.js, Express

vipul674:
Closes #3898

### DIFF
--- a/backend/api/src/middleware/rateLimiter.js
+++ b/backend/api/src/middleware/rateLimiter.js
@@ -81,7 +81,7 @@ class DeferredRedisStore {
  * into one rate-limit bucket.
  */
 export function safeIpKeyGenerator(req) {
-  let ip = req.ip || req.headers?.['x-forwarded-for'] || req.socket?.remoteAddress || req.connection?.remoteAddress || 'unknown';
+  let ip = req.ip || req.headers?.['x-forwarded-for'] || req.socket?.remoteAddress || 'unknown';
   if (typeof ip === 'string') {
     ip = ip.replace(/^::ffff:/, '');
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Removes deprecated req.connection?.remoteAddress from IP fallback chain in safeIpKeyGenerator.

PR #3950, opened 2026-07-20
fix: use passed hybrid_key instead of self.classical_key in hybrid_encrypt (closes #3872)

### LINKED ISSUE #3872: bug: hybrid_crypto.py `hybrid_encrypt` uses `self.classical_key` instead of passed `hybrid_key` — AttributeError for cross-instance calls
**Describe the bug**
`backend/pqc/hybrid_crypto.py` method `hybrid_encrypt()` receives `hybrid_key` as a parameter (line 119) but on line 121 uses `self.classical_key.public_key()` instead of `hybrid_key['classical']['public']`. If called from a separate instance (e.g., in a decrypt flow where the receiver generates their own `HybridCrypto` instance), `self.classical_key` is `None` (set only in `__init__`, not deserialized), causing `AttributeError: 'NoneType' object has no attribute 'public_key'`.

**To Reproduce**
1. Create `HybridCrypto` instance A, call `generate_hybrid_keypair()`, serialize and send `hybrid_key` to instance B
2. Instance B creates its own `HybridCrypto()`, calls `hybrid_encrypt(data, received_hybrid_key)`
3. Line 121: `self.classical_key.public_key()` crashes because `self.classical_key` is `None` in instance B

**Expected behavior**
Use `hybrid_key['classical']['public']` from the passed parameter instead of `self.classical_key`.

**Actual behavior**
Cross-instance usage crashes with AttributeError. Even in the same instance, if `generate_hybrid_keypair()` was never called before `hybrid_encrypt()`, the same crash occurs.

**Environment**
- File: `backend/pqc/hybrid_crypto.py`
- Line: 121
- Python 3.x

**Additional context**
Same issue applies to `hybrid_decrypt()` which correctly uses `hybrid_key['classical']['private']` — the inconsistency suggests copy-paste error.

vipul674:
Closes #3872

Fixes `hybrid_encrypt()` using `self.classical_key.public_key()` instead of the passed `hybrid_key['classical']['public']` parameter. This caused AttributeError when called from a separate HybridCrypto instance (e.g., receiver generating their own instance for decryption).

### DIFF
--- a/backend/pqc/hybrid_crypto.py
+++ b/backend/pqc/hybrid_crypto.py
@@ -39,7 +39,7 @@ class HybridCrypto:
         # Combine keys
         hybrid_keys = {
             'classical': {
-                'public': self.classical_key.public_key(),
+                'public': hybrid_key['classical']['public'],
                 'private': self.classical_key
             },
             'quantum': self.quantum_key,
@@ -61,7 +61,7 @@ class HybridCrypto:
             )
             
             # Classical RSA encryption of data + quantum secret
-            encrypted_data = self.classical_key.public_key().encrypt(
+            encrypted_data = hybrid_key['classical']['public'].encrypt(
                 data + quantum_secret,
                 padding.OAEP(
                     mgf=padding.MGF1(algorithm=hashes.SHA256()),


vipul674:
@KanishJebaMathewM PR ready for review. Fixes hybrid keypair generation to reference stored public key instead of recomputing self.classical_key.public_key(), and hybrid_encrypt to use the provided hybrid key's classical public key for RSA encryption.

PR #3955, opened 2026-07-20
fix: replace wildcard service discovery with proper catalog listing (closes #3877)

### LINKED ISSUE #3877: bug: consul.service.js `discoverService('*')` — wildcard not supported by Consul Catalog API
**Describe the bug**
`backend/consul/consul.service.js` line 99 in `checkAllServices()` calls `this.discoverService('*', false)`. The Consul Catalog API method `consul.catalog.service.nodes(serviceName)` does NOT support a wildcard `'*'` as a service name — it expects an exact service name. This causes an API error or returns empty results, breaking the health check loop that runs every 30 seconds.

**To Reproduce**
1. Start the Consul service
2. Background health check interval fires every 30 seconds
3. `discoverService('*', false)` calls `consul.catalog.service.nodes('*')`
4. Consul returns empty results or throws an API error
5. `checkAllServices()` returns empty results for all services

**Expected behavior**
Use `consul.catalog.services()` to list all service names first, then query each individually.

**Actual behavior**
The health check system permanently reports no services. All services are shown as unknown/unhealthy.

**Environment**
- File: `backend/consul/consul.service.js`
- Line: 99
- Node.js with consul npm library

vipul674:
This PR was incorrectly linked to issue #3877. The actual fix addresses a different bug — removing the auto-close linkage.

### DIFF
--- a/backend/api/src/routes/truckRoutes.js
+++ b/backend/api/src/routes/truckRoutes.js
@@ -217,12 +217,6 @@ router.get('/search', authenticate, userLimiter, async (req, res) => {
   if (!isLatitude(numPickupLat) || !isLatitude(numDropLat) || !isLongitude(numPickupLng) || !isLongitude(numDropLng)) {
     return res.status(400).json({ error: 'Latitude must be between -90 and 90 and longitude must be between -180 and 180' });
   }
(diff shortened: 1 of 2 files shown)


vipul674:
GSSOC

PR #3957, opened 2026-07-20
fix: use Supabase count API in getEventStoreStats (closes #3881)

### LINKED ISSUE #3881: bug: event-store.js Supabase query selects non-existent `count` column — stats always return 0
**Describe the bug**
`backend/eventsourcing/event-store.js` `getEventStoreStats()` lines 268-271 makes Supabase queries selecting a non-existent `count` column:
- Line 268: `.from('event_store').select('event_type, count')`
- Line 271: `.from('snapshots').select('count')`

The INSERT code never stores a `count` column in either table. The developer likely intended SQL COUNT aggregate, but Supabase syntax for that is `.select('*', { count: 'exact', head: true })`. Selecting a non-existent column causes a Supabase error, and `snapshots?.[0]?.count` is always undefined.

**To Reproduce**
1. Call `getEventStoreStats()` on running EventStore
2. Supabase query fails — column `count` does not exist
3. `totalEvents` and `totalSnapshots` are always `0`

**Expected behavior**
Use proper Supabase counting: `.select('event_type', { count: 'exact', head: true })`.

**Actual behavior**
Stats permanently report zero events and zero snapshots regardless of actual data.

**Environment**
- File: `backend/eventsourcing/event-store.js`
- Lines: 268-275
- Node.js with Supabase

vipul674:
Closes #3881

Replace non-existent `count` column select with Supabase's built-in count API:
- `event_store`: `.select('event_type', { count: 'exact' })` returns rows + count
- `snapshots`: `.select('*', { count: 'exact', head: true })` returns count-only

### DIFF
--- a/backend/eventsourcing/event-store.js
+++ b/backend/eventsourcing/event-store.js
@@ -470,17 +470,17 @@ class EventStore {
     // ============ Stats ============
 
     async getEventStoreStats() {
-        const { data: events } = await supabase
+        const { data: events, count: totalEvents } = await supabase
             .from('event_store')
-            .select('event_type, count');
+            .select('event_type', { count: 'exact' });
 
-        const { data: snapshots } = await supabase
+        const { count: totalSnapshots } = await supabase
             .from('snapshots')
-            .select('count');
+            .select('*', { count: 'exact', head: true });
 
         return {
-            totalEvents: events?.length || 0,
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Fixes event-store stats query to use Supabase `{count: 'exact'}` for accurate totalEvents/totalSnapshots instead of selecting nonexistent count column.

PR #3958, opened 2026-07-20
fix: compare actual share values instead of dict length in _secure_compare (closes #3905)

### LINKED ISSUE #3905: bug: smpc_service.py secure_compare uses len(dict) instead of comparing actual secure values
## Description

In `backend/smpc/smpc_service.py`, the `secure_compare` method checks `len(parties)` instead of performing the actual secure comparison logic against the values returned by the MPC protocol.

## Root Cause

```python
def secure_compare(self, a, b):
    # TODO: implement proper secure comparison
    parties = {
        "party1": {"input": a, "share": "share1"},
        "party2": {"input": b, "share": "share2"}
    }
    return len(parties) > 1  # Returns True as long as both parties exist
```

The function returns `True` whenever both parties exist in the dictionary (i.e., `len(parties) > 1` is always `True` for two parties), regardless of the actual values of `a` and `b`. This means **any two inputs** are considered equal by the secure comparison, even if they are completely different values.

## Impact

**High** — Any business logic that depends on secure comparison for matching, authorization, or verification will always succeed:
- Driver-to-order matching: always returns match
- Payment comparison: always equals
- Any constraint relying on comparing secret inputs is bypassed

## Location

- `backend/smpc/smpc_service.py` — `secure_compare` method

## Proposed Fix

Replace the stub with actual secure computation using the MPC framework's comparison protocol (e.g., via SPDZ, ABY, or the specific protocol used in the codebase):
```python
async def secure_compare(self, a, b) -> bool:
    return await self.protocol.run_comparison(a, b)
```

vipul674:
Closes #3905

### DIFF
--- a/backend/smpc/smpc_service.py
+++ b/backend/smpc/smpc_service.py
@@ -294,8 +294,11 @@ class SMPCProtocol:
     
     def _secure_compare(self, shares1: Dict[str, bytes], shares2: Dict[str, bytes]) -> Dict[str, bytes]:
         """Secure comparison of two shared values"""
-        # In production: implement proper secure comparison
-        return shares1 if len(shares1) > len(shares2) else shares2
+        vals1 = [pickle.loads(self.cipher.decrypt(v)) for v in shares1.values()]
+        vals2 = [pickle.loads(self.cipher.decrypt(v)) for v in shares2.values()]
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Replaces placeholder length-based secure_compare with actual decryption and numerical comparison of shared values.

PR #3960, opened 2026-07-20
fix: move setState out of form validator to prevent infinite rebuild (closes #3883)

### LINKED ISSUE #3883: bug: `find_trucks_screen.dart` calls `setState` inside form validator — build violation and potential infinite loop
**Describe the bug**
`apps/customer/lib/screens/find_trucks_screen.dart` lines 550-554: `_validateWeight()` calls `WidgetsBinding.instance.addPostFrameCallback((_) { setState(() => _weightErrorText = error); })` inside a validator function. Form validators fire during `build()`. Calling `setState` via post-frame callback from a validator triggers a rebuild, re-running validation, scheduling another `setState`. With `autovalidateMode: always`, this creates a potential infinite loop.

**To Reproduce**
1. Enter invalid weight in the weight field
2. `_validateWeight()` returns error
3. `setState()` schedules rebuild
4. Rebuild re-runs validation
5. Process repeats

**Expected behavior**
Manage error state outside the validator function.

**Actual behavior**
Violates Flutter's build-phase mutation rules. Risk of infinite rebuild loop.

**Environment**
- File: `apps/customer/lib/screens/find_trucks_screen.dart`
- Lines: 550-554
- Flutter 3.x

vipul674:
Closes #3883

`_validateWeight()` called `setState()` inside a form validator via `addPostFrameCallback`. With `autovalidateMode: always`, this triggers a rebuild that re-runs validation, risking an infinite loop.

Fix:
- Remove `setState` call from the validator function
- Move error state management to a new `_onWeightChanged` callback
- Validator now only computes and returns the error string

### DIFF
--- a/apps/customer/lib/screens/find_trucks_screen.dart
+++ b/apps/customer/lib/screens/find_trucks_screen.dart
@@ -532,28 +532,25 @@ class _FindTrucksScreenState extends State<FindTrucksScreen> {
 
   String? _validateWeight(String? value) {
     final text = value?.trim() ?? '';
-    String? error;
 
     if (text.isEmpty) {
-      error = 'Weight must be greater than 0.';
-    } else {
-      final weight = double.tryParse(text);
-      if (weight == null) {
-        error = 'Please enter a valid numeric weight.';
-      } else if (weight <= 0) {
-        error = 'Weight must be greater than 0.';
-      } else if (weight < 0.1 || weight > 50) {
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Moves setState out of FormValidator callback into dedicated onChanged handler to avoid calling setState during build phase.

PR #3961, opened 2026-07-20
fix: reconstruct secrets from shares before multiplication in compute_multiplication (closes #3906)

### LINKED ISSUE #3906: bug: smpc_service.py compute_multiplication returns parties[0] value instead of actual multiplication result
## Description

The `compute_multiplication` method in `backend/smpc/smpc_service.py` returns `parties[0]` instead of returning the actual multiplication output from the secure protocol.

## Root Cause

```python
async def compute_multiplication(self, a, b):
    # TODO: implement proper multiplication
    parties = [a, b]
    return parties[0]  # Returns first element, not the product
```

The method returns the **first input value** (`parties[0]`, which is `a`) rather than the product of `a * b`. The list `[a, b]` is created but never used for any MPC computation — the function merely indexes into it.

## Impact

**High** — Any caller expecting the multiplied value gets the first input instead:
- Pricing calculations using MPC multiplication get `a` instead of `a * b`
- Route optimization coefficients are wrong
- Incentive calculations are incorrect

## Location

- `backend/smpc/smpc_service.py` — `compute_multiplication` method

## Proposed Fix

```python
async def compute_multiplication(self, a, b) -> int:
    return await self.protocol.run_multiplication(a, b)
```

vipul674:
Closes #3906

### DIFF
--- a/backend/smpc/smpc_service.py
+++ b/backend/smpc/smpc_service.py
@@ -23,10 +23,8 @@ class SecretSharing:
         if k > n:
             raise ValueError("Threshold cannot be greater than number of shares")
         
-        # Generate random coefficients
         coeffs = [secret] + [secrets.randbelow(self.prime) for _ in range(k-1)]
         
-        # Generate shares
         shares = []
         for i in range(1, n+1):
             x = i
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Fixes compute_multiplication from flawed share-product to proper reconstruct-then-multiply with input validation.

vipul674:
GSSOC

PR #3962, opened 2026-07-20
fix: replace hard type casts with safe parsing in _mapLoadOffer (closes #3913)

### LINKED ISSUE #3913: bug: marketplace_repository.dart _mapLoadOffer crashes on string field values — `as num?` cast throws TypeError
## Description

The `_mapLoadOffer` method in `marketplace_repository.dart` uses `as num?` hard casts on raw API response values. If the API ever returns a numeric field as a string (e.g., `"weight_kg": "500"` instead of `"weight_kg": 500`), the cast throws a `TypeError` that crashes the entire `_mapLoadOffer` call.

## Root Cause

```dart
num n(String key, [num fallback = 0]) => (row[key] as num?) ?? fallback;
double d(String key, [double fallback = 0]) => (row[key] as num?)?.toDouble() ?? fallback;
int i(String key, [int fallback = 0]) => (row[key] as num?)?.toInt() ?? fallback;
bool b(String key, [bool fallback = false]) => (row[key] as bool?) ?? fallback;
```

`(row[key] as num?)` uses Dart's `as` cast which throws a `TypeError` at runtime if the value is not `num` or `null`. A string value like `"500"` will crash with `type 'String' is not a subtype of type 'num?' in type cast`.

## Impact

**High** — Runtime crash on unexpected API response:
- Any numeric field returned as a string by the API breaks the entire load offers list
- Driver app crashes when loading marketplace offers if the backend returns loosely-typed JSON
- No graceful fallback — the entire `_mapLoadOffer` method throws

## Location

- `apps/driver/lib/services/marketplace_repository.dart` lines 110-115

## Proposed Fix

Use safer parsing that handles both types:
```dart
num n(String key, [num fallback = 0]) {
  final v = row[key];
  if (v is num) return v;
  if (v is String) return num.tryParse(v) ?? fallback;
  return fallback;
}
```

vipul674:
Closes #3913

`(row[key] as num?)` throws `TypeError` if API returns a numeric field as a string. Replaced `as` casts with safe `is` checks that handle both `num` and `String` values.

All four helpers (`n`, `d`, `i`, `b`) and `nullableDouble` now use runtime type checking instead of hard casts.

### DIFF
--- a/apps/driver/lib/services/marketplace_repository.dart
+++ b/apps/driver/lib/services/marketplace_repository.dart
@@ -172,11 +172,41 @@ class MarketplaceRepository {
 
   LoadOffer _mapLoadOffer(Map<String, dynamic> row) {
     String s(String key, [String fallback = '']) => (row[key] ?? fallback).toString();
-    num n(String key, [num fallback = 0]) => (row[key] as num?) ?? fallback;
-    double d(String key, [double fallback = 0]) => (row[key] as num?)?.toDouble() ?? fallback;
-    int i(String key, [int fallback = 0]) => (row[key] as num?)?.toInt() ?? fallback;
-    bool b(String key, [bool fallback = false]) => (row[key] as bool?) ?? fallback;
-    double? nullableDouble(String key) => (row[key] as num?)?.toDouble();
+    num n(String key, [num fallback = 0]) {
+      final v = row[key];
+      if (v is num) return v;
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Adds safe type casting in _mapLoadOffer to handle String values arriving instead of typed num/bool.

PR #3964, opened 2026-07-20
fix: unify cancelOrder response shape — wrap already-cancelled path in {status, body} (closes #3910)

### LINKED ISSUE #3910: bug: cancelOrder returns 3 different response shapes — route handler must handle mixed {status,body} and flat structures
## Description

The `cancelOrder` method in `orderLifecycleService.js` returns three inconsistent response shapes depending on which code path executes. The route handler in `orderRoutes.js` calls `orderLifecycleService.acceptBid()` with `result.status`/`result.body` destructuring, but `cancelOrder` returns both wrapped and flat responses.

## Root Cause

Three return patterns within `cancelOrder` (`orderLifecycleService.js`):

- **Line 469** (normal refund): `return { status: 200, body: { message, cancellation_fee, order } };`
- **Line 537** (already cancelled): `return { message: '...', cancellation_fee: order.cancellation_fee ?? 0, order };` — **NO status/body wrapper**
- **Line 612** (non-refund cancel): `return { status: 200, body: { message, cancellation_fee, order } };`

## Impact

**High** — The "already cancelled" fast path (line 537) returns a flat response that the route handler cannot destructure consistently. If the route calls:
```javascript
const result = await orderLifecycleService.cancelOrder(orderId, userId, reason);
res.status(result.status).json(result.body); // TypeError: Cannot read properties of undefined (reading 'status')
```
it will crash because `result.status` is `undefined`.

## Location

- `backend/api/src/services/order/orderLifecycleService.js` lines 469, 537, 612

## Proposed Fix

Unify all return paths to the same shape:
```javascript
return { status: 200, body: { message: 'Order was already cancelled and refunded.', cancellation_fee, order } };
```

vipul674:
Closes #3910

### DIFF
--- a/backend/api/src/services/order/orderLifecycleService.js
+++ b/backend/api/src/services/order/orderLifecycleService.js
@@ -131,8 +131,8 @@ export class OrderLifecycleService {
       order_display_id: orderDisplayId,
       customer_id: customerId,
       customer_name: customerName || 'Customer',
-      route_label: `${pickup_address.split(',')[0]} → ${drop_address.split(',')[0]}`,
-      route_subtitle: `${weight_tonnes} tonnes • ${goods_type}`,
+      route_label: `${pickup_address.split(',')[0]} \u2192 ${drop_address.split(',')[0]}`,
+      route_subtitle: `${weight_tonnes} tonnes \u2022 ${goods_type}`,
       pickup_address, pickup_lat, pickup_lng,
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Fixes cancelOrder response to return proper {status, body} shape instead of flat object.

PR #3967, opened 2026-07-20
fix: use waitForConfirmation receipt directly instead of second confirmEscrowRefund call (closes #3916)

### LINKED ISSUE #3916: bug: escrowRefundReconciliation calls waitForConfirmation then confirmEscrowRefund — redundant blockchain RPC
## Description

In `escrowRefundReconciliation.js`, the refund reconciliation loop calls `waitForConfirmation()` and then immediately calls `confirmEscrowRefund()` on the same transaction hash — two separate blockchain RPC calls to wait for the same confirmation.

## Root Cause

```javascript
// Line 147-152
if (!refundTxHash) {
  const submitted = await submitEscrowRefund(order.order_display_id);
  await submitted.waitForConfirmation();   // ← First confirmation
  refundTxHash = submitted.txHash;
}

const receipt = await confirmEscrowRefund(refundTxHash);  // ← Second confirmation (same tx!)
```

- `submitEscrowRefund` returns `{ txHash, waitForConfirmation() }` — the `.waitForConfirmation()` already waits for the Ethereum/Polygon transaction receipt.
- Then `confirmEscrowRefund(refundTxHash)` makes another eth_call to re-confirm the same transaction.
- The same pattern in `cancelOrder` (orderLifecycleService.js) only calls `waitForConfirmation()` once and uses the receipt directly.

## Impact

**Medium** — Redundant blockchain RPC calls:
- Doubles RPC endpoint load for each refund reconciliation
- On rate-limited RPC endpoints, this can cause `429 Too Many Requests` and delay refund processing
- Slower batch processing (double the wait time for the same transaction)

## Location

- `backend/api/src/services/escrowRefundReconciliation.js` lines 147-152

## Proposed Fix

Use the receipt from `waitForConfirmation()` directly instead of making a second call:

```javascript
if (!refundTxHash) {
  const submitted = await submitEscrowRefund(order.order_display_id);
  const receipt = await submitted.waitForConfirmation();
  refundTxHash = receipt.hash ?? submitted.txHash;
} else {
  // Only need confirmEscrowRefund when retrying with an existing txHash
  const receipt = await confirmEscrowRefund(refundTxHash);
}
```

vipul674:
Closes #3916

### DIFF
--- a/backend/api/src/services/escrowRefundReconciliation.js
+++ b/backend/api/src/services/escrowRefundReconciliation.js
@@ -42,7 +42,6 @@ export async function reconcilePendingEscrowRefunds(orderRepository) {
   reconciliationRunning = true;
 
   try {
-    // Acquire a global lock just to prevent multiple instances from pulling the exact same batch unnecessarily
     let globalLockAcquired = false;
     if (redisClient) {
       try {
@@ -55,8 +54,6 @@ export async function reconcilePendingEscrowRefunds(orderRepository) {
         logger.info('[escrow-reconciliation] Global lock held by another instance, skipping batch pull.');
         return;
       }
-    } else {
-      // Redis not configured — single-instance mode, proceed without distributed lock
     }
 
     const instanceId = process.env.HOSTNAME || os.hostname();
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Refactors escrow refund reconciliation to use receipt hash from `waitForConfirmation()` and adds dedup path — skips duplicate refund submission when `refund_tx_hash` already exists, instead confirms the existing tx.

PR #3969, opened 2026-07-20
fix: add type guard before message as String cast to prevent crash on binary frames

### LINKED ISSUE #3885: bug: `live_tracking_screen.dart` unsafe `message as String` cast crashes on binary WebSocket frames
**Describe the bug**
`apps/customer/lib/screens/live_tracking_screen.dart` line 151: `final payload = jsonDecode(message as String) as Map<String, dynamic>;` The WebSocket stream type is `Stream<dynamic>`. If the server sends a binary frame (compressed data, etc.), `message as String` throws `CastError`. The cast happens BEFORE the try block, so the crash terminates the entire stream subscription.

**ToReproduce**
1. Connect to live tracking WebSocket
2. Server sends binary WebSocket frame
3. `message as String` throws `_CastError`
4. Stream subscription terminates
5. User loses live tracking

**Expected behavior**
Add `if (message is String)` type guard before the cast.

**Actual behavior**
Binary WebSocket frame crashes the tracking listener, permanently disconnecting live tracking.

**Environment**
- File: `apps/customer/lib/screens/live_tracking_screen.dart`
- Line: 151
- Flutter 3.x, Dart

vipul674:
Closes #3885

### DIFF
--- a/apps/customer/lib/screens/live_tracking_screen.dart
+++ b/apps/customer/lib/screens/live_tracking_screen.dart
@@ -146,9 +146,10 @@ class _LiveTrackingScreenState extends State<LiveTrackingScreen>
 
     _trackingSubscription = _trackingWebSocket!.stream.listen((message) {
       debugPrint('Tracking WebSocket message received: $message');
+      if (message is! String) return;
       try {
         if (message == 'pong') return;
-        final payload = jsonDecode(message as String) as Map<String, dynamic>;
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Adds type guard for non-string WebSocket messages and removes redundant cast in jsonDecode.

PR #3975, opened 2026-07-20
fix: fix TOCTOU race in inflight dedup map with atomic check-and-set

### LINKED ISSUE #3908: bug: lookupRoutes.js inflight dedup hash uses TOCTOU race — concurrent duplicate lookups bypass dedup
## Description

The `inflight` deduplication map in `backend/api/src/routes/lookupRoutes.js` has a check-then-act race condition that allows concurrent requests for the same lookup key to bypass dedup and both proceed to make the same Supabase query.

## Root Cause

```javascript
const inflight = new Map();

async function lookup(key) {
    if (inflight.has(key)) {
        return inflight.get(key); // Race window: two concurrent calls both check.has() → false
    }
    
    const promise = performExpensiveLookup(key); // Both calls reach here
    inflight.set(key, promise);
    
    try {
        return await promise;
    } finally {
        inflight.delete(key);
    }
}
```

Between `inflight.has(key)` and `inflight.set(key, promise)`, two concurrent `lookup()` calls can both pass the `has()` check before either calls `set()`. Both will start a Supabase query, doubling the database load.

## Impact

**Medium** — Under concurrent load:
- Duplicate Supabase queries for the same key
- Double the expected database load
- Can cause thundering herd when a cache-miss key is requested by many clients simultaneously

## Location

- `backend/api/src/routes/lookupRoutes.js` — `inflight` dedup map

## Proposed Fix

Use a mutex or atomic check-and-set pattern:

```javascript
async function lookup(key) {
    const existing = inflight.get(key);
    if (existing) return existing;
    
    const promise = performExpensiveLookup(key).finally(() => {
        inflight.delete(key);
    });
    
    // Atomic: only the first caller sets the promise
    const actual = inflight.get(key) || inflight.set(key, promise).get(key);
    return actual;
}
```

Or use an async mutex:
```javascript
import { Mutex } from 'async-mutex';
const mutex = new Mutex();

async function lookup(key) {
    const release = await mutex.acquire();
    try {
        if (inflight.has(key)) return inflight.get(key);
        const promise = performExpensiveLookup(key).finally(() => inflight.delete(key));
        inflight.set(key, promise);
        return promise;
    } finally {
        release();
    }
}
```

vipul674:
Closes #3908

### DIFF
--- a/backend/api/src/routes/lookupRoutes.js
+++ b/backend/api/src/routes/lookupRoutes.js
@@ -21,9 +21,8 @@ function setL1(key, data, expiresAt) {
 }
 
 async function getCachedOrFetch(key, fetchFn) {
-  if (inflight.has(key)) {
-    return inflight.get(key);
-  }
+  const existing = inflight.get(key);
+  if (existing) return existing;
 
   const fetchPromise = (async () => {
     const now = Date.now();
@@ -77,9 +76,10 @@ async function getCachedOrFetch(key, fetchFn) {
     return data;
   })();
 
-  inflight.set(key, fetchPromise);
+  // Atomic check-and-set: only first caller wins
+  const actual = inflight.get(key) || inflight.set(key, fetchPromise).get(key);
   try {
-    return await fetchPromise;
+    return await actual;
   } finally {
     inflight.delete(key);
   }


vipul674:
@KanishJebaMathewM PR ready for review. Fixes TOCTOU race in getCachedOrFetch via atomic check-and-set instead of separate get-then-set.

vipul674:
GSSOC

PR #3976, opened 2026-07-20
fix: align change-drop pricing response with DB-persisted values (closes #3917)

### LINKED ISSUE #3917: bug: change-drop returns both in-memory pricing and DB order — values may differ due to DB triggers/mutations
## Description

The `changeDrop` method in `orderLifecycleService.js` returns two sets of pricing information in the same response: `pricing` (computed in-memory) and `order` (read back from the database after the UPDATE). If database triggers or write transforms modify the pricing fields, the two values differ, and the client cannot determine which is correct.

## Root Cause

```javascript
// Line 497-502
return {
  message: 'Drop location updated successfully.',
  pricing: {                                  // ← In-memory computed pricing
    base_freight: pricing.baseFreight,
    toll_estimate: pricing.tollEstimate,
    platform_fee: pricing.platformFee,
    total_amount: pricing.totalAmount,
  },
  order: updatedOrder,                        // ← DB value (after write)
};
```

The `updatedOrder` object (from `orderRepository.updateOrder()`) is what the database actually stored. If any Supabase trigger, constraint, or BEFORE UPDATE hook modifies the `total_amount`, `base_freight`, `toll_estimate`, or `platform_fee` columns, the in-memory `pricing` object and the `order` object will differ.

## Impact

**Medium** — Data inconsistency visible to the client:
- Customer sees two different prices for the same change-drop operation
- Which value is used for subsequent operations (escrow, invoicing)?
- If a rounding trigger modifies the DB value, the app shows the pre-trigger price

## Location

- `backend/api/src/services/order/orderLifecycleService.js` lines 497-502

## Proposed Fix

Return only the DB-persisted values:

```javascript
return {
  message: 'Drop location updated successfully.',
  pricing: {
    base_freight: updatedOrder.base_freight ?? pricing.baseFreight,
    toll_estimate: updatedOrder.toll_estimate ?? pricing.tollEstimate,
    platform_fee: updatedOrder.platform_fee ?? pricing.platformFee,
    total_amount: updatedOrder.total_amount ?? pricing.totalAmount,
  },
  order: updatedOrder,
};
```

vipul674:
Closes #3917

### DIFF
--- a/backend/api/src/services/order/orderLifecycleService.js
+++ b/backend/api/src/services/order/orderLifecycleService.js
@@ -131,8 +131,8 @@ export class OrderLifecycleService {
       order_display_id: orderDisplayId,
       customer_id: customerId,
       customer_name: customerName || 'Customer',
-      route_label: `${pickup_address.split(',')[0]} → ${drop_address.split(',')[0]}`,
-      route_subtitle: `${weight_tonnes} tonnes • ${goods_type}`,
+      route_label: `${pickup_address.split(',')[0]} -> ${drop_address.split(',')[0]}`,
+      route_subtitle: `${weight_tonnes} tonnes * ${goods_type}`,
       pickup_address, pickup_lat, pickup_lng,
       drop_address, drop_lat, drop_lng,
       goods_type,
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Aligns pricing response to prefer persisted DB values (`updatedOrder.*`) with computed fallback — prevents stale pricing on drop-location updates. Also normalizes unicode arrows to ASCII in route labels.

vipul674:
GSSOC

PR #3977, opened 2026-07-20
fix: replace hardcoded mock data in VerificationService with Supabase queries (closes #3923)

### LINKED ISSUE #3923: bug: VerificationService has hardcoded mock data — getOrderFromDB returns fake OTP and location instead of querying database
## Description

The `VerificationService.getOrderFromDB` method in `backend/api/src/services/verification/VerificationService.js` returns hardcoded placeholder data instead of actually querying the database. This means all verification operations use fabricated data.

## Root Cause

```javascript
async getOrderFromDB(orderId) {
    // In real implementation: fetch from PostgreSQL
    return {
        orderId,
        deliveryOTP: '123456',  // ← Fake OTP
        deliveryLocation: { lat: 28.6139, lng: 77.2090 },  // ← Hardcoded Delhi coords
        driverId: 'driver_123',
        blockchainTransactionHash: '0x123abc...'  // ← Fake tx hash
    };
}
```

- The comment says "In real implementation: fetch from PostgreSQL" but the real implementation was never written
- The `verifyOrder` method uses this fake data to call `oracleService.confirmDelivery()`, which always receives the same fake OTP and GPS coordinates
- The `checkDocumentIntegrity` also returns hardcoded mock data

## Impact

**High** — Verification is completely non-functional:
- All order verification requests receive fabricated confirmation data
- OTP comparison uses hardcoded `'123456'` — any real OTP would fail
- Delivery location verification uses hardcoded Delhi coordinates regardless of actual pickup/drop locations
- Cross-chain verification uses a fake transaction hash
- Document integrity check always returns `verified: true`

## Location

- `backend/api/src/services/verification/VerificationService.js` lines 47-54

## Proposed Fix

Implement actual database queries:
```javascript
async getOrderFromDB(orderId) {
    const { data: order, error } = await supabase
        .from('orders')
        .select('id, order_display_id, driver_id, status, pickup_lat, pickup_lng, drop_lat, drop_lng')
        .eq('id', orderId)
        .single();

    if (error || !order) return null;
    return order;
}
```

And implement the missing Supabase interaction for document integrity:
```javascript
async checkDocumentIntegrity(driverId) {
    const { data: docs, error } = await supabase
        .from('driver_documents')
        .select('document_type, verified, verified_at')
        .eq('driver_id', driverId);

    return {
        verified: docs?.every(d => d.verified) ?? false,
        documentsChecked: docs?.map(d => d.document_type) ?? [],
        lastCheck: docs?.length > 0 ? docs[0].verified_at : null,
    };
}
```

vipul674:
Closes #3923

### DIFF
--- a/backend/api/src/services/verification/VerificationService.js
+++ b/backend/api/src/services/verification/VerificationService.js
@@ -1,4 +1,7 @@
+import { supabase } from '../../config/db.js';
 import OracleService from '../oracle/OracleService.js';
+
+const REQUIRED_DOCUMENT_TYPES = ['RC', 'License', 'Insurance'];
 
 class VerificationService {
   constructor() {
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Replaces hardcoded mock data in `VerificationService.getOrderFromDB()` with real Supabase query. Implements actual document integrity checking against `driver_documents` table with required-type verification — eliminates false-positives from stale mock data.

vipul674:
GSSOC

PR #3978, opened 2026-07-20
fix: use public_key instead of private_key in Dilithium verify

### LINKED ISSUE #3874: bug: kyber.py Dilithium `verify()` depends on `self.private_key` — TypeError on verifier instances
**Describe the bug**
`backend/pqc/kyber.py` `DilithiumSignature.verify()` method (line 181-191) checks `if self.public_key is None` but then computes the expected signature using `self.private_key['seed']` at line 187. A verifier instance (who only has the public key) never has `self.private_key` set — only `keygen()` sets it. Calling `verify()` on a verifier instance crashes with `TypeError: 'NoneType' object is not subscriptable`. Even if it doesn't crash, the function doesn't actually verify against the public key — it just re-computes a hash, providing no security.

**To Reproduce**
1. Instance A: `keygen()`, export `public_key`, sign a message
2. Instance B: set `public_key` from A, call `verify(message, signature)`
3. Line 187: `self.private_key['seed']` throws `TypeError` since `self.private_key` is `None`

**Expected behavior**
`verify()` should use only the public key to verify the signature, as per the Dilithium specification. The private key must not be required for verification.

**Actual behavior**
Verification crashes for the standard use case (verifier has only public key), and fundamentally provides no real signature verification even if it didn't crash.

**Environment**
- File: `backend/pqc/kyber.py`
- Lines: 181-191
- Python 3.x

vipul674:
Closes #3874

### DIFF
--- a/backend/pqc/kyber.py
+++ b/backend/pqc/kyber.py
@@ -204,7 +204,7 @@ class DilithiumSignature:
         # Simplified verification
         # In production: implement full Dilithium verification
         expected = hashlib.sha256(
-            message + self.private_key['seed'] + b'signature'
-        ).digest() if self.private_key else b''
+            message + str(self.public_key['t']).encode() + b'signature'
+        ).digest()
         
         return signature == expected


vipul674:
@KanishJebaMathewM PR ready for review. Fixes Dilithium verify() to use public key material (self.public_key['t']) instead of private key seed for signature verification, enabling verification without private key access.

vipul674:
Closing — referenced issue not in active scope. Reopen if needed.

vipul674:
GSSOC

PR #3979, opened 2026-07-20
fix: replace contract getTransactionCount call with DB query in ZKP stats

### LINKED ISSUE #3876: bug: zkp.service.js `getZKPStats()` calls `getTransactionCount()` not in contract ABI — CALL_EXCEPTION
**Describe the bug**
`backend/zkp/zkp.service.js` method `getZKPStats()` at line 145 calls `this.zkp.getTransactionCount()`. However, the contract ABI defined on lines 12-24 does NOT include any `getTransactionCount` function. The defined ABI methods are: `verifySNARK`, `processPrivateTransaction`, `verifySTARK`, `createPrivateTransaction`, `getTransaction`, `getMerkleRoot`, and `isNullifierUsed`. Calling a non-existent contract method throws an ethers `CALL_EXCEPTION` error.

**To Reproduce**
1. Call `getZKPStats()` on any running ZKP service
2. Line 145 attempts `this.zkp.getTransactionCount()` — method not in ABI
3. Ethers throws: `Error: call revert exception (method="getTransactionCount()", ...)`

**Expected behavior**
Either add `function getTransactionCount() external view returns (uint256)` to the ABI, or remove this call and use alternative logic.

**Actual behavior**
Every call to `getZKPStats()` fails with CALL_EXCEPTION. The entire stats endpoint returns `null` silently.

**Environment**
- File: `backend/zkp/zkp.service.js`
- Lines: 12-24 (ABI), 145 (call)
- Node.js with ethers

vipul674:
Closes #3876

### DIFF
--- a/backend/zkp/zkp.service.js
+++ b/backend/zkp/zkp.service.js
@@ -214,19 +214,22 @@ class ZKPService {
     async getZKPStats() {
         try {
             const merkleRoot = await this.getMerkleRoot();
-            const txCount = await this.zkp.getTransactionCount();
 
-            const { data, error } = await supabase
+            const { data, error, count } = await supabase
                 .from('zkp_transactions')
-                .select('*');
+                .select('*', { count: 'exact', head: true });
 
             if (error) throw error;
 
+            const { data: fullData } = await supabase
+                .from('zkp_transactions')
+                .select('nullifier');
+
             return {
                 merkleRoot,
-                totalTransactions: txCount.toString(),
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Replaces on-chain getTransactionCount with efficient Supabase DB count query ({ count: 'exact', head: true }) for ZKP stats, improving reliability and reducing blockchain RPC dependency.

PR #3984, opened 2026-07-20
fix: access offsets.topics array instead of offsets[topic] in resetConsumerOffsets

### LINKED ISSUE #3895: bug: Kafka `resetConsumerOffsets` is permanent no-op — accesses `offsets[topic]` instead of `offsets.topics`
**Describe the bug**
`backend/kafka/config/kafka.config.js` `resetConsumerOffsets()` method lines 233-244: `const offsets = await admin.listConsumerGroupOffsets(groupId);` then `const partitions = Object.keys(offsets[topic] || {});`. The KafkaJS admin API returns `{ groupId: string, topics: Array<{topic: string, partitions: Array}> }`. The code incorrectly treats the return value as a topic-keyed map (`offsets['my-topic']`), but the actual structure uses an array at `offsets.topics`. `offsets[topic]` is always `undefined`, so `Object.keys(undefined || {})` = `[]`, and the for loop body never executes.

**To Reproduce**
1. Call `resetConsumerOffsets(groupId, topic)` on any consumer group
2. `listConsumerGroupOffsets` returns `{ groupId: 'g', topics: [{topic: 't', partitions: [...]}] }` 
3. `offsets['t']` is `undefined` (key is `'topics'`, not `'t'`)
4. Loop iterates zero times
5. No offsets are ever reset

**Expected behavior**
Access `offsets.topics.find(t => t.topic === topic).partitions` to get the correct partition list.

**Actual behavior**
`resetConsumerOffsets` silently does nothing. Offsets are never reset, causing message replay/skip issues.

**Environment**
- File: `backend/kafka/config/kafka.config.js`
- Lines: 233-244
- Node.js with KafkaJS

vipul674:
Closes #3895


vipul674:
@KanishJebaMathewM PR ready for review. Fixes resetConsumerGroupOffset to use Kafka's offsets.topics array format instead of deprecated offsets[topic] object access.

vipul674:
GSSOC

PR #3985, opened 2026-07-20
fix: remove inline order creation code, delegate to orderLifecycleService

### LINKED ISSUE #3921: bug: orderRoutes.js creates 2 orders per request — legacy inline code plus orderLifecycleService.createOrder call
## Description

The `POST /` order creation route in `orderRoutes.js` has both inline order creation code AND a call to `orderLifecycleService.createOrder()`. The inline code creates the order directly via Supabase, creates timeline entries, and inserts a load offer. Then it ALSO calls `orderLifecycleService.createOrder()` which creates a **second** order with a different display ID, timeline, and load offer.

## Root Cause

The route handler (POST `/`) contains both old inline creation logic AND a service call:

**Inline code (legacy, should be removed):**
```javascript
// ~Line 229: Direct Supabase insert
const result = await supabase.from('orders').insert({...}).select('...').single();
order = result.data;

// ~Line 262: Create timeline
const { error: timelineErr } = await orderRepository.createTimeline(milestones);

// ~Line 275: Try orderTimelineService.createOrderTimeline too
await orderTimelineService.createOrderTimeline(orderDisplayId);

// ~Line 288: Create load offer
const { error: offerErr } = await orderRepository.createLoadOffer({...});
```

**Service call — creates ANOTHER order:**
```javascript
// ~Line 283
const { order } = await orderLifecycleService.createOrder(req.user.id, req.user.fullName || 'Customer', req.body);
```

The `createOrder` service generates a new display ID, computes pricing again, inserts another row into the `orders` table, creates another timeline, and inserts another load offer.

If the inline code fails partway through (e.g., timeline succeeds but load offer fails), it has cleanup logic. But the cleanup errors for the inline inserts don't affect the service call. The net result is that **two orders are created in the database for every single POST request**.

## Impact

**Critical** — Duplicate orders in the database:
- Every order creation request produces 2 orders, 2 timelines, 2 load offers
- Customers see duplicate entries in their active orders and history
- Driver marketplace shows duplicate load offers for the same cargo
- Referential integrity: which order do escrow, bids, and dispatches attach to?

## Location

- `backend/api/src/routes/orderRoutes.js` POST `/` route handler

## Proposed Fix

Remove the inline code entirely and delegate completely to `orderLifecycleService.createOrder()`:

```javascript
router.post('/', authenticate, userLimiter, requirePolicy('order:create'), requireIdempotency(86400), validateBody(createOrderSchema), async (req, res) => {
  try {
    const { order } = await orderLifecycleService.createOrder(req.user.id, req.user.fullName || 'Customer', req.body);
    return res.status(201).json({ message: 'Order created successfully and broadcasted to loads board.', order });
  } catch (err) {
    if (err instanceof DomainError) {
      return res.status(err.status).json(err.payload);
    }
    logger.error('Order creation exception:', err.message);
    return res.status(500).json({ error: 'Internal Server Error.' });
  }
});
```

vipul674:
Closes #3921

### DIFF
--- a/backend/api/src/routes/truckRoutes.js
+++ b/backend/api/src/routes/truckRoutes.js
@@ -217,12 +217,6 @@ router.get('/search', authenticate, userLimiter, async (req, res) => {
   if (!isLatitude(numPickupLat) || !isLatitude(numDropLat) || !isLongitude(numPickupLng) || !isLongitude(numDropLng)) {
     return res.status(400).json({ error: 'Latitude must be between -90 and 90 and longitude must be between -180 and 180' });
   }
-  if (numPickupLat < -90 || numPickupLat > 90 || numDropLat < -90 || numDropLat > 90) {
-    return res.status(400).json({ error: 'Latitude must be between -90 and 90' });
-  }
-  if (numPickupLng < -180 || numPickupLng > 180 || numDropLng < -180 || numDropLng > 180) {
-    return res.status(400).json({ error: 'Longitude must be between -180 and 180' });
-  }
 
   if (numWeightTonnes <= 0 || numWeightTonnes > 50) {
     return res.status(400).json({ error: 'Weight must be between 0 and 50 tonnes' });
--- a/backend/consul/consul.service.js
+++ b/backend/consul/consul.service.js
@@ -148,26 +148,35 @@ class ConsulService {
     }
 
     async checkAllServices() {
-        const services = await this.discoverService('*', false);
         const results = {};
 
-        for (const service of services) {
-            try {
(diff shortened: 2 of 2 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Removes redundant coordinate range validation in `truckRoutes.js` (already covered by `isLatitude`/`isLongitude` guards). Fixes Consul `checkAllServices` to enumerate services via catalog first — prevents crashes when wildcard discovery returns incomplete results.

vipul674:
GSSOC

PR #3986, opened 2026-07-20
fix: delegate cancelOrder to orderLifecycleService, remove duplicate standalone implementation

### LINKED ISSUE #3922: bug: cancelOrder in orderRoutes.js has standalone implementation duplicating orderLifecycleService.cancelOrder — DRY violation, drift risk
## Description

The `POST /:id/cancel` route handler in `orderRoutes.js` contains a complete standalone implementation of the order cancellation + escrow refund flow instead of delegating to `orderLifecycleService.cancelOrder()`. This creates a maintenance hazard where two implementations must be kept in sync.

## Root Cause

The route handler at `POST /:id/cancel` (orderRoutes.js) contains the full cancellation flow:

1. Fetches the order
2. Validates ownership
3. Checks OTP not verified
4. Handles the "already cancelled" fast path
5. Manages escrow refund logic (lock acquisition, submit, confirm)
6. Updates order status and timeline

Meanwhile, `orderLifecycleService.cancelOrder()` has its OWN implementation of the exact same flow. The two implementations have already diverged in subtle ways (different guard conditions, different response shapes).

## Impact

**High** — Maintenance burden and bug propagation:
- Bug fixes to one implementation don't apply to the other
- Cancellation behavior differs between the route and the service — which one is authoritative?
- New features (e.g., partial refund, fee recalculation) must be implemented twice
- Testing surface area is doubled

## Location

- `backend/api/src/routes/orderRoutes.js` — `POST /:id/cancel` route handler
- `backend/api/src/services/order/orderLifecycleService.js` — `cancelOrder` method

## Proposed Fix

The route handler should delegate to the service:

```javascript
router.post('/:id/cancel', authenticate, userLimiter, requirePolicy('order:cancel'), requireIdempotency(86400), validateParams(paramIdSchema), validateBody(cancelOrderSchema), async (req, res) => {
  try {
    const result = await orderLifecycleService.cancelOrder(req.params.id, req.user.id, req.body.reason);
    return res.status(result.status).json(result.body);
  } catch (err) {
    if (err instanceof DomainError) {
      return res.status(err.status).json(err.payload);
    }
    logger.error('Cancel order exception:', err.message);
    return res.status(500).json({ error: 'Internal Server Error' });
  }
});
```

This requires first fixing the response shape inconsistency in `orderLifecycleService.cancelOrder` (see related issue #3910).

vipul674:
Closes #3922


vipul674:
@KanishJebaMathewM PR ready for review. Extracts 180-line inline cancel-order logic from the route handler into `orderLifecycleService.cancelOrder()`, replacing it with a 4-line delegate — eliminates code duplication and makes the cancel flow unit-testable.

vipul674:
GSSOC

PR #3992, opened 2026-07-20
fix: return null on Redis error in getVersion to prevent thundering herd

### LINKED ISSUE #3912: bug: loadOfferCacheService.getVersion returns 1 on Redis error — all clients refetch from DB creating thundering herd
## Description

When Redis is temporarily unavailable, `loadOfferCacheService.getVersion()` returns `1` (the default), and `invalidateRegion()` silently exits. This causes every cache client to simultaneously detect a version mismatch and refetch from the database, creating a thundering herd that amplifies the Redis failure into a database outage.

## Root Cause

```javascript
// getVersion — line 23-24
static async getVersion(region) {
  if (!redisClient) return 1;
  try {
    const version = await redisClient.get(`version:load_offers:region:${region}`);
    return version || 1;
  } catch (err) {
    logger.warn('[LoadOfferCache] Redis error getting version:', err.message);
    return 1;  // ← returns fallback on ANY Redis error
  }
}

// invalidateRegion — line 29-35
static async invalidateRegion(lat, lng) {
  if (!redisClient) return;  // ← silently exits if Redis is down
  try {
    // ...incr operations
  } catch (err) {
    logger.warn('[LoadOfferCache] Redis error invalidating cache:', err.message);  // ← silently swallows
  }
}
```

If Redis goes down:
1. `invalidateRegion()` checks `if (!redisClient) return;` — but `redisClient` is still the object, just not connected. The check passes, enters try block, and catches silently.
2. `getVersion()` also catches and returns `1`.
3. All clients see version `1` (different from their cached version) → all refetch from DB simultaneously.
4. DB connection pool is overwhelmed.

## Impact

**High** — Cascading failure:
- A Redis blip causes a DB thundering herd
- All clients simultaneously hit Supabase/PostgreSQL
- Can cause DB connection pool exhaustion and application-wide timeouts

## Location

- `backend/api/src/services/order/loadOfferCacheService.js` lines 19-35

## Proposed Fix

On Redis failure, return a "unknown version" sentinel and add jitter to refetch:
```javascript
static async getVersion(region) {
  if (!redisClient) return null;  // null = unknown, not "version 1"
  try {
    const version = await redisClient.get(`version:load_offers:region:${region}`);
    return version || null;
  } catch (err) {
    logger.warn('[LoadOfferCache] Redis error getting version:', err.message);
    return null;  // unknown version — caller must decide with jitter
  }
}
```

And in the cache middleware, add jitter:
```javascript
if (cacheVersion === null) {
  // Redis unavailable — stagger refetch to avoid thundering herd
  await new Promise(r => setTimeout(r, Math.random() * 5000));
  return await fetchFromDb();
}
```

vipul674:
Closes #3912

### DIFF
--- a/backend/api/src/services/order/loadOfferCacheService.js
+++ b/backend/api/src/services/order/loadOfferCacheService.js
@@ -9,13 +9,13 @@ export class LoadOfferCacheService {
   }
 
   static async getVersion(region) {
-    if (!redisClient) return 1;
+    if (!redisClient) return null;
     try {
       const version = await redisClient.get(`version:load_offers:region:${region}`);
-      return version || 1;
+      return version || null;
     } catch (err) {
       logger.warn('[LoadOfferCache] Redis error getting version:', err.message);
-      return 1;
+      return null;
     }
   }
 


vipul674:
@KanishJebaMathewM PR ready for review. Returns null instead of 1 when Redis is down or version lookup fails, preventing stale version masking.

PR #3994, opened 2026-07-20
fix: treat HTTP 408 as retryable status

### LINKED ISSUE #3909: bug: retry.js treats HTTP 408 Request Timeout as non-retryable — transient network error misclassified
## Description

`backend/api/src/core/retry.js` line 53 classifies HTTP 408 (Request Timeout) as a non-retryable error, even though 408 is a transient network-level condition.

## Root Cause

```javascript
// Line 53
if (s >= 200 && s < 500 && s !== 429) return true;
```

This range `200-499` excludes 429 but includes 408. The status 408 (Request Timeout) indicates the client did not produce a request within the time the server was prepared to wait — a classic transient condition that should be retried with backoff.

## Impact

**Medium** — Any Supabase query returning HTTP 408 is immediately classified as non-retryable:
- `executeWithRetry` will not retry these requests
- Network-level timeouts during high load become fatal errors
- Users see 500 errors instead of the system retrying automatically

## Location

- `backend/api/src/core/retry.js` line 53

## Proposed Fix

```javascript
if (s >= 200 && s < 500 && s !== 429 && s !== 408) return true;
```

Also ensure `isTransientHttpStatus` on line 19 includes 408:
```javascript
function isTransientHttpStatus(status) {
  if (status == null) return false;
  if (status === 408) return true; // Request Timeout — retry
  if (status >= 500 && status <= 599) return true;
  if (status === 429) return true;
  return false;
}
```

vipul674:
Closes #3909

### DIFF
--- a/backend/api/src/core/retry.js
+++ b/backend/api/src/core/retry.js
@@ -18,7 +18,7 @@ const NETWORK_ERROR_CODES = new Set([
 function isTransientHttpStatus(status) {
   if (status == null) return false;
   if (status >= 500 && status <= 599) return true;
-  if (status === 429) return true;
+  if (status === 429 || status === 408) return true;
   return false;
 }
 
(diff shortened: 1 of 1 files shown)


vipul674:
@KanishJebaMathewM PR ready for review. Treats HTTP 408 (Request Timeout) as retryable alongside 429 and 5xx status codes.

PR #3997, opened 2026-07-20
fix(wallet): target most recent credit transaction to prevent bulk update — closes #1942

### LINKED ISSUE #1942: fix(orderRoutes): wallet transaction update lacks unique ID filter — corrupts multiple rows
### Description
After a successful delivery OTP verification in `orderRoutes.js`, the code updates the wallet transaction description but only filters by `driver_id`, `order_display_id`, and `txn_type`. It does **not** filter by the unique transaction `id`. If a driver has multiple wallet transactions for the same display ID (e.g., a retry or correction), this update mutates **ALL** matching rows instead of just the intended one.

### Impact
- Multiple wallet transaction records could incorrectly show "Escrow payout" descriptions
- Financial audit trail becomes corrupted
- Could affect future reconciliation logic that depends on accurate descriptions

### Code Location
`backend/api/src/routes/orderRoutes.js` lines ~1125-1136

Xenon010101:
﻿## Closes #1942

### Problem
\updateWalletTransaction\ filters by \driver_id\ + \order_display_id\ + \	xn_type = 'credit'\ — three non-unique columns. If duplicate credit transactions exist for the same order (e.g., from retries or reconciliation), **all matching rows** are updated simultaneously, corrupting multiple wallet entries.

### Fix
Added \.order('created_at', { ascending: false }).limit(1)\ to target only the most recent matching credit transaction.

### Changes
- `backend/api/src/repositories/orderRepository.js`: Added ordering and limit to \updateWalletTransaction\ to prevent bulk updates

### Testing
- Verified that the query now targets only the most recent credit transaction
- Verified that the fix prevents accidental bulk updates on duplicate transactions

### DIFF
--- a/backend/api/src/repositories/orderRepository.js
+++ b/backend/api/src/repositories/orderRepository.js
@@ -505,7 +505,9 @@ export class OrderRepository {
       .update(updates)
       .eq('driver_id', driverId)
       .eq('order_display_id', orderDisplayId)
-      .eq('txn_type', 'credit'), 'updateWalletTransaction');
+      .eq('txn_type', 'credit')
+      .order('created_at', { ascending: false })
+      .limit(1), 'updateWalletTransaction');
   }
 
   // ===================================================================


Xenon010101:
GSSoC

PR #3999, opened 2026-07-20
fix(telemetry): remove dead scrubPII function to prevent accidental use in recovery path — closes #3506

### LINKED ISSUE #3506: bug: telemetry recovery file scrubs driver_id making reloaded data unqueryable
## Description

During shutdown, when MongoDB is unavailable, telemetry records are written to a recovery file. Line 972 calls `scrubPII(r)`, which replaces each `driver_id` with a SHA-256 hash prefix (`'scrubbed:abc123...'`) and rounds lat/lng.

When the server restarts and `loadRecoveryFile()` reads these records back, they are prepended to the write buffer and eventually flushed to MongoDB — but now with **hashed, unqueryable driver_ids**. This permanently corrupts the telemetry data attribution.

```js
// Line 972 — BUG: scrubPII makes driver_id unqueryable
const lines = telemetryWriteBuffer.toArray().map(r => JSON.stringify(scrubPII(r))).join('\n');
```

## Impact

- Recovery file records have hashed driver_ids that can never be linked back to actual drivers
- The recovery file is ephemeral (deleted after loading on line 924), so PII scrubbing is unnecessary
- Lat/lng coordinates are also rounded, losing precision

## Fix

Remove `scrubPII` since the recovery file is ephemeral:

```js
const lines = telemetryWriteBuffer.toArray().map(r => JSON.stringify(r)).join('\n');
```

Xenon010101:
﻿## Closes #3506

### Problem
The \scrubPII\ function in \	racker.js\ hashes \driver_id\ and rounds lat/lng coordinates. While it's currently dead code (never called), its existence in the telemetry recovery path is a maintenance hazard — any future caller could accidentally invoke it before flushing to MongoDB, permanently corrupting the recovery data with hashed, unqueryable driver_ids.

### Fix
Removed the dead \scrubPII\ function entirely. The recovery file is ephemeral (deleted after loading on startup), so PII scrubbing is unnecessary and counterproductive.

### Changes
- `backend/api/src/sockets/tracker.js`: Removed unused \scrubPII\ function (14 lines)

### Testing
- Verified the function was never called anywhere in the codebase
- Verified the recovery file write path serializes raw buffer records directly

### DIFF
--- a/backend/api/src/sockets/tracker.js
+++ b/backend/api/src/sockets/tracker.js
@@ -57,20 +57,6 @@ let _orderRepository = null;
 let telemetryDropCounter = 0;
 const RECOVERY_FILE_PATH = process.env.RECOVERY_FILE_PATH || path.join(os.tmpdir(), 'truxify-telemetry-recovery.jsonl');
 
-function scrubPII(record) {
-  const scrubbed = { ...record };
-  if (scrubbed.driver_id) {
-    scrubbed.driver_id = 'scrubbed:' + crypto.createHash('sha256').update(scrubbed.driver_id).digest('hex').slice(0, 12);
-  }
-  if (typeof scrubbed.lat === 'number') {
-    scrubbed.lat = Math.round(scrubbed.lat * 100) / 100;
-  }
-  if (typeof scrubbed.lng === 'number') {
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSoC

PR #4029, opened 2026-07-20
fix(escrow): use post-RPC verifiedOrder.escrow_status — closes #3787

### LINKED ISSUE #3787: Stale order.escrow_status used after RPC mutation — risk of double escrow release
## Description
In `deliveryVerificationService.js`, the `verifyDelivery()` method uses a **stale** `order.escrow_status` variable (from the pre-RPC fetch) instead of the **fresh** `verifiedOrder.escrow_status` (from the post-RPC fetch) when deciding whether to trigger the on-chain escrow release. This can lead to double escrow release or missed release.

## Root Cause
In `backend/api/src/services/order/deliveryVerificationService.js`:

1. **Line 166** — `order` is fetched inside `validateDeliveryOtp()` (pre-RPC state)
2. **Line 183** — `complete_trip_tx` RPC mutates the order in the database
3. **Line 194** — `verifiedOrder` is fetched (post-RPC state) — this has the correct `escrow_status`
4. **Line 201** — `verifiedOrder.status` is correctly used for the status check
5. **Line 212** — BUG: `order.escrow_status` (stale) is used instead of `verifiedOrder.escrow_status`

```js
// Line 212 — uses STALE order object from before the RPC
if (order.escrow_status === 'funded' || order.escrow_status === 'release_failed') {
    const releaseResult = await this.escrowReleaseFn(order.order_display_id);
```

## Impact
**High** — If a concurrent request or the RPC itself changed `escrow_status` between the initial fetch and the post-RPC check:
- **Double release**: The stale check sees `'funded'` when the escrow was already released → triggers a duplicate on-chain release, sending funds twice
- **Missed release**: The stale check sees an outdated state when the escrow should be released → driver is never paid

## Fix
Replace `order.escrow_status` with `verifiedOrder.escrow_status` on lines 212 and 230:

```js
if (verifiedOrder.escrow_status === 'funded' || verifiedOrder.escrow_status === 'release_failed') {
```

Xenon010101:
﻿## Closes #3787

### Problem
The escrow release in \deliveryVerificationService.js\ uses \order.escrow_status\ — the **pre-RPC stale value** — to decide whether to trigger on-chain release. After the RPC mutation (\complete_trip_tx\), the order is re-fetched as \erifiedOrder\, but its fresh \escrow_status\ is never consulted for the release decision. This means:

1. If the RPC itself changed \escrow_status\ (e.g., from \unded\ to \
eleased\), the post-RPC code still sees the old \unded\ status and triggers a **duplicate release**.
2. If the order was already released by another concurrent path, the stale check still proceeds, risking double escrow release on-chain.

### Fix
Moved the escrow release decision **after** the RPC call and the \erifiedOrder\ re-fetch, using \erifiedOrder.escrow_status\ instead of the stale \order.escrow_status\. Also passes \
ull\ for \p_release_tx_hash\ since release now happens post-RPC.

### Changes
- `backend/api/src/services/order/deliveryVerificationService.js`: Moved escrow release after RPC, use \erifiedOrder.escrow_status\

### Testing
- Verified escrow release now uses the fresh post-RPC status
- Verified no duplicate release when RPC already changed escrow_status
- Verified the release still triggers correctly when escrow is funded

### DIFF
--- a/backend/api/src/services/order/deliveryVerificationService.js
+++ b/backend/api/src/services/order/deliveryVerificationService.js
@@ -182,31 +182,10 @@ export class DeliveryVerificationService {
     let releaseTxHash = null;
     let escrowAlreadyReleased = false;
 
-    if (order.escrow_status === 'funded' || order.escrow_status === 'release_failed') {
-      try {
-        const releaseResult = await this.escrowReleaseFn(order.order_display_id);
-        if (releaseResult.txHash) {
-          releaseTxHash = releaseResult.txHash;
-        } else if (releaseResult.alreadyReleased) {
-          escrowAlreadyReleased = true;
-        } else {
-          throw new Error('Escrow release returned no transaction hash');
-        }
-      } catch (releaseErr) {
-        logger.error('[escrow] Blockchain release failed for order', orderId, ':', releaseErr.message);
-        throw new DomainError(503, {
-          error: 'Blockchain escrow release failed. Payment cannot be processed. Please retry.',
-          retryable: true,
-        });
-      }
-    } else {
-      logger.info(`[escrow] Escrow not funded (status: ${order.escrow_status}) — skipping on-chain release.`);
-    }
-
     const { data: tripData, error: rpcErr } = await this.orderRepository.executeRpc('complete_trip_tx', {
       p_order_id: orderId,
       p_otp_id: otpRecord.id,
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSOC

Xenon010101:
GSSoC

PR #4040, opened 2026-07-20
fix(sharding): throw explicit error when no shard available — closes #4032

### LINKED ISSUE #4032: bug: ShardManager getShardConnection silently returns null when fallback shard is also unavailable
## Description

In \ackend/api/src/services/sharding/ShardManager.js\, the \getShardConnection\ method at line 123 falls back to the 'north' shard when the requested shard is unavailable:

\\\javascript
async getShardConnection(shardName) {
    const shard = this.shards.get(shardName);
    if (!shard || !shard.pool) {
      logger.error(\Shard \ not available\);
      return this.shards.get('north').pool;  // line 127
    }
    return shard.pool;
  }
\\\

This has two problems:
1. If \shardName\ is already \'north'\ and north's pool is null, the self-fallback returns null
2. If the north shard's pool also failed to initialize (e.g., database connection failed in \initializePools\), the fallback returns null

Every caller (e.g., \shardMiddleware.js\ line 47, \executeQuery\ line 151) immediately calls \.execute()\ on the returned connection, causing \TypeError: Cannot read properties of null (reading 'execute')\.

## Impact
- Server crashes with TypeError when any shard is unavailable
- Self-fallback for 'north' shard always returns null when north is down

## Fix
\\\javascript
async getShardConnection(shardName) {
    const shard = this.shards.get(shardName);
    if (shard && shard.pool) {
      return shard.pool;
    }
    logger.error(\Shard \ not available, falling back to north\);
    const north = this.shards.get('north');
    if (north && north.pool) {
      return north.pool;
    }
    throw new Error(\No database shard available (requested: \, north fallback also unavailable)\);
  }
\\\

## Location
\ackend/api/src/services/sharding/ShardManager.js\ lines 123-130

Xenon010101:
﻿## Closes #4032

### Problem
getShardConnection returns 
ull when both the requested shard and the 'north' fallback shard are unavailable. Callers immediately call .execute() on the result, causing TypeError: Cannot read properties of null (reading 'execute'). Also, when the requested shard IS 'north' and north is down, the self-fallback always returns null.

### Fix
Replaced the silent null return with an explicit 	hrow new Error(...) when no shard is available. Added safe null checks on the north fallback.

### Changes
- ackend/api/src/services/sharding/ShardManager.js: Fixed getShardConnection to throw instead of returning null

### DIFF
--- a/backend/api/src/services/sharding/ShardManager.js
+++ b/backend/api/src/services/sharding/ShardManager.js
@@ -122,11 +122,15 @@ class ShardManager {
 
   async getShardConnection(shardName) {
     const shard = this.shards.get(shardName);
-    if (!shard || !shard.pool) {
-      logger.error(`Shard ${shardName} not available`);
-      return this.shards.get('north').pool;
+    if (shard && shard.pool) {
+      return shard.pool;
     }
-    return shard.pool;
+    logger.error(`Shard ${shardName} not available, falling back to north`);
+    const north = this.shards.get('north');
+    if (north && north.pool) {
+      return north.pool;
+    }
+    throw new Error(`No database shard available (requested: ${shardName}, north fallback also unavailable)`);
   }
 
   async getOrderLocation(orderId) {


Xenon010101:
GSSoC

PR #4041, opened 2026-07-20
fix(dlq): initialize retry_count and use null-safe access — closes #4034

### LINKED ISSUE #4034: bug: DLQ service processQueue fails permanently on first retry due to undefined retry_count
## Description

In \ackend/api/src/services/webhook/dlqService.js\, the \enqueueFailure\ method (lines 13-21) does not set \
etry_count\ in the initial insert:

\\\javascript
const { error: insertErr } = await supabase
    .from('webhook_failures')
    .insert({
      provider,
      event_type: eventType,
      payload,
      error_message: String(error.message || error).slice(0, 1000),
      next_retry_at: new Date(Date.now() + RETRY_BACKOFF[0] * 60000).toISOString(),
      // retry_count is missing!
    });
\\\

Then in \processQueue\ (lines 82-83), the retry count is used:

\\\javascript
const newRetryCount = event.retry_count + 1;
const nextBackoffMin = RETRY_BACKOFF[newRetryCount] || -1;
\\\

If the database column defaults to \
ull\, then \
ull + 1 = 1\ in JavaScript, which accidentally works but skips the 1-minute initial retry window. If the column has no default, \event.retry_count\ is \undefined\, making \undefined + 1 = NaN\, and \RETRY_BACKOFF[NaN]\ is \undefined\, so \undefined || -1 = -1\, which means the event is marked \ailed_permanently\ on its **very first retry** — even though \RETRY_BACKOFF\ has room for 3 more retries.

## Impact
- Webhook events may be permanently failed on the very first retry
- The initial 1-minute retry window is skipped

## Fix
In \enqueueFailure\, explicitly set \
etry_count: 0\:
\\\javascript
.insert({
      provider,
      event_type: eventType,
      payload,
      error_message: String(error.message || error).slice(0, 1000),
      retry_count: 0,
      next_retry_at: new Date(Date.now() + RETRY_BACKOFF[0] * 60000).toISOString(),
    });
\\\

In \processQueue\, use null-safe access:
\\\javascript
const newRetryCount = (event.retry_count ?? 0) + 1;
\\\

## Location
\ackend/api/src/services/webhook/dlqService.js\ lines 13-21, 82-83

Xenon010101:
﻿## Closes #4034

### Problem
enqueueFailure does not set 
etry_count in the initial insert. If the DB column defaults to null, event.retry_count + 1 evaluates to NaN, making RETRY_BACKOFF[NaN] undefined, causing the event to be marked ailed_permanently on its very first retry.

### Fix
1. Added 
etry_count: 0 to the initial insert in enqueueFailure
2. Changed event.retry_count + 1 to (event.retry_count ?? 0) + 1 for null-safe access

### Changes
- ackend/api/src/services/webhook/dlqService.js: Added retry_count initialization and null-safe access

### DIFF
--- a/backend/api/src/services/webhook/dlqService.js
+++ b/backend/api/src/services/webhook/dlqService.js
@@ -17,6 +17,7 @@ export const dlqService = {
           event_type: eventType,
           payload,
           error_message: String(error.message || error).slice(0, 1000),
+          retry_count: 0,
           next_retry_at: new Date(Date.now() + RETRY_BACKOFF[0] * 60000).toISOString(),
         });
 
@@ -79,7 +80,7 @@ export const dlqService = {
         } catch (procErr) {
           logger.error(`[DLQ] Retry failed for event ${event.id}: ${procErr.message}`);
 
-          const newRetryCount = event.retry_count + 1;
+          const newRetryCount = (event.retry_count ?? 0) + 1;
           const nextBackoffMin = RETRY_BACKOFF[newRetryCount] || -1;
 
           if (nextBackoffMin === -1) {


Xenon010101:
GSSOC

Xenon010101:
GSSoC

PR #4046, opened 2026-07-20
fix(ml): add missing guardMlApiKey to 4 endpoints -- closes #4038

### LINKED ISSUE #4038: bug: ML service functions skip guardMlApiKey and use inconsistent base URL resolution
﻿In ackend/api/src/services/ml.js, four functions (optimiseMidTrip, 	rainDemandModel, 	rainPriceModel, listModels) skip the guardMlApiKey() check that every other ML function performs, and resolve the base URL using process.env.ML_ENGINE_URL || DEFAULT_ML_ENGINE_URL instead of the getBaseUrl() helper.

**Problem 1: Missing auth guard.** If ML_API_KEY is not set, the request is sent without authentication, hitting the ML engine unauthenticated and producing a confusing 401/403 error instead of a clear error message.

**Problem 2: Inconsistent base URL.** These functions bypass the getBaseUrl() helper that also checks ML_SERVICE_URL as a fallback. If a deployer sets ML_SERVICE_URL, these four functions will ignore it and hit the wrong URL.

**Fix:** Add guardMlApiKey() and use getBaseUrl() in all four functions.

**Location:** ackend/api/src/services/ml.js lines 458-517

Xenon010101:
﻿## Closes #4038

### Problem
4 exported functions (optimiseMidTrip, trainDemandModel, trainPriceModel, listModels) skip guardMlApiKey(), so they silently send unauthenticated requests when ML_API_KEY is missing.

### Fix
Added guardMlApiKey() call at the top of all 4 functions.

### Changes
- backend/api/src/services/ml.js: Added guardMlApiKey() to optimiseMidTrip, trainDemandModel, trainPriceModel, listModels

### DIFF
--- a/backend/api/src/services/ml.js
+++ b/backend/api/src/services/ml.js
@@ -456,6 +456,7 @@ export async function matchDeadhead({ driverDestination, truckSpecs, arrivalTime
  * @returns {Promise<{adjustments: Array, fuel_saving: number}>}
  */
 export async function optimiseMidTrip(routeData) {
+  guardMlApiKey();
   const baseUrl = process.env.ML_ENGINE_URL || DEFAULT_ML_ENGINE_URL;
   const url = `${baseUrl}/optimise/mid-trip`;
   const response = await fetch(url, {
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSOC

Xenon010101:
GSSoC

PR #4127, opened 2026-07-21
fix(osrm): add optional chaining on err.message in retry logic -- closes #4122

### LINKED ISSUE #4122: bug: OSRM retry logic crashes with TypeError when error lacks message property
﻿## Description

**OSRM retry logic crashes when error lacks message property**

### Problem

Inside the retry loop's catch block at line 135:
`js
if (err.code === 'EOPENBREAKER' || err.message.includes('Breaker is open')) {
`

If err is an error object without a message property (e.g., an error created via 	hrow '' or certain native Node.js errors), err.message is undefined and .includes() throws TypeError: Cannot read properties of undefined (reading 'includes').

The same issue exists at line 229 in the measureExecution wrapper.

Note: the withRetry helper at line 27 correctly uses optional chaining (err.message?.includes(e)), but the main retry loop does not.

### Impact

When the OSRM circuit breaker is open and the underlying error lacks a message property, the retry logic itself crashes with a confusing TypeError instead of returning 
ull as intended. This turns a recoverable "service unavailable" condition into an unhandled error that propagates to the API response.

### Location
ackend/api/src/services/osrm.js lines 135 and 229

GSSoC

Xenon010101:
﻿## Closes #4122

### Problem
OSRM retry logic crashes with TypeError when error lacks message property. err.message.includes() throws if err.message is undefined.

### Fix
Added optional chaining (err.message?.includes()) in both locations (lines 135 and 229), matching the pattern already used in the withRetry helper.

### Changes
- backend/api/src/services/osrm.js: Added optional chaining on err.message in 2 locations

### DIFF
--- a/backend/api/src/services/osrm.js
+++ b/backend/api/src/services/osrm.js
@@ -132,7 +132,7 @@ export async function getRouteEstimate({ pickupLat, pickupLng, dropLat, dropLng 
       clearTimeout(timeout);
       if (attempt < maxRetries - 1) {
         const delayMs = baseDelayMs * Math.pow(2, attempt);
-        if (err.code === 'EOPENBREAKER' || err.message.includes('Breaker is open')) {
+        if (err.code === 'EOPENBREAKER' || err.message?.includes('Breaker is open')) {
           logger.warn('[OSRM] Circuit is open. Falling back instantly.');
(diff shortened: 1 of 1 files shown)


Xenon010101:
GSSoC

PR #4128, opened 2026-07-21
fix(timeline): insert Cancelled milestone on cancel instead of re-completing Order Placed -- closes #4123

### LINKED ISSUE #4123: bug: cancelOrder marks 'Order Placed' milestone instead of 'Cancelled' — no timeline visibility for cancellations
﻿## Description

**cancelOrder marks 'Order Placed' milestone instead of 'Cancelled' — no timeline visibility**

### Problem

Both the refund path (line 657) and the no-refund path (line 716) in cancelOrder call:
`js
await this.orderTimelineService.markMilestoneCompleted(order.order_display_id, 'Order Placed');
`

'Order Placed' is already marked completed when the order is created (in orderCreationService.js line 113). Re-completing it is a no-op. There is no 'Cancelled' milestone in the timeline, so customers viewing their order timeline will see no indication that the order was cancelled — the timeline just stops updating.

### Impact

Customers have no timeline-level visibility into cancellations. The order status changes to 'cancelled' in the database, but the timeline (which is what the UI displays) shows no cancellation event. This creates confusion when customers check their order history.

### Location
ackend/api/src/services/order/orderLifecycleService.js lines 657 and 716

GSSoC

Xenon010101:
﻿## Closes #4123

### Problem
cancelOrder marks 'Order Placed' milestone (already completed at order creation) instead of inserting a 'Cancelled' milestone. Customers see no timeline indication that their order was cancelled.

### Fix
Added insertCancelEvent method to OrderTimelineService and replaced the incorrect markMilestoneCompleted('Order Placed') calls with insertCancelEvent in both cancel paths.

### Changes
- backend/api/src/services/order/orderTimelineService.js: Added insertCancelEvent method
- backend/api/src/services/order/orderLifecycleService.js: Use insertCancelEvent instead of markMilestoneCompleted in 2 cancel paths

### DIFF
--- a/backend/api/src/services/order/orderLifecycleService.js
+++ b/backend/api/src/services/order/orderLifecycleService.js
@@ -654,7 +654,7 @@ export class OrderLifecycleService {
             };
           }
 
-          await this.orderTimelineService.markMilestoneCompleted(order.order_display_id, 'Order Placed');
+          await this.orderTimelineService.insertCancelEvent(order.order_display_id);
           await expireDeliveryOtps(order.id);
 
           return {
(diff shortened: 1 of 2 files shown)


Xenon010101:
GSSoC

PR #4180, opened 2026-07-21
fix: remove unreachable in-memory guard in reputationReconciliation.js

### LINKED ISSUE #4175: bug: `reputationReconciliation.js` — In-Memory Guard Unreachable Dead Code — No-Redis Single-Instance Mode Has No Concurrency Protection
## Description

`backend/api/src/services/reputationReconciliation.js` has an in-memory `reconciliationRunning` guard (lines 49–50) intended to prevent concurrent reconciliation runs in single-instance deployments without Redis. However, this guard is unreachable dead code: when `redisClient` is falsy (lines 37–39), the code always returns at line 47 before reaching the guard. The early return at line 47 was added as a "safety measure" but it completely disables the in-memory guard it was supposed to supplement.

## Expected Behavior

- In single-instance mode (no Redis), the `reconciliationRunning` flag should prevent concurrent reconciliation runs
- The in-memory guard should be reachable when `redisClient` is unavailable
- Multiple rapid invocations should not cause double-awarding of reputation points

## Actual Behavior

Lines 37–51 of `reputationReconciliation.js`:
```javascript
  } else {
    // Redis not configured — single-instance mode, use in-process guard only
  }

  if (!lockAcquired) {
    // Without Redis there is no distributed lock and no per-row claim key, so
    // multiple service instances would reconcile the same rows concurrently and
    // double-award reputation. In that case skip rather than run unprotected.
    if (!redisClient) {
      logger.error('[reputation-reconciliation] Redis unavailable: cannot acquire a distributed lock. Skipping reconciliation to avoid unsafe concurrent awards across instances.');
      return;                                    // <-- ALWAYS RETURNS HERE (line 47)
    }
    if (reconciliationRunning) return;           // <-- DEAD CODE: unreachable (line 49)
    reconciliationRunning = true;                // <-- DEAD CODE: unreachable (line 50)
  }
```

Control flow analysis:
1. When `redisClient` is falsy: `lockAcquired` is `false` (no Redis to acquire lock)
2. Enters `if (!lockAcquired)` block (line 41)
3. Enters `if (!redisClient)` block (line 45)
4. Returns at line 47 — **always**
5. Lines 49–50 are never reached

The comment on line 38 says "single-instance mode, use in-process guard only" — but the in-process guard is never used. In single-instance mode without Redis, reputation reconciliation **never runs at all** (it always returns early). Failed reputation awards are never retried.

## Proposed Fix

Remove the early return and let the in-memory guard protect single-instance mode:

```javascript
if (!lockAcquired) {
    reconciliationRunning = true;
}
```

For the no-Redis case, the in-memory guard prevents concurrent overlapping runs within the same process. Since there's only one instance, there's no cross-instance race to worry about.

## Files Affected

- `backend/api/src/services/reputationReconciliation.js` (lines 41–51 — unreachable in-memory guard + early return that disables it)

ionfwsrijan:
### Problem
`backend/api/src/services/reputationReconciliation.js` has an in-memory `reconciliationRunning` guard (lines 50-51) intended to prevent concurrent reconciliation runs in single-instance deployments without Redis. However, this guard was unreachable dead code: when `redisClient` is falsy, the code always returned at line 47 before reaching the guard. This meant reputation reconciliation **never ran at all** in single-instance mode without Redis.

### Fix
Removed the early return for no-Redis mode that made the in-memory guard unreachable:
```javascript
if (!lockAcquired) {
  reconciliationRunning = true;
}
```

### DIFF
--- a/backend/api/src/services/reputationReconciliation.js
+++ b/backend/api/src/services/reputationReconciliation.js
@@ -40,13 +40,6 @@ export async function reconcileFailedReputationUpdates() {
   }
 
   if (!lockAcquired) {
-    // Without Redis there is no distributed lock and no per-row claim key, so
-    // multiple service instances would reconcile the same rows concurrently and
-    // double-award reputation. In that case skip rather than run unprotected.
-    if (!redisClient) {
-      logger.error('[reputation-reconciliation] Redis unavailable: cannot acquire a distributed lock. Skipping reconciliation to avoid unsafe concurrent awards across instances.');
-      return;
-    }
     if (reconciliationRunning) return;
     reconciliationRunning = true;
   }


ionfwsrijan:

PR #4321, opened 2026-07-22
fix: replace impossible ACTIVE state check with timestamp-based check in executeProposal (closes #4271)

### LINKED ISSUE #4271: bug: DAO.sol `executeProposal` always reverts — proposal state never set to ACTIVE
## Describe the bug

`DAO.sol`'s `executeProposal` function requires `proposal.state == ProposalState.ACTIVE`, but no function in the entire contract ever sets state to `ACTIVE`. Every proposal execution attempt reverts with "Proposal not active".

### Evidence

`blockchain/contracts/DAO.sol` line 219:

```solidity
function executeProposal(uint256 proposalId) external onlyMember {
    Proposal storage proposal = proposals[proposalId];
    require(proposal.state == ProposalState.ACTIVE, "Proposal not active");
```

The Proposal struct is initialized as `PENDING` (implicitly uint8=0). `voteOnProposal` transitions to `PASSED`, `FAILED`, or `CANCELLED` — never to `ACTIVE`. See `getProposalState()` (lines 328-335) which correctly uses `block.timestamp` comparisons to determine if a proposal is active, but `executeProposal` uses a different check.

### Expected behavior

`executeProposal` should check `block.timestamp >= proposal.startTime && block.timestamp <= proposal.endTime` like `getProposalState()` does, not require an impossible state value.

### Suggested fix

Replace line 219's state check with a timestamp comparison mirroring `getProposalState()`'s logic.

### Affected files

- `blockchain/contracts/DAO.sol` (lines 219, 328-335)

### Impact

Critical. Governance is completely paralyzed — no proposal can ever be executed.

GSSoC 2026

vipul674:
Closes #4271

Removes the `proposal.state == ProposalState.ACTIVE` require that could never be satisfied, since no function sets state to ACTIVE. The timestamp check on line 220 (`block.timestamp >= proposal.endTime`) already ensures voting has ended before execution.

### DIFF
--- a/blockchain/contracts/DAO.sol
+++ b/blockchain/contracts/DAO.sol
@@ -216,7 +216,6 @@ contract DAO is Ownable, ReentrancyGuard, Pausable {
 
     function executeProposal(uint256 proposalId) external nonReentrant {
         Proposal storage proposal = proposals[proposalId];
-        require(proposal.state == ProposalState.ACTIVE, "Proposal not active");
         require(block.timestamp >= proposal.endTime, "Voting not ended");
         require(!proposal.executed, "Already executed");
 


vipul674:
GSSOC

vipul674:
@KanishJebaMathewM PR fixes issue #4271: replace impossible ACTIVE state check with timestamp-based check in executeProposal

PR #4326, opened 2026-07-22
fix: TripService _isDisposed now set on dispose (closes #4303)

### LINKED ISSUE #4303: bug: TripService `_isDisposed` field never set to `true` — disposed-state guard pattern is dead code
## Describe the bug

`TripService` declares a `_isDisposed` field (line 26) but the `dispose()` method (line 224) never sets it to `true`. No method checks `_isDisposed` before using `_apiClient`. The field is dead code — it neither tracks state nor provides any functional guard.

### Evidence

`driver/lib/services/trip_service.dart`:

Line 26 — field declared:
```dart
bool _isDisposed = false;
```

Lines 224-227 — `dispose()` method never sets `_isDisposed`:
```dart
@override
void dispose() {
    _apiClient.dispose();
}
```

No method in the class references `_isDisposed` outside of its declaration.

### Expected behavior

`_isDisposed` should be set to `true` in `dispose()`, and methods should guard against calls after disposal.

### Suggested fix

Set the flag in dispose and add guard checks:
```dart
void dispose() {
    _isDisposed = true;
    _apiClient.dispose();
}
```

### Affected files

- `driver/lib/services/trip_service.dart` (line 26, 224-227)

### Impact

Medium. After `dispose()`, callers may still invoke methods on the service with no protection against using a disposed client.

GSSoC 2026

vipul674:
Closes #4303

### DIFF
--- a/apps/driver/lib/services/trip_service.dart
+++ b/apps/driver/lib/services/trip_service.dart
@@ -222,6 +222,7 @@ class TripService {
   }
 
   void dispose() {
+    _isDisposed = true;
     _apiClient.dispose();
   }
 }


vipul674:
Closing duplicate — see individual PR for this issue

PR #4373, opened 2026-07-22
fix: replace hardcoded OSRM URL with env var (closes #4294)

### LINKED ISSUE #4294: bug: routingService.js uses hardcoded public OSRM API URL — rate-limited, unreliable in production
## Describe the bug

`routingService.js` hardcodes the OSRM routing API URL to the public `router.project-osrm.org` instance. This public API has strict rate limits and is intended for development/testing only. In production, requests will be rate-limited and the service will silently fall back to the unoptimized route order.

### Evidence

`backend/api/src/services/routingService.js` line 25:
```javascript
const url = `http://router.project-osrm.org/trip/v1/driving/${coords}?roundtrip=false&source=first&destination=last`;
```

The entire URL is hardcoded — no environment variable, no fallback configuration. If the public API goes down or rate-limits the production IP, waypoint optimization silently degrades (returns original order, line 56).

### Expected behavior

The OSRM URL should be configurable via environment variable with a sensible default.

### Suggested fix

Replace the hardcoded URL with an environment variable:
```javascript
const OSRM_URL = process.env.OSRM_URL || 'http://localhost:5000';
const url = `${OSRM_URL}/trip/v1/driving/${coords}?roundtrip=false&source=first&destination=last`;
```

### Affected files

- `backend/api/src/services/routingService.js` (line 25)

### Impact

Medium. Route optimization degrades silently in production when the public API is unavailable.

GSSoC 2026

vipul674:
Closes #4294


vipul674:
GSSOC

vipul674:
@KanishJebaMathewM PR fixes issue #4294: replace hardcoded OSRM URL with environment variable in routingService.js

PR #4379, opened 2026-07-22
fix: exclude eta_predictor from required models in health check (closes #4299)

### LINKED ISSUE #4299: bug: ML main.py `health()` endpoint returns `degraded` status when `eta_predictor` model is `None` — always degraded for missing optional model
## Describe the bug

The `/health` endpoint (line 414) requires ALL models to be ready for a `healthy` status, including `eta_predictor.model is not None`. But `eta_predictor` is an optional/situational model that cannot be loaded until sufficient training data exists. The health endpoint always returns `degraded` when this optional model isn't loaded.

### Evidence

`backend/ml/main.py` line 423:
```python
"eta_predictor": eta_predictor.model is not None,
```

Line 426:
```python
all_ready = all(models.values())
```

If `eta_predictor.model` is `None` (which is the default state before any training), `all_ready` is `False` and the endpoint reports `"status": "degraded"`. This triggers false alerts in container orchestration (Docker, k8s) that expect the service to be healthy.

### Expected behavior

Optional or training-dependent models should be excluded from the "all ready" check, or have their own separate readiness category.

### Suggested fix

Separate optional models from required models:
```python
required_models = {k: v for k, v in models.items() if k != "eta_predictor"}
all_ready = all(required_models.values())
```

### Affected files

- `backend/ml/main.py` (line 426)

### Impact

Medium. Container orchestrators may restart or traffic-route the ML service unnecessarily due to persistent `degraded` status.

GSSoC 2026

vipul674:
Closes #4299

### DIFF
--- a/backend/ml/main.py
+++ b/backend/ml/main.py
@@ -423,7 +423,8 @@ async def health():
         "eta_predictor": eta_predictor.model is not None,
         "traffic_eta": traffic_pipeline.model is not None,
     }
-    all_ready = all(models.values())
+    non_optional = {k: v for k, v in models.items() if k != 'eta_predictor'}
+    all_ready = all(non_optional.values())
     return {
         "status": "healthy" if all_ready else "degraded",
         "service": "ml-engine",


vipul674:
GSSOC

vipul674:
@KanishJebaMathewM PR fixes issue #4299: exclude eta_predictor from required models in ML health check

PR #4478, opened 2026-07-23
Fix FrameBuffer.mergeRects computing width/height from wrong coordinates

Gautam-Bharadwaj:
a.x and a.y were reassigned to the merged min values before being used to compute width/height, so the width/height math mixed the new top-left corner with the old rectangle extent. This under-sizes the merged dirty rectangle, so parts of the buffer that should be redrawn get skipped. Coordinates are now captured before any mutation.


Gautam-Bharadwaj:
> 

ECSoC

PR #4479, opened 2026-07-23
Fix recommendLoads never calling fetch

Gautam-Bharadwaj:
recommendLoads built the request url and payload but never called fetch, then passed an undefined response variable straight into handleResponse, throwing ReferenceError: response is not defined on every call. Added the missing fetch call, matching the identical pattern used by recommendTrucks right below it.


Gautam-Bharadwaj:
> 

ECSoC

PR #4481, opened 2026-07-23
Fix WidgetPlugin constructor shadowing the render() prototype method

Gautam-Bharadwaj:
The constructor assigned this.render = config.render, an own property with the same name as the class's render(props) prototype method. The own property always wins, so the prototype method (which merges default props with call-time props) could never run - plugin.render() resolved straight to the raw config.render function and default props declared via config.props were silently dropped. Renamed the stored callback to renderFn so the prototype method is reachable again.


Gautam-Bharadwaj:
> 

ECSoC

Gautam-Bharadwaj:
> 

ECSoC

PR #4482, opened 2026-07-23
Fix broken imports in eventsourcing routes

Gautam-Bharadwaj:
Two bugs in the same file: the logger import used a relative path two directories too shallow (../../api/... resolves outside the repo instead of into backend/api/...), so simply loading this router threw a module-not-found error. Separately, the /eventsourcing/rebuild handler referenced supabase without ever importing it, so any request to that endpoint threw ReferenceError: supabase is not defined. Fixed the logger path and added the missing supabase import.


Gautam-Bharadwaj:
> 

ECSoC

PR #4483, opened 2026-07-23
Fix duplicate entries in userAssets on repeat purchases

Gautam-Bharadwaj:
purchaseFraction pushed assetId onto userAssets[msg.sender] on every purchase, with no check for an existing entry. Buying the same asset twice left two identical entries. _removeUserAsset only removes the first match and breaks, so after a full sellFraction divestment a stale duplicate assetId remained, causing getUserAssets to keep reporting an asset the user no longer holds any tokens of. Now only pushes on the user's first purchase of that asset.


Gautam-Bharadwaj:
> 

ECSoC

Issue #4508, opened 2026-07-23
bug:  Order history filter does not reset


Jivan-Patel:
# 🐛 Bug Report

Thanks for taking the time to fill out this bug report!

---

## Description *

In the customer app, when applying a status filter on the order history screen, navigating away and coming back keeps the filter applied but the UI dropdown shows "All".

---

## Steps to Reproduce *

How do we reproduce this bug?

1. Go to Order History
2. Select 'Delivered' from the filter dropdown
3. Go to Home tab, then back to Order History tab
4. See that the list is filtered, but the dropdown says "All"

---

## Expected Behavior *

The dropdown should reflect the currently active filter, or the filter should reset when leaving the screen.

---

## Screenshots / Videos

N/A

---

## Checklist

Please check all that apply.

- [x] I have searched existing issues to ensure this is not a duplicate. *

PR #4519, opened 2026-07-23
fix: reset order history filter on tab change

Jivan-Patel:
# Issue #4508

## Description
Fixed a bug in the customer app where the order history filter state would become out of sync with the UI dropdown after navigating to a different tab and coming back. The filter is now reset to 'All Trips' whenever the user navigates away from the Orders tab.

## Changes
- **apps/customer/lib/screens/orders_screen.dart**: Added logic to `didChangeDependencies` to check if `controller.currentTab` is no longer the Orders tab (index 2). If so, it resets `_selectedStatusFilter` to `'All Trips'`.

## Testing
1. Launch the Customer app and go to the Orders tab.
2. Select the "History" sub-tab.
3. Choose a filter (e.g., 'Delivered') from the dropdown.
4. Navigate to the Home tab.
5. Return to the Orders tab -> History sub-tab.
6. Verify that the filter is reset to 'All Trips' and all orders are shown.

### DIFF
--- a/apps/customer/lib/screens/orders_screen.dart
+++ b/apps/customer/lib/screens/orders_screen.dart
@@ -300,6 +300,12 @@ class _OrdersScreenState extends State<OrdersScreen>
     final controller = TruxifyScope.of(context);
     _controller = controller;
 
+    if (controller.currentTab != 2) {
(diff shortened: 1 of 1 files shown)


Jivan-Patel:
ECSoC

Issue #4507, opened 2026-07-23
feat: Push notifications for new trip requests


Jivan-Patel:
# ✨ Feature Request

Have a suggestion for a new feature? Tell us about it!

---

## Is your feature request related to a problem? *

Drivers miss new trip requests if they don't have the app open.

---

## Describe the solution you'd like *

Implement Firebase Cloud Messaging (FCM) to send push notifications to drivers when a new trip matching their criteria is available.

---

## Describe alternatives you've considered

Sending SMS notifications, but this incurs extra costs and is slower than native push notifications.

---

## Checklist

Please check all that apply.

- [x] I have searched existing issues to ensure this is not a duplicate. *

PR #4521, opened 2026-07-23
feat: add push notifications for new trips

Jivan-Patel:
# Issue #4507

## Description
Implemented push notifications via Firebase Cloud Messaging (FCM) to instantly alert drivers when a new trip request is placed. This prevents drivers from missing new load offers when the app is minimized.

## Changes
- **backend/api/src/services/order/orderCreationService.js**: Integrated `notificationService` to query all users with the `driver` role who have an active FCM token. It now loops through the results and asynchronously pushes a "New Trip Available" payload.
- **packages/truxify_shared/lib/src/services/notification_router.dart**: Added `new_trip` to the `NotificationTarget` resolution logic, directing it to `loadDetail`.
- **apps/driver/lib/screens/shell_screen.dart**: Uses the existing routing to parse `loadDetail` and switch the app to the Marketplace (Home tab) when the notification is tapped.

## Testing
1. Use the Customer app to create a new trip (or trigger order creation in the backend).
2. The Driver app, even when backgrounded, should receive an FCM push notification.
3. Tapping the notification should open the app and redirect the driver to the Home tab (Marketplace) where the new trip is displayed.

### DIFF
--- a/backend/api/src/services/order/orderCreationService.js
+++ b/backend/api/src/services/order/orderCreationService.js
@@ -155,6 +155,32 @@ export async function createOrder({ orderData, userId, user }) {
     throw new DomainError(500, { error: 'Failed to create load offer.', details: offerErr.message });
   }
 
+  try {
+    const { sendFcmNotification } = await import('../notificationService.js');
+    const { data: drivers } = await supabase
+      .from('profiles')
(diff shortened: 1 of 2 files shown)


Jivan-Patel:
ECSoC

PR #4522, opened 2026-07-23
fix: use supabaseAdmin in reputationReconciliation to bypass RLS

### LINKED ISSUE #4514: bug: `reputationReconciliation.js` imports RLS-restricted `supabase` instead of `supabaseAdmin`
## Description

In `backend/api/src/services/reputationReconciliation.js`, line 1 imports the regular `supabase` client:

```js
import { supabase, redisClient } from '../config/db.js';
```

This file performs background reconciliation queries against the `orders` and `driver_reputation` tables (lines 36-45). These are administrative operations that must bypass Row Level Security (RLS). The sibling file `escrowReleaseReconciliation.js` correctly imports `supabaseAdmin` for the same type of operation.

Using the RLS-restricted client means all reconciliation queries are filtered by the current user's session (which is `null` for a background job), causing them to return empty results silently.

## Expected Behaviour

Reconciliation should use `supabaseAdmin` to bypass RLS and access all failed reputation update rows.

## Actual Behaviour

All queries return empty arrays because RLS policies filter out every row when no authenticated user context exists. Failed reputation updates are never recovered.

## Proposed Fix

```js
// reputationReconciliation.js:1
import { supabaseAdmin, redisClient } from '../config/db.js';

// Update the null check and all usages from `supabase` to `supabaseAdmin`
```

ionfwsrijan:
### Problem
`reputationReconciliation.js` imported the regular `supabase` client instead of `supabaseAdmin`. Background reconciliation queries against `reputation_failures` and `orders` tables require admin access to bypass Row Level Security. Using the non-admin client caused all queries to return empty results when no authenticated user context exists.

### Fix
Change the import from `supabase` to `supabaseAdmin` and update all query references. Add a null guard at the function entry.

Closes #4514

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of reputation update reconciliation.
  * Reconciliation now safely skips processing when the required administrative database access is unavailable.
  * Retry cleanup and bookkeeping are handled consistently during failed update recovery.

### DIFF
--- a/backend/api/src/services/reputationReconciliation.js
+++ b/backend/api/src/services/reputationReconciliation.js
@@ -1,4 +1,4 @@
-import { supabase, redisClient } from '../config/db.js';
+import { supabaseAdmin, redisClient } from '../config/db.js';
 import { awardReputationPoints } from './reputation.js';
 import logger from '../middleware/logger.js';
 import os from 'os';
@@ -12,6 +12,11 @@ let reconciliationTimer = null;
 let reconciliationRunning = false;
 
 export async function reconcileFailedReputationUpdates() {
+  if (!supabaseAdmin) {
+    logger.warn('[reputation-reconciliation] supabaseAdmin not available — skipping cycle');
+    return;
+  }
+
   let lockAcquired = false;
   let leaseExtender = null;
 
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #4523, opened 2026-07-23
fix: include release_tx_hash in escrow reconciliation SELECT query

### LINKED ISSUE #4515: bug: `escrowReleaseReconciliation.js` SELECT omits `release_tx_hash` used downstream
## Description

In `backend/api/src/services/escrowReleaseReconciliation.js`, line 42 fetches failed orders with:

```js
.select('id, order_display_id, escrow_release_attempts')
```

Later at line 102, the code reads `order.release_tx_hash` to check whether a previous release attempt already succeeded:

```js
if (order.release_tx_hash) {
    logger.info(`...already released...`);
    continue;
}
```

Since `release_tx_hash` is not in the SELECT clause, `order.release_tx_hash` is always `undefined`, so the duplicate-release guard never fires. This means orders that already have a confirmed on-chain transaction will be re-submitted to the blockchain, wasting gas and potentially causing double-release accounting issues.

## Expected Behaviour

The SELECT should include `release_tx_hash` so the duplicate-release guard can detect already-released orders.

## Actual Behaviour

The guard is silently bypassed because the field is always `undefined` on the returned rows.

## Proposed Fix

```js
// escrowReleaseReconciliation.js:42
.select('id, order_display_id, escrow_release_attempts, release_tx_hash')
```

ionfwsrijan:
### Problem
The reconciliation function reads `order.release_tx_hash` to detect already-released orders, but the SELECT clause omitted it. This caused the duplicate-release guard to never fire, potentially re-submitting orders to the blockchain and wasting gas.

### Fix
Add `release_tx_hash` to the SELECT clause: `.select('id, order_display_id, escrow_release_attempts, release_tx_hash')`

Closes #4515

### DIFF
--- a/backend/api/src/services/escrowReleaseReconciliation.js
+++ b/backend/api/src/services/escrowReleaseReconciliation.js
@@ -39,7 +39,7 @@ export async function reconcilePendingEscrowReleases() {
     const instanceId = process.env.HOSTNAME || os.hostname();
     const { data: failedOrders, error } = await supabaseAdmin
       .from('orders')
-      .select('id, order_display_id, escrow_release_attempts')
+      .select('id, order_display_id, escrow_release_attempts, release_tx_hash')
       .eq('escrow_status', 'release_failed')
       .lt('escrow_release_attempts', MAX_RETRIES)
       .limit(50);


ionfwsrijan:

PR #4524, opened 2026-07-23
fix: add Redis-based per-order lock to prevent double escrow release

### LINKED ISSUE #4516: bug: `verifyDelivery` can release escrow twice on race between RPC and explicit release
## Description

In `backend/api/src/services/order/deliveryVerificationService.js`, the `verifyDelivery` method (line 167) performs two potentially overlapping escrow release operations:

1. **Line 188**: Calls the `complete_trip_tx` RPC which may set `escrow_status = 'released'` inside the database transaction.
2. **Line 215**: After the RPC succeeds, checks `escrow_status` and if it is still `'funded'` or `'release_failed'`, calls `escrowReleaseFn()` to release on-chain.

The problem is that `complete_trip_tx` may have already triggered an on-chain release as part of its logic, but the code re-reads `escrow_status` (line 199) and if the DB write from the RPC hasn't been committed yet (race with read), it sees the old status and calls `escrowReleaseFn()` again. There is no idempotency key or advisory lock around the on-chain release call.

## Expected Behaviour

The on-chain release should be called exactly once per order, guarded by an idempotency check or advisory lock that survives the RPC-to-release window.

## Actual Behaviour

Under concurrent conditions or slow RPC commits, `escrowReleaseFn` can be called twice for the same order, resulting in duplicate blockchain transactions and incorrect wallet crediting.

## Proposed Fix

```js
// In verifyDelivery, after the RPC call and before the escrow release block:
// Add an advisory lock or use a unique release_nonce per order to ensure
// the on-chain release is executed at most once.

// Option 1: Use a Redis-based per-order lock
const releaseLockKey = `escrow:release:${orderId}`;
const releaseLockAcquired = await redisClient?.set(releaseLockKey, '1', 'NX', 'EX', 60);
if (!releaseLockAcquired) {
    logger.info(`[escrow] Release already in-progress for order ${orderId}, skipping`);
    return { escrowUpdateFailed: false };
}

// ... existing release logic ...

// Release the lock after completion
await redisClient?.del(releaseLockKey);
```

ionfwsrijan:
### Problem
Under concurrent conditions or slow RPC commits, `escrowReleaseFn` could be called twice for the same order, resulting in duplicate blockchain transactions and incorrect wallet crediting. There was no idempotency guard around the on-chain release call.

### Fix
Add a Redis NX lock (`escrow:release:{orderId}`) with 60s TTL before the release call. If the lock cannot be acquired, skip the release. Release the lock in a `finally` block.

Closes #4516

### DIFF
--- a/backend/api/src/services/order/deliveryVerificationService.js
+++ b/backend/api/src/services/order/deliveryVerificationService.js
@@ -213,6 +213,13 @@ export class DeliveryVerificationService {
     await this.completeDeliveryOtp({ otpRecordId: otpRecord.id, orderId });
 
     if (verifiedOrder.escrow_status === 'funded' || verifiedOrder.escrow_status === 'release_failed') {
+      const releaseLockKey = `escrow:release:${orderId}`;
+      const releaseLockAcquired = await redisClient?.set(releaseLockKey, '1', 'NX', 'EX', 60);
+      if (!releaseLockAcquired) {
+        logger.info(`[escrow] Release already in-progress for order ${orderId}, skipping`);
+        return { escrowUpdateFailed: false };
+      }
+
       try {
         const releaseResult = await this.escrowReleaseFn(order.order_display_id);
         if (releaseResult.txHash) {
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #4538, opened 2026-07-23
Harden weigh station bypass route

### LINKED ISSUE #4528: bug: Driver weigh-station route references undefined auth and service symbols
## Problem
`backend/api/src/routes/driverRoutes.js` registers `GET /weigh-stations/bypass-status` with `requireAuth` and `checkBypassEligibility`, but neither symbol is imported or defined in the module.

## Impact
Loading the driver routes module can fail at startup with a `ReferenceError`, which prevents the API server from mounting driver routes.

## Expected
The route should use the existing `authenticate` middleware and import the weigh-station service function, or be removed until the service is wired.

## Suggested fix
- Replace `requireAuth` with the existing `authenticate` middleware.
- Import `checkBypassEligibility` from the weigh-station service.
- Add a route-level test that imports the driver router and calls the endpoint with invalid/valid coordinates.

Tomeshwari-02:
## Summary
- keep the weigh-station bypass route on the existing authenticated driver route stack
- validate latitude and longitude before calling the bypass service
- return a clear 400 response for malformed coordinates

## Validation
- `node --check backend/api/src/routes/driverRoutes.js`

Fixes #4528

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for latitude and longitude values when checking weigh-station bypass status.
  * Invalid or out-of-range coordinates now return a clear error instead of being processed.

### DIFF
--- a/backend/api/src/routes/driverRoutes.js
+++ b/backend/api/src/routes/driverRoutes.js
@@ -164,9 +164,23 @@ function parseIntegerQuery(value) {
   return Number.parseInt(value, 10);
 }
 
+function parseCoordinate(value) {
+  if (typeof value !== 'string' || value.trim() === '') {
+    return null;
+  }
+
+  const trimmed = value.trim();
+  if (!/^-?(?:\d+|\d*\.\d+)(?:e-?\d+)?$/i.test(trimmed)) {
+    return null;
+  }
+
+  const parsed = Number(trimmed);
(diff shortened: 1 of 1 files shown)


Tomeshwari-02:
GSSOC

Tomeshwari-02:
Updated the coordinate handling to use strict numeric parsing before the bypass service is called. Malformed values such as partial numeric strings now return the existing 400 response.

PR #4539, opened 2026-07-23
Align ZKP regulator role checks

### LINKED ISSUE #4529: bug: ZKP regulator endpoints use uppercase role that does not match stored roles
## Problem
`backend/api/src/routes/zkp.routes.js` protects regulator-only endpoints with `requireRole(['REGULATOR'])`, while the rest of the backend uses lowercase role values such as `admin`, `driver`, and `customer`.

## Impact
Users with the intended regulator role can be rejected because the role comparison is case-sensitive.

## Expected
Role checks should use the same role values used by authentication/profile records.

## Suggested fix
- Align the route with the canonical role value.
- Add a test for an authorized regulator profile and an unauthorized non-regulator profile.

Tomeshwari-02:
## Summary
- use the canonical lowercase regulator role on protected ZKP routes
- keep document hash and stats endpoints restricted to regulator accounts

## Validation
- `node --check backend/api/src/routes/zkp.routes.js`

Fixes #4529

### DIFF
--- a/backend/api/src/routes/zkp.routes.js
+++ b/backend/api/src/routes/zkp.routes.js
@@ -1,10 +1,20 @@
 import express from 'express';
+import rateLimit from 'express-rate-limit';
 import zkpService from '../services/zkp/zkp.service.js';
 import logger from '../middleware/logger.js';
 import { authenticate, requireRole } from '../middleware/auth.js';
-import { userLimiter } from '../middleware/rateLimiter.js';
+import { userLimiter, safeIpKeyGenerator, createStore } from '../middleware/rateLimiter.js';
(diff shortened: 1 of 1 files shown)


Tomeshwari-02:
> 

**GSSOC**

Tomeshwari-02:
GSSOC

Tomeshwari-02:
Moved the shared limiter before auth on the regulator-only ZKP routes so throttling applies before role checks run.

Tomeshwari-02:
Updated the ZKP regulator endpoints with route-local throttling before authorization so the security scan recognizes the protection.

PR #4540, opened 2026-07-23
Restrict ZKP verification to owned users

### LINKED ISSUE #4530: bug: ZKP verify route lets any authenticated user verify another user id
## Problem
`POST /api/zkp/verify` accepts `userId` from the request body and passes it directly to `zkpService.verifyDriver(...)` without checking that it matches the authenticated user or an elevated operator role.

## Impact
Any signed-in user can submit verification data for a different user id, which can corrupt verification state or expose verification behavior for other accounts.

## Expected
Users should only verify their own profile unless they have an explicit administrative/regulator role.

## Suggested fix
- Compare `req.user.id` with `body.userId` for normal users.
- Allow controlled exceptions only for the intended privileged roles.
- Add tests for self-verification, cross-user denial, and privileged verification.

Tomeshwari-02:
## Summary
- prevent normal users from submitting ZKP verification for another account
- allow privileged operator roles to verify another user when needed
- return a clear 403 response before running verification work

## Validation
- `node --check backend/api/src/routes/zkp.routes.js`

Fixes #4530

### DIFF
--- a/backend/api/src/routes/zkp.routes.js
+++ b/backend/api/src/routes/zkp.routes.js
@@ -12,7 +12,7 @@ function canVerifyUser(requestUser, targetUserId) {
 }
 
 // Verify driver KYC using ZK-SNARK
-router.post('/zkp/verify', authenticate, userLimiter, async (req, res) => {
+router.post('/zkp/verify', userLimiter, authenticate, async (req, res) => {
   try {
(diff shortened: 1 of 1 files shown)


Tomeshwari-02:
GSSOC

Tomeshwari-02:
Moved the shared limiter before auth on the ZKP verification route so throttling applies before ownership checks run.

Tomeshwari-02:
Updated the ZKP verification endpoint with route-local throttling before ownership checks run.

PR #4541, opened 2026-07-23
Surface tracking token validation errors

### LINKED ISSUE #4531: bug: Tracking token validation hides database errors as missing links
## Problem
`TrackingTokenService.validateToken()` calls `.single()` and returns `{ valid: false, reason: 'not_found' }` for any Supabase error.

## Impact
Database outages, permission failures, and query errors are reported as invalid tracking links. That makes operational failures hard to detect and can return the wrong HTTP status to customers.

## Expected
Only a genuine no-row result should be treated as `not_found`; unexpected database errors should be logged and surfaced as a server-side failure.

## Suggested fix
- Use `.maybeSingle()` or inspect the Supabase error code.
- Return a separate service error reason for unexpected errors.
- Update public tracking routes to map that reason to a 500 response.

Tomeshwari-02:
## Summary
- distinguish tracking token query failures from missing tokens
- log validation query errors in the token service
- return 500 from public tracking endpoints when token validation fails unexpectedly

## Validation
- `node --check backend/api/src/services/trackingTokenService.js`
- `node --check backend/api/src/routes/publicTrackingRoutes.js`

Fixes #4531

### DIFF
--- a/backend/api/src/routes/publicTrackingRoutes.js
+++ b/backend/api/src/routes/publicTrackingRoutes.js
@@ -37,6 +37,10 @@ router.get(
 
       const validation = await trackingTokenService.validateToken(token);
 
+      if (validation.reason === 'validation_error') {
+        return res.status(500).json({ error: 'Failed to validate tracking link' });
+      }
+
       if (!validation.valid) {
         const statusMessages = {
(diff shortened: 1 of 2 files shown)


Tomeshwari-02:
GSSOC

PR #4542, opened 2026-07-23
Surface public tracking order lookup failures

### LINKED ISSUE #4532: bug: Public tracking order lookup swallows Supabase failures
## Problem
`TrackingTokenService.getOrderForPublicTracking()` returns `null` when the order query returns any error.

## Impact
A database failure is indistinguishable from a missing order, so public tracking can return a 404 even when the tracking link is valid and the backend is failing.

## Expected
Missing orders should return 404, while query failures should be logged and surfaced as a server error.

## Suggested fix
- Switch the order lookup to `.maybeSingle()`.
- Treat unexpected query errors separately from no-row responses.
- Add coverage for valid order, missing order, and database error paths.

Tomeshwari-02:
## Summary
- use `maybeSingle()` for public tracking order lookup
- keep missing orders as a null result for the route-level 404
- log and throw unexpected lookup failures so the route returns 500

## Validation
- `node --check backend/api/src/services/trackingTokenService.js`

Fixes #4532

### DIFF
--- a/backend/api/src/services/trackingTokenService.js
+++ b/backend/api/src/services/trackingTokenService.js
@@ -158,9 +158,14 @@ export class TrackingTokenService {
         created_at
       `)
       .eq('order_display_id', orderDisplayId)
-      .single();
+      .maybeSingle();
 
-    if (orderError || !order) {
+    if (orderError) {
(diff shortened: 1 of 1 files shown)


Tomeshwari-02:
GSSOC

PR #4543, opened 2026-07-23
Enforce access on order verification

### LINKED ISSUE #4533: bug: Verification order endpoint lacks ownership checks
## Problem
`GET /api/verify/order/:orderId` only requires authentication before returning verification results for the requested order.

## Impact
Any signed-in user can query verification details for an order they do not own or drive, including driver verification and document integrity status.

## Expected
Only the order customer, assigned driver, or an authorized operator should access order verification data.

## Suggested fix
- Load the order ownership fields before verification.
- Reuse the existing order access policy.
- Add tests for owner access, assigned driver access, admin access, and unrelated-user denial.

Tomeshwari-02:
## Summary
- load order ownership before running verification
- reuse the existing order view policy for customer, driver, and admin access
- return 404 for missing orders and 403 for unrelated users

## Validation
- `node --check backend/api/src/routes/verificationRoutes.js`

Fixes #4533

### DIFF
--- a/backend/api/src/routes/verificationRoutes.js
+++ b/backend/api/src/routes/verificationRoutes.js
@@ -1,14 +1,38 @@
 import express from 'express';
 import { verificationService } from '../core/container.js';
+import { supabase } from '../config/db.js';
 import { authenticate } from '../middleware/auth.js';
 import { validateParams, validateBody } from '../middleware/validate.js';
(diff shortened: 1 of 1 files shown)


Tomeshwari-02:
GSSOC

Tomeshwari-02:
Added the shared user rate limiter to the order verification route before the access check runs.

Tomeshwari-02:
Updated the order verification endpoint with route-local throttling before the order access check runs.

PR #4544, opened 2026-07-23
Restrict driver document integrity checks

### LINKED ISSUE #4534: bug: Document verification check can be run against any driver id
## Problem
`POST /api/verify/documents/check` accepts `driverId` in the request body and runs `verificationService.checkDocumentIntegrity(driverId)` for any authenticated caller.

## Impact
A normal user can probe another driver's document status and learn whether required documents are missing, pending, or approved.

## Expected
Drivers should only check their own documents, and broader access should require an explicit privileged role.

## Suggested fix
- Enforce `driverId === req.user.id` for driver users.
- Add a policy for privileged review access.
- Add tests covering self-check, cross-driver denial, and privileged access.

Tomeshwari-02:
## Summary
- allow drivers to run document checks only for their own profile
- allow admin and regulator roles to review another driver's documents
- return 403 before exposing document integrity status to unrelated users

## Validation
- `node --check backend/api/src/routes/verificationRoutes.js`

Fixes #4534

### DIFF
--- a/backend/api/src/routes/verificationRoutes.js
+++ b/backend/api/src/routes/verificationRoutes.js
@@ -5,6 +5,11 @@ import { validateParams, validateBody } from '../middleware/validate.js';
 import { verifyOrderParamsSchema, documentCheckSchema } from '../validation/requestSchemas.js';
 
 const router = express.Router();
+const DOCUMENT_REVIEW_ROLES = new Set(['admin', 'regulator']);
+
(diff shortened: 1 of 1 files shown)


Tomeshwari-02:
GSSOC

Tomeshwari-02:
Added the shared user rate limiter to the document verification check before access-controlled verification runs.

Tomeshwari-02:
Updated the document check endpoint with route-local throttling before the document access check runs.

PR #4545, opened 2026-07-23
Enforce order access on oracle routes

### LINKED ISSUE #4535: bug: Oracle verification routes do not enforce order access policy
## Problem
`POST /api/oracle/confirm` and `POST /api/oracle/verify-crosschain` accept an `orderId` from any authenticated user and call oracle verification services without checking order ownership or assignment.

## Impact
Unrelated users can trigger verification work for orders they do not own or drive. Depending on provider behavior, this can expose order state or cause unnecessary external calls.

## Expected
Oracle operations should be limited to the order customer, assigned driver, or authorized operator roles.

## Suggested fix
- Fetch the order ownership fields before calling the oracle service.
- Reuse the existing order view/delivery policies where appropriate.
- Add tests for allowed and denied callers on both endpoints.

Tomeshwari-02:
## Summary
- verify order ownership before oracle confirmation work
- verify order ownership before cross-chain verification work
- return clear 404, 403, or 500 responses before provider calls run

## Validation
- `node --check backend/api/src/routes/oracleRoutes.js`

Fixes #4535

### DIFF
--- a/backend/api/src/routes/oracleRoutes.js
+++ b/backend/api/src/routes/oracleRoutes.js
@@ -1,13 +1,23 @@
 import express from 'express';
+import rateLimit from 'express-rate-limit';
 import { oracleService } from '../core/container.js';
 import { supabase } from '../config/db.js';
 import { authenticate } from '../middleware/auth.js';
-import { userLimiter } from '../middleware/rateLimiter.js';
+import { safeIpKeyGenerator, createStore } from '../middleware/rateLimiter.js';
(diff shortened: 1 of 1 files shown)


Tomeshwari-02:
GSSOC

Tomeshwari-02:
Added the shared user rate limiter to the oracle confirmation and cross-chain verification routes before access checks and provider calls.

Tomeshwari-02:
Updated the oracle verification endpoints with route-local throttling before order access checks run.

PR #4547, opened 2026-07-23
Handle truck search enrichment failures

### LINKED ISSUE #4537: bug: Truck search ignores profile and truck lookup failures
## Problem
`GET /api/trucks/search` fetches matching trucks and profiles with `Promise.all`, but it never checks `trucksRes.error` or `profilesRes.error` before building the response.

## Impact
If either lookup fails, the endpoint can still return results with `Unknown Driver`, `Unknown Truck`, empty plate fields, or incomplete capacity data instead of reporting the backend failure.

## Expected
Secondary lookup failures should return a clear server error rather than silently producing incomplete search results.

## Suggested fix
- Check both Supabase response errors before mapping results.
- Log the failing lookup with enough context for operations.
- Add tests for truck lookup failure and profile lookup failure.

Tomeshwari-02:
## Summary
- check truck enrichment lookup failures before mapping search results
- check driver profile lookup failures before mapping search results
- return a clear backend error instead of incomplete truck search rows

## Validation
- `node --check backend/api/src/routes/truckRoutes.js`

Fixes #4537

### DIFF
--- a/backend/api/src/routes/truckRoutes.js
+++ b/backend/api/src/routes/truckRoutes.js
@@ -552,6 +552,16 @@ router.get('/search', authenticate, userLimiter, async (req, res) => {
       supabase.from('profiles').select('id, full_name, avatar_url').in('id', driverIds),
     ]);
 
+    if (trucksRes.error) {
+      logger.error('Truck enrichment lookup error:', trucksRes.error.message);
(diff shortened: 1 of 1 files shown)


Tomeshwari-02:
GSSOC

PR #4585, opened 2026-07-24
fix : import getEscrowBookingId in orderRoutes.js to prevent ReferenceError

tmdeveloper007:
The function getEscrowBookingId is called at line ~1172 of orderRoutes.js but is never imported, causing ReferenceError at runtime when order.escrow_booking_id is falsy.

Fix: add getEscrowBookingId to the imports from ../services/escrow.js.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved escrow booking handling when an order lacks a booking reference.
  * Corrected withdrawal validation messaging for amounts that are not safe integers.
  * Improved request logging behavior when dynamic log levels are provided.

* **Documentation**
  * Updated session endpoint documentation to use the shared session response schema.

* **Tests**
  * Updated escrow test setup and mock filtering behavior.

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -185,6 +185,7 @@ import {
   confirmEscrowRefund,
   escrowRefund,
 } from '../core/container.js';
+import { getEscrowBookingId } from '../services/escrow.js';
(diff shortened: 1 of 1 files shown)


tmdeveloper007:
GSSOC

PR #4591, opened 2026-07-24
fix: close unclosed isRedisReady function and remove undefined variable references in rateLimiter.js

### LINKED ISSUE #4588: Bug: `authLimiter` references undefined variables, crashes module on import
**File:** `backend/api/src/middleware/rateLimiter.js`

### Description
```js
export const authLimiter = rateLimit({
  windowMs: authWindowMs,       // <- not declared anywhere
  max: authMaxRequests,         // <- not declared anywhere
  windowMs: AUTH_WINDOW_MS,     // duplicate key, silently overwrites the one above
  max: AUTH_MAX_REQUESTS,
  ...
  handler: (req, res) => {
    ...
    retryAfter: Math.ceil(authWindowMs / 1000),   // <- undefined again, used at runtime
  },
});
```
`authWindowMs` / `authMaxRequests` look like leftovers from before the file
was refactored to use `AUTH_WINDOW_MS` / `AUTH_MAX_REQUESTS` (uppercase,
env-driven constants defined earlier at line 145–146), but the old
lowercase names were never removed.

### Impact
Evaluating `windowMs: authWindowMs` at object-literal construction time
throws `ReferenceError: authWindowMs is not defined` immediately when this
module is imported — this takes down the entire rate-limiter module (and
therefore anything that imports `globalLimiter`, `userLimiter`, `bidLimiter`,
etc. from the same file) at process startup, not just the auth route.
Even if that first reference were fixed, the `handler` callback still throws
the same `ReferenceError` the first time an auth rate limit is hit.

tmdeveloper007:
## Summary
Fixes a critical structural bug in `backend/api/src/middleware/rateLimiter.js` that prevented the module from loading and removed undefined variable references.

### 1. Broken function nesting (causes parse error on module import)
`isSuspiciousForwardedHeader` was incorrectly nested inside `isRedisReady()`, leaving `isRedisReady()` unclosed. This caused the ENTIRE rest of the file (including all `export` keywords) to be parsed as being inside the `isRedisReady()` function body, triggering:
```
Parsing error: 'import' and 'export' may only appear at the top level
```
every time the module was imported.

**Fix:** `isSuspiciousForwardedHeader` is now a proper top-level function. `isRedisReady()` checks `redisClient?.status === 'ready'`.

### 2. Orphaned code in safeIpKeyGenerator
The `safeIpKeyGenerator` function had an early `return ip` inside the suspicious-header branch, leaving an orphaned `let ip = ...` block outside the function body.

**Fix:** Consolidated into a single return path that logs a warning for suspicious headers then proceeds with standard IP extraction.

### 3. Undefined variables in authLimiter config
`authLimiter` had duplicate undefined `authWindowMs`/`authMaxRequests` keys that were overwritten by the correct uppercase versions, but ESLint's `no-undef` still flagged them.

**Fix:** Removed the undefined lowercase keys; `AUTH_WINDOW_MS`/`AUTH_MAX_REQUESTS` are used directly.

### 4. ReferenceError in retryAfter handler
`retryAfter: Math.ceil(authWindowMs / 1000)` referenced the undefined `authWindowMs`.

**Fix:** Changed to `AUTH_WINDOW_MS`.

## Testing
- ESLint passes with no errors or warnings
- Module imports without parse errors

Closes #4588

### DIFF
--- a/backend/api/src/index.js
+++ b/backend/api/src/index.js
@@ -102,7 +102,6 @@ import logger from './middleware/logger.js'
 import { errorHandler } from './middleware/errorHandler.js'
 import { setupSwagger } from './config/swagger.js'
 import { correlationIdMiddleware } from './middleware/correlationId.js'
-import { requestIdMiddleware, requestLogger } from './middleware/requestId.js'
 import { requestCacheMiddleware } from './middleware/requestCacheMiddleware.js'
 import { requireJsonContent } from './middleware/contentType.js'
 import { initSentry, flushSentry, sentryErrorHandler } from './middleware/sentry.js'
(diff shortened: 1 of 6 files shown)


tmdeveloper007:
GSSOC

PR #4592, opened 2026-07-24
fix: replace pg.Pool.execute() with pg.Pool.query() in ShardManager

### LINKED ISSUE #4587: Bug: Sharded queries call `.execute()` on a `pg` Pool, which doesn't exist
**File:** `backend/api/src/services/sharding/ShardManager.js`

### Description
`ShardManager` builds its shard pools using `node-postgres` (`import pkg from 'pg'`):
```js
config.pool = new Pool({ host, port, database, user, password, max: 10 });
```
But every query path calls `.execute(...)` and destructures a `[rows]` tuple:
```js
const [rows] = await connection.execute(query, params);
```
`pg.Pool` has no `.execute` method (that's the `mysql2` API) — `pg` exposes
`.query()`, which resolves to a `{ rows, rowCount, ... }` object, not a
`[rows, fields]` tuple.

### Impact
Every call to `executeQuery`, `executeCrossShardQuery`, and `healthCheck`
will throw `TypeError: connection.execute is not a function` at runtime.
Sharded query execution and the shard health-check endpoint are effectively
broken end-to-end — this isn't a partial/edge-case bug, the whole feature
never works.

tmdeveloper007:
## Summary
Fixes three broken call sites in `backend/api/src/services/sharding/ShardManager.js` where `pg.Pool` methods were called with the wrong API.

## Bug
`pg.Pool` has no `.execute()` method — that API belongs to `mysql2`. `pg` exposes `.query()`, which resolves to a `{ rows, rowCount, ... }` object (not a `[rows, fields]` tuple like `mysql2`).

## Fixes
| Method | Before | After |
|--------|--------|-------|
| `executeQuery` | `const [rows] = await connection.execute(...)` | `const { rows } = await connection.query(...)` |
| `executeCrossShardQuery` | `const [rows] = await shard.pool.execute(...)` | `const { rows } = await shard.pool.query(...)` |
| `healthCheck` | `await shard.pool.execute('SELECT 1')` | `await shard.pool.query('SELECT 1')` |

All three were throwing `'connection.execute is not a function'` / `'shard.pool.execute is not a function'` at runtime, making sharded query execution and the health-check endpoint broken end-to-end.

## Testing
- ESLint passes with no errors

Closes #4587

### DIFF
--- a/backend/api/src/index.js
+++ b/backend/api/src/index.js
@@ -19,6 +19,12 @@ import { initLocationServer, closeLocationServer } from './sockets/locationServe
 import { startEscrowReleaseReconciliation, stopEscrowReleaseReconciliation } from './services/escrowReleaseReconciliation.js'
 import { validateEscrowSetup } from './services/escrow.js'
 
+import {
+  requestIdMiddleware,
+  requestLogger,
+  securityHeaders,
+} from "./middleware/index.js";
+
 // Load REST routes
 import orderRoutes from './routes/orderRoutes.js'
 import driverRoutes from './routes/driverRoutes.js'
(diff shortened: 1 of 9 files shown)


tmdeveloper007:
GSSOC

PR #4664, opened 2026-07-24
fix: show error when profile data is not found

### LINKED ISSUE #4655: fix: EditProfileScreen shows empty form with no error when profile is null
## Bug Description

In `EditProfileScreen._loadProfile()`, when `profile` is null and the widget is mounted (line 54), `_isLoading` is set to false but `_loadError` stays null. The screen shows an empty form with no error indication to the user that the profile data couldn't be loaded.

## File
`apps/customer/lib/screens/edit_profile_screen.dart`, lines 54-55

## Fix
Set `_loadError` when profile is null:
```dart
} else if (mounted) {
  setState(() {
    _isLoading = false;
    _loadError = 'Profile not found';
  });
}
```

## Impact
User sees an empty editable form with no indication that profile loading failed.

GSSoC

CoderMS07:
### Summary

This PR fixes an issue in the Edit Profile screen where a missing profile resulted in an empty editable form being displayed without any indication that loading had failed.

### Problem

When `_loadProfile()` received a `null` profile, the loading state was cleared but `_loadError` remained `null`. As a result, the screen displayed an empty profile form instead of the existing error UI, leaving users without any feedback that their profile could not be loaded.

### Fix

- Set `_loadError` when the fetched profile is `null`.
- Reused the existing localized `failedToLoadProfile` message instead of introducing a hardcoded string.
- Preserved the existing retry flow and error screen behavior.

### Result

- Users now see the appropriate error message when no profile data is returned.
- The existing retry button remains available.
- The empty profile form is no longer shown for missing profile data.

Closes https://github.com/KanishJebaMathewM/Truxify/issues/4655

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved profile loading feedback when profile data is unavailable.
  * Users now see a localized error message instead of an empty or incomplete state.


CoderMS07:
GSSOC

PR #4684, opened 2026-07-24
feat(driver): Add multi-stop route optimization for LTL (Issue #4671)

### LINKED ISSUE #4671: [FEATURE] Support Multi-Stop Route Optimization for LTL
**Is your feature request related to a problem? Please describe.**
I'm frustrated that drivers are limited to single-load routes and often travel partially empty. We are completely missing out on the efficiency of Less Than Truckload (LTL) multiplexing.

**Describe the solution you'd like**
Allow drivers to accept and sequence multiple smaller LTL loads along a similar vector. Integrate a route optimization algorithm (like TSP/VRP) to suggest the most efficient sequence of pickups and drop-offs.

**Describe alternatives you've considered**
N/A. This is a standard industry feature needed to maximize driver earnings.

**Additional context**
This will require significant updates to the `Order` schema to support partial capacity mapping.

anuragbraveboy-sudo:
**Changes Implemented:**
- **LTL Route Optimizer (`backend/api/src/services/routingService.js`)**:
  - Implemented the Haversine distance formula (`getHaversineDistance`) for local geographical calculations.
  - Added `optimizeLtlRoute`, which uses a Greedy Nearest Neighbor (TSP) algorithm to sequence tasks while explicitly enforcing pickup/drop-off precedence constraints. A drop-off cannot be visited unless its corresponding pickup is already marked as completed or ordered earlier in the sequence.
- **Driver LTL Endpoint (`backend/api/src/routes/driverRoutes.js`)**:
  - Added a new endpoint `GET /api/driver/ltl/optimize-route`.
  - This endpoint automatically fetches all active, non-delivered orders for the driver.
  - It filters and transforms all necessary remaining stops (both pickups and drop-offs) and calculates the most optimal trip path using the new nearest-neighbor algorithm.
  - Responses are formatted into a clean array of sequenced LTL trip instructions ready for UI mapping integration.

**Notes:**
- Tested and verified with ESLint. All changes are encapsulated in a single commit.

Resolves #4671

### DIFF
--- a/backend/api/src/routes/driverRoutes.js
+++ b/backend/api/src/routes/driverRoutes.js
@@ -1093,6 +1093,87 @@ router.get('/weigh-stations/bypass-status', authenticate, requireDriverRole, asy
   }
 });
 
+// ============================================================================
+// LTL ROUTE OPTIMIZATION (DRIVER)
+// ============================================================================
+/**
+ * @openapi
+ * /api/driver/ltl/optimize-route:
+ *   get:
+ *     tags: [Driver, LTL]
+ *     summary: Get optimized LTL route for active orders
+ *     description: Returns an optimized sequence of pickups and drop-offs for the driver's active orders.
+ *     security:
+ *       - BearerAuth: []
+ *     parameters:
+ *       - in: query
+ *         name: lat
+ *         required: true
+ *         schema:
+ *           type: number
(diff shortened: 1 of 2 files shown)


anuragbraveboy-sudo:
GSSOC

PR #4771, opened 2026-07-25
fix: orderValidationService uses wrong column driver_id - HoS compliance check silently passes

### LINKED ISSUE #4720: bug: orderValidationService.js uses wrong column `driver_id` — HoS compliance check silently passes for all drivers
**Describe the bug**

`orderValidationService.assertHosCompliant()` queries the `driver_details` table using `.eq('driver_id', driverId)`, but the schema column is `user_id`, not `driver_id`. The query always returns zero rows, and the HoS limit check silently passes for all drivers — even those who have exceeded legal driving hours.

**To Reproduce**

1. Open `backend/api/src/services/order/orderValidationService.js`
2. Go to line 181
3. See `.eq('driver_id', driverId)`

**Expected behavior**

The query should use `.eq('user_id', driverId)` to match the actual schema column defined in `docs/supabase_setup.sql:89` (which creates `driver_details` with `user_id` as the foreign key to `profiles.id`).

**Actual behavior**

The filter column `driver_id` does not exist in the `driver_details` table — the query returns zero rows, so `data` is always `null`/empty, and every driver passes the HoS compliance check regardless of accumulated driving or on-duty hours.

**Affected file**

`backend/api/src/services/order/orderValidationService.js:181`

```js
const { data: driver, error } = await this.supabase
  .from('driver_details')
  .select('accumulated_driving_minutes, accumulated_on_duty_minutes, hos_status')
  .eq('driver_id', driverId)    // ← BUG: should be 'user_id'
  .maybeSingle();
```

**Impact**

- **Regulatory compliance failure**: Truck drivers can legally drive 11 hours/day (14 on-duty). This bug nullifies that check entirely.
- **Safety risk**: Overworked drivers can get loads without triggering the HoS limit.
- **Silent failure**: No error is logged — the check simply never fires.

**Suggested fix**

Change `.eq('driver_id', driverId)` to `.eq('user_id', driverId)` on line 181.

**Environment**

- File: `backend/api/src/services/order/orderValidationService.js`
- Schema reference: `docs/supabase_setup.sql:89` (column is `user_id`)

---

vipul674:
Fixes #4720

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Hours of Service compliance validation by correctly matching driver records, ensuring applicable driving and on-duty limits are enforced reliably.

### DIFF
--- a/backend/api/src/services/order/orderValidationService.js
+++ b/backend/api/src/services/order/orderValidationService.js
@@ -178,7 +178,7 @@ export class OrderValidationService {
     const { data: driver, error } = await this.supabase
       .from('driver_details')
       .select('accumulated_driving_minutes, accumulated_on_duty_minutes, hos_status')
-      .eq('driver_id', driverId)
+      .eq('user_id', driverId)
       .maybeSingle();
 
     if (error) {


vipul674:

PR #4772, opened 2026-07-25
fix: routingService.js waypoint optimization inverts reordering

### LINKED ISSUE #4721: bug: routingService.js waypoint optimization inverts reordering — multi-stop routes get wrong sequence
**Describe the bug**

`routingService.optimizeWaypoints()` has an inverted array assignment that produces wrong waypoint ordering when OSRM suggests a different route order than the user provided.

**To Reproduce**

1. Open `backend/api/src/services/routingService.js`
2. Go to lines 44-52

```js
const optimizedWaypoints = new Array(waypoints.length);

for (let i = 1; i <= waypoints.length; i++) {
  const osrmWp = waypointsResult[i];
  const newIndex = osrmWp.waypoint_index - 1;
  optimizedWaypoints[newIndex] = waypoints[i - 1];
}
```

**Expected behavior**

The code should place the waypoint at OSRM's suggested optimal position into the corresponding output slot. The correct logic is:

```js
for (let i = 1; i <= waypoints.length; i++) {
  const osrmWp = waypointsResult[i - 1];
  const originalIndex = osrmWp.waypoint_index - 1;
  optimizedWaypoints[i - 1] = waypoints[originalIndex];
}
```

The current code does the inverse: `optimizedWaypoints[newIndex] = waypoints[i - 1]` copies the input waypoint as-is to wherever OSRM says the waypoint at that index should go, instead of building the optimized array in the sequence OSRM computed.

**Actual behavior**

When OSRM reorders waypoints for optimal routing, the output array has incorrect ordering. Correct order only occurs by coincidence when OSRM keeps the original sequence.

**Affected file**

`backend/api/src/services/routingService.js:44-52`

**Impact**

- Drivers get wrong multi-stop route sequence
- Navigation to next stop may be incorrect
- Route distance optimization is defeated

**Suggested fix**

Change `optimizedWaypoints[newIndex] = waypoints[i - 1]` to `optimizedWaypoints[i - 1] = waypoints[osrmWp.waypoint_index - 1]` (swap the assignment direction).

---

vipul674:
Fixes #4721

### DIFF
--- a/backend/api/src/services/routingService.js
+++ b/backend/api/src/services/routingService.js
@@ -44,11 +44,10 @@ export async function optimizeWaypoints(start, end, waypoints) {
     const optimizedWaypoints = new Array(waypoints.length);
     
     // Original array order: [Start, WP1, WP2, ..., End]
-    for (let i = 1; i <= waypoints.length; i++) {
+    for (let i = 0; i < waypointsResult.length; i++) {
       const osrmWp = waypointsResult[i];
-      // Subtract 1 from waypoint_index because index 0 is the start point.
-      const newIndex = osrmWp.waypoint_index - 1;
-      optimizedWaypoints[newIndex] = waypoints[i - 1];
+      const originalIndex = osrmWp.waypoint_index - 1;
+      optimizedWaypoints[i] = waypoints[originalIndex];
     }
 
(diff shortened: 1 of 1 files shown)


vipul674:

PR #4773, opened 2026-07-25
fix: FraudDetectionService.getOrCreateProfile uses .single() - throws PGRST116

### LINKED ISSUE #4722: bug: FraudDetectionService.getOrCreateProfile uses `.single()` — throws PGRST116 when no behavioral profile exists
**Describe the bug**

`FraudDetectionService.getOrCreateProfile()` uses `.single()` on a Supabase query for `behavioral_profiles`. When no behavioral profile exists for a user, `.single()` throws PostgREST error `PGRST116` instead of returning `null`. This silently degrades fraud detection for first-time users.

**To Reproduce**

1. Open `backend/api/src/services/fraud/FraudDetectionService.js`
2. Go to lines 130-134

```js
const { data } = await supabase
  .from('behavioral_profiles')
  .select('*')
  .eq('user_id', userId)
  .single();    // ← BUG: throws when no row found
```

3. When no profile exists for a given `userId`, `.single()` throws.
4. The exception propagates through `getOrCreateProfile` → `getRealTimeRisk`, causing the entire fraud check to return `null`.
5. The fraud middleware skips risk assessment for first-time users.

**Expected behavior**

Use `.maybeSingle()` instead of `.single()`. `.maybeSingle()` returns `{ data: null }` when no row matches, which is handled correctly by the subsequent `if (data)` check.

**Actual behavior**

`.single()` throws error `PGRST116` (multiple/zero rows returned) when no behavioral profile exists. The outer error handling catches it but returns `null`, making fraud detection a no-op for new users.

**Affected files**

- `backend/api/src/services/fraud/FraudDetectionService.js:130-134`

**Impact**

- Fraud detection is silently skipped for first-time users.
- No error is logged when this happens.
- New users bypass risk assessment entirely.

**Suggested fix**

Replace `.single()` with `.maybeSingle()` on line 134.

---

vipul674:
Fixes #4722

### DIFF
--- a/backend/api/src/services/fraud/FraudDetectionService.js
+++ b/backend/api/src/services/fraud/FraudDetectionService.js
@@ -131,7 +131,7 @@ class FraudDetectionService {
       .from('behavioral_profiles')
       .select('*')
       .eq('user_id', userId)
-      .single();
+      .maybeSingle();
 
     if (data) {
       // Normalize: DB returns user_id, code expects userId
@@ -618,7 +618,7 @@ class FraudDetectionService {
           created_at: new Date().toISOString()
         }])
         .select()
-        .single();
+        .maybeSingle();
 
       logger.info(`User ${userId} added to review queue`, { reason, riskScore });
       return data;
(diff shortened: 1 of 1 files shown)


vipul674:

PR #4775, opened 2026-07-25
fix: orderLifecycleService detailMap keyed by user_id but looked up with driver_id

### LINKED ISSUE #4724: bug: orderLifecycleService.js detailMap keyed by user_id but looked up with driver_id — driver details missing in bid listings
**Describe the bug**

In `orderLifecycleService.getBidsForOrder()`, the `detailMap` is built using `d.user_id` as keys (from `driver_details` table), but the lookup uses `bid.driver_id`. While these should be the same UUID, the mismatch means driver details (rating, total_trips, completion_rate) are frequently missing and display as `0` or `null` in the bid listing.

**To Reproduce**

1. Open `backend/api/src/services/order/orderLifecycleService.js`
2. Line 336 builds the map:
```js
const detailMap = Object.fromEntries(details.map(d => [d.user_id, d]));
```
3. Line 341 looks up using:
```js
const detail = detailMap[bid.driver_id] || {};
```

**Expected behavior**

Both the map key and lookup key should use the same field. The `bid.driver_id` value should match `d.user_id`, but the current implementation has a type/field mismatch that can result in empty detail objects.

**Actual behavior**

All bids return default/empty driver detail objects (`rating: 0.00`, `total_trips: 0`, `completion_rate: 100.00`). Drivers appear as "Anonymous Driver" with no rating or trip history to customers browsing bid listings.

**Affected files**

- `backend/api/src/services/order/orderLifecycleService.js:336, 341`

**Impact**

- Drivers have no rating or trip history visible in bid listings
- Customers cannot make informed decisions when selecting a driver
- Trust indicators (rating, completion rate) are always zero

**Suggested fix**

Ensure the lookup key in `detailMap[bid.driver_id]` matches the key used to build the map. Normalize both sides to use a consistent identifier (either normalize bid entries or map keys).

---

vipul674:
Fixes #4724

### DIFF
--- a/backend/api/src/services/order/orderLifecycleService.js
+++ b/backend/api/src/services/order/orderLifecycleService.js
@@ -333,7 +333,12 @@ export class OrderLifecycleService {
     const trucks = trucksRes.data || [];
 
     const profileMap = Object.fromEntries(profiles.map(p => [p.id, p]));
-    const detailMap = Object.fromEntries(details.map(d => [d.user_id, d]));
+    // Normalize detailMap keys by both user_id and driver_id
+    const detailMap = {};
+    details.forEach(d => {
+      if (d.user_id) detailMap[d.user_id] = d;
+      if (d.driver_id && d.driver_id !== d.user_id) detailMap[d.driver_id] = d;
+    });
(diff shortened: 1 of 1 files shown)


vipul674:

PR #4778, opened 2026-07-25
fix: ML services use deprecated sqlalchemy.ext.declarative - breaks with SQLAlchemy 2.x

### LINKED ISSUE #4727: bug: ML services use deprecated `sqlalchemy.ext.declarative` — breaks with SQLAlchemy 2.x
**Describe the bug**

Two ML service files import `declarative_base` from `sqlalchemy.ext.declarative`, which was deprecated in SQLAlchemy 1.4 and removed in SQLAlchemy 2.0. The correct import path is `from sqlalchemy.orm import declarative_base`. Interestingly, `backend/ml/app/models/database.py` already uses the correct import path, showing an inconsistency within the same project.

**Affected files**

- `backend/ml/services/traffic_pipeline.py:10`
- `backend/ml/services/ab_testing.py:6`

Both have:
```python
from sqlalchemy.ext.declarative import declarative_base  # DEPRECATED
```

**Expected behavior**

```python
from sqlalchemy.orm import declarative_base  # CORRECT
```

**Actual behavior**

With SQLAlchemy 2.0+, these files raise `ModuleNotFoundError: No module named 'sqlalchemy.ext.declarative'`.

**Impact**

ML service crashes if SQLAlchemy >= 2.0 is installed. Currently works only because an older version is pinned, making upgrades risky.

**Suggested fix**

Replace `from sqlalchemy.ext.declarative import declarative_base` with `from sqlalchemy.orm import declarative_base` in both files.

---

vipul674:
Fixes #4727

### DIFF
--- a/backend/ml/services/ab_testing.py
+++ b/backend/ml/services/ab_testing.py
@@ -3,7 +3,7 @@ import random
 from datetime import datetime
 from typing import Dict, Any, Optional
 from sqlalchemy import create_engine, Column, String, Float, DateTime, Integer
-from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import declarative_base
 from sqlalchemy.orm import sessionmaker
 import pandas as pd
 import json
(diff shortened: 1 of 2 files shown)


vipul674:

PR #4798, opened 2026-07-25
fix: DirtyRenderer.mergeDirtyRects() drops all non-overlapping rects - canvas never invalidated

### LINKED ISSUE #4745: bug: DirtyRenderer.mergeDirtyRects() drops all non-overlapping rects — canvas never properly invalidated
**Describe the bug**

`backend/renderer/DirtyRenderer.js` `mergeDirtyRects()` has a control-flow bug. When rects merge, the outer loop breaks via `hasMerged=true`, but `newRects` is still empty — the merged result and all remaining unprocessed rects are silently dropped.

**To Reproduce**

1. Open `backend/renderer/DirtyRenderer.js:157-201`
2. The algorithm starts with `rects = [r0, r1, r2, r3]`
3. When `r0` overlaps `r1`, they merge → `hasMerged=true` → `break` outer loop
4. `newRects` is still `[]` (nothing was pushed)
5. `rects = newRects = []` — all 4 rects are lost

```js
while (true) {
  const newRects = [];
  let hasMerged = false;
  for (let i = 0; i < rects.length && !hasMerged; i++) {
    // ... if overlap found, merge and set hasMerged=true → breaks
  }
  if (!hasMerged) break;
  rects = newRects; // ← empty! All rects lost
}
```

**Expected behavior**

The merge should preserve the merged rect plus all remaining unprocessed rects in `newRects`.

**Actual behavior**

All dirty rects are dropped. The canvas is never properly invalidated.

**Affected files**

- `backend/renderer/DirtyRenderer.js:157-201`

**Impact**

Terminal output rendering and any canvas rendering using this method produces incorrect output. Visual display in the renderer is unreliable.

**Suggested fix**

Restructure the merge algorithm to preserve unprocessed rects alongside merged results. The merged rect and all rects not yet processed should be carried over to `newRects` before breaking.

---

vipul674:
Fixes #4745

### DIFF
--- a/backend/renderer/DirtyRenderer.js
+++ b/backend/renderer/DirtyRenderer.js
@@ -162,31 +162,28 @@ class DirtyRenderer {
         
         while (merged) {
             merged = false;
-            const newRects = [];
+            const mergedIndices = new Set();
+            const mergedRects = [];
             
             for (let i = 0; i < rects.length; i++) {
+                if (mergedIndices.has(i)) continue;
                 let mergedRect = rects[i];
-                let hasMerged = false;
                 
                 for (let j = i + 1; j < rects.length; j++) {
+                    if (mergedIndices.has(j)) continue;
                     if (this.rectsOverlap(mergedRect, rects[j])) {
                         mergedRect = this.mergeRects(mergedRect, rects[j]);
-                        hasMerged = true;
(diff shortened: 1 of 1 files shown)


vipul674:

PR #4799, opened 2026-07-25
fix: DAO service stores off-by-one proposal ID - on-chain and off-chain IDs out of sync

### LINKED ISSUE #4748: bug: DAO service stores off-by-one proposal ID — on-chain and off-chain IDs out of sync
**Describe the bug**

`backend/dao/dao.service.js` stores `getTotalProposals()` as the proposal ID after creation. But `getTotalProposals()` returns a count (e.g., 5), while the actual proposal ID is 0-indexed (e.g., ID 4). This creates a persistent off-by-one mismatch between Supabase-stored IDs and on-chain proposal IDs.

**To Reproduce**

1. Open `backend/dao/dao.service.js:85-88`:
```js
const receipt = await tx.wait();
const proposalId = await this.dao.getTotalProposals();
// If there are 5 proposals, getTotalProposals() returns 5
// But the newly created proposal has ID 4 (0-indexed)
```

**Expected behavior**

Parse the `ProposalCreated` event from the transaction receipt logs to get the exact on-chain proposal ID, or use `getTotalProposals() - 1`.

**Actual behavior**

Stored proposal IDs in Supabase are consistently 1 higher than the actual on-chain ID. Any subsequent query using the stored ID to look up on-chain data fails or returns the wrong proposal.

**Affected files**

- `backend/dao/dao.service.js:87, 169`

**Impact**

Proposal lookup by ID returns wrong data. Linked on-chain and off-chain records are permanently misaligned.

**Suggested fix**

Parse the `ProposalCreated` event from `receipt.logs` to extract the exact proposal ID emitted by the contract, and use that as the stored ID.

---

vipul674:
Fixes #4748

### DIFF
--- a/backend/dao/dao.service.js
+++ b/backend/dao/dao.service.js
@@ -84,7 +84,16 @@ class DAOService {
             );
             const receipt = await tx.wait();
 
-            const proposalId = await this.dao.getTotalProposals();
+            // Parse proposal ID from ProposalCreated event
+            const eventLog = receipt.logs.find(log => {
+                try {
+                    const parsed = this.dao.interface.parseLog(log);
+                    return parsed.name === 'ProposalCreated';
+                } catch { return false; }
+            });
+            const proposalId = eventLog
+                ? this.dao.interface.parseLog(eventLog).args[0]
+                : (await this.dao.getTotalProposals()) - 1n;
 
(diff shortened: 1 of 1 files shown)


vipul674:

PR #4800, opened 2026-07-25
fix: token.service.js purchaseFraction missing null check on getAsset() - TypeError

### LINKED ISSUE #4749: bug: token.service.js `purchaseFraction()` missing null check on `getAsset()` return value — TypeError
**Describe the bug**

`backend/tokenization/token.service.js` `purchaseFraction()` calls `this.getAsset(assetId)` at line 67, which can return `null` when the asset doesn't exist (error caught internally at line 229). Line 68 then accesses `asset.tokenPrice` without a null check, throwing `TypeError: Cannot read properties of null`.

**To Reproduce**

1. Open `backend/tokenization/token.service.js:67-68`:
```js
const asset = await this.getAsset(assetId);  // returns null on error
const totalCost = parseFloat(asset.tokenPrice) * amount;  // TypeError if null
```

2. `getAsset()` at line 229 catches errors and returns `null` silently.

**Expected behavior**

Add a null check after `getAsset()`:
```js
const asset = await this.getAsset(assetId);
if (!asset) throw new DomainError(404, { error: 'Asset not found' });
```

**Actual behavior**

`TypeError: Cannot read properties of null (reading 'tokenPrice')` with no clear error message.

**Affected files**

- `backend/tokenization/token.service.js:67-68`

**Impact**

Asset token purchases fail with a confusing error when the asset doesn't exist. Instead of a clear "Asset not found" error, users see an unhandled exception.

**Suggested fix**

Add `if (!asset) throw new DomainError(404, { error: 'Asset not found' });` after the `getAsset()` call.

---

vipul674:
Fixes #4749

### DIFF
--- a/backend/tokenization/token.service.js
+++ b/backend/tokenization/token.service.js
@@ -65,6 +65,9 @@ class TokenizationService {
     async purchaseFraction(assetId, amount, userAddress) {
         try {
             const asset = await this.getAsset(assetId);
+            if (!asset) {
+                throw new Error('Asset not found');
+            }
             const totalCost = parseFloat(asset.tokenPrice) * amount;
 
             const tx = await this.token.purchaseFraction(


vipul674:

PR #4802, opened 2026-07-25
fix: RenderScheduler.cancel() removes task from map before confirming queue splice

### LINKED ISSUE #4752: bug: RenderScheduler.cancel() removes task from map before confirming queue splice — task ID permanently lost
**Describe the bug**

`backend/scheduler/RenderScheduler.js` `cancel()` deletes the task ID from `taskMap` at line 116 before the queue `splice` at line 119. If the splice fails, the task object remains in the queue but the task ID is permanently removed from the map — making the task unfindable and un-cancelable.

**To Reproduce**

1. Open `backend/scheduler/RenderScheduler.js:104-129`
2. Line 116: `this.taskMap.delete(found.id);`
3. Line 119: `queue.splice(found.queueIndex, 1);` — if queue structure is damaged, this silently fails

**Expected behavior**

The task should be removed from the queue first, then the map. Or use atomic removal that cannot leave inconsistent state.

**Actual behavior**

If `splice` fails:
- Task ID is deleted from `taskMap` (line 116 already ran)
- Task object persists in the queue array
- No way to identify or remove the orphaned task
- Memory leak and potential for zombie task execution

**Affected files**

- `backend/scheduler/RenderScheduler.js:116`

**Impact**

Orphaned tasks in the queue that cannot be found or canceled. Potential for zombie tasks to execute despite being "canceled."

**Suggested fix**

Reverse the order: remove from the queue first (line 119), then delete from the map (line 116). Wrap both in a try-finally to handle errors gracefully.

---

vipul674:
Fixes #4752

### DIFF
--- a/backend/scheduler/RenderScheduler.js
+++ b/backend/scheduler/RenderScheduler.js
@@ -113,8 +113,8 @@ class RenderScheduler extends EventEmitter {
                 queue.splice(index, 1);
                 task.status = 'cancelled';
                 this.stats.cancelledTasks++;
-                this.taskMap.delete(taskId);
                 this.emit('taskCancelled', { taskId });
+                this.taskMap.delete(taskId);
                 logger.debug(`Task ${taskId} cancelled`);
                 return true;
             }


vipul674:

PR #4805, opened 2026-07-25
fix: cutomer_app ml_dashboard shows infinite loading spinner on non-200 API responses

### LINKED ISSUE #4754: bug: cutomer_app ml_dashboard.dart shows infinite loading spinner on non-200 API responses
**Describe the bug**

`cutomer_app/lib/screens/ml_dashboard.dart` only sets `isLoading = false` inside the `if (response.statusCode == 200)` branch. When the API returns a non-200 status (4xx, 5xx), `isLoading` stays `true` forever, showing an infinite `CircularProgressIndicator`.

**To Reproduce**

1. Open `cutomer_app/lib/screens/ml_dashboard.dart:30-35`:
```dart
if (response.statusCode == 200) {
  final data = json.decode(response.body);
  // ...
  isLoading = false;  // Only set in success branch
}
// No else branch — isLoading stays true forever
```

2. When the server returns 500 or any non-200 status, `isLoading` is never set to `false`.

**Expected behavior**

Add an `else` branch that handles errors and sets `isLoading = false`.

**Actual behavior**

User sees a permanent spinner with no error message or retry option.

**Affected files**

- `cutomer_app/lib/screens/ml_dashboard.dart:30-35`

**Impact**

- App appears frozen after any API failure
- No error recovery or retry available
- User cannot use the dashboard if the ML service returns an error

**Suggested fix**

Add an `else` block after `if (statusCode == 200)` that handles errors, sets `isLoading = false`, and shows an appropriate error UI.

---

vipul674:
Fixes #4754

### DIFF
--- a/cutomer_app/lib/screens/ml_dashboard.dart
+++ b/cutomer_app/lib/screens/ml_dashboard.dart
@@ -32,6 +32,10 @@ class _MLDashboardState extends State<MLDashboard> {
           metrics = json.decode(response.body);
           isLoading = false;
         });
+      } else {
+        setState(() {
+          isLoading = false;
+        });
       }
     } catch (e) {
       setState(() {


vipul674:

PR #4809, opened 2026-07-25
fix: Customer app public_tracking_screen unsafe as num cast - crashes on string coordinates

### LINKED ISSUE #4759: bug: Customer app public_tracking_screen.dart unsafe `as num` cast — crashes on unexpected coordinate types
**Describe the bug**

`apps/customer/lib/screens/public_tracking_screen.dart` uses `as num` casts on dynamic map values for latitude/longitude coordinates. If the backend returns coordinates as strings (e.g., `"28.6139"`) instead of numbers, the cast throws a `TypeError`.

**To Reproduce**

1. Open `apps/customer/lib/screens/public_tracking_screen.dart:136-155`:
```dart
LatLng? _getPickupLatLng() {
  final lat = _order?['pickup_lat'];
  final lng = _order?['pickup_lng'];
  if (lat == null || lng == null) return null;
  return LatLng((lat as num).toDouble(), (lng as num).toDouble());
}
```

The null check only verifies non-null but does not verify the type is `num`. If the value is `String`, `as num` throws `TypeError`.

**Expected behavior**

Use `(lat as num?)?.toDouble()` or `double.tryParse(lat.toString())` for safe conversion.

**Actual behavior**

Runtime crash with `TypeError: type 'String' is not a subtype of type 'num'` when backend returns coordinates in an unexpected format.

**Affected files**

- `apps/customer/lib/screens/public_tracking_screen.dart:136-155`

**Impact**

- Public tracking page crashes on unexpected API response formats
- No graceful fallback for coordinate display

**Suggested fix**

Use `double.tryParse(lat.toString())` instead of `(lat as num).toDouble()` to safely handle both numeric and string coordinate formats.

---

vipul674:
Fixes #4759

### DIFF
--- a/apps/customer/lib/screens/public_tracking_screen.dart
+++ b/apps/customer/lib/screens/public_tracking_screen.dart
@@ -137,21 +137,21 @@ class _PublicTrackingScreenState extends State<PublicTrackingScreen> {
     final lat = _order?['pickup_lat'];
     final lng = _order?['pickup_lng'];
     if (lat == null || lng == null) return null;
-    return LatLng((lat as num).toDouble(), (lng as num).toDouble());
+    return LatLng(double.tryParse(lat.toString()) ?? 0.0, double.tryParse(lng.toString()) ?? 0.0);
   }
 
   LatLng? _getDropLatLng() {
     final lat = _order?['drop_lat'];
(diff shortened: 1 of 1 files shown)


vipul674:

PR #4812, opened 2026-07-25
fix: race condition in tracker.js - flushMutex released before async insert

### LINKED ISSUE #4764: bug: tracker.js flushMutex released before async MongoDB flush — race condition on telemetry buffer
**Describe the bug**

`backend/api/src/sockets/tracker.js` sets `flushMutex = false` on line 822 before the async MongoDB insert starts on line 829. The mutex is meant to prevent concurrent flush operations, but releasing it before the async write allows a second flush to run in parallel with the first one.

**To Reproduce**

1. Open `backend/api/src/sockets/tracker.js`:
- Line 808: `flushMutex = true;` (lock acquired)
- Line 822: `flushMutex = false;` (lock released before async write)
- Line 829: `const result = await db.collection('telemetry').insertMany(flushSet);` (async write starts)
- Line 872: `flushMutex = false;` (redundant release in finally)

**Expected behavior**

The mutex should be released only after the async database write completes, either on success (in the try block) or in the finally block.

**Actual behavior**

Between line 822 and line 829, a concurrent event loop cycle sees `flushMutex === false` and starts a second flush, causing overlapping MongoDB writes.

**Affected files**

- `backend/api/src/sockets/tracker.js:808-872`

**Impact**

- Potential race condition on MongoDB telemetry buffer
- Duplicate or corrupted telemetry data from overlapping flush operations
- No actual concurrency protection despite the mutex

**Suggested fix**

Remove the premature `flushMutex = false` at line 822. Keep only the finally block release at line 872.

---

vipul674:
Fixes #4764

### DIFF
--- a/backend/api/src/sockets/tracker.js
+++ b/backend/api/src/sockets/tracker.js
@@ -819,8 +819,6 @@ async function flushTelemetryBuffer() {
   telemetryFlushBuffer = [];
   telemetryWriteBuffer.clear();
 
-  flushMutex = false;
-
   if (recordsToFlush.length === 0) {
     flushMutex = false;
     return;


vipul674:

PR #4919, opened 2026-07-26
Refactor rate limiter middleware for clarity and efficiency

### LINKED ISSUE #4588: Bug: `authLimiter` references undefined variables, crashes module on import

ishitaajain22-tech:
### Problem
`isRedisReady()` was missing its closing brace. `isSuspiciousForwardedHeader`
was defined *inside* it instead of at module scope:
```js
function isRedisReady() {
  function isSuspiciousForwardedHeader(header) { ... }
// <- isRedisReady() never closes here
```
Everything after this point in the file — including the `export function
safeIpKeyGenerator(...)` a few lines down — ended up nested inside
`isRedisReady()`'s body. `export` is illegal anywhere except a module's top
level, so **the file threw a `SyntaxError` and failed to load at all**
(confirmed with `node --check`). Since this module is imported by the auth
route chain, this breaks the whole API process on startup.

A second, independent bug was also present in `authLimiter`: its config
object referenced undefined variables `authWindowMs` / `authMaxRequests`
(leftovers from before the file switched to the uppercase, env-driven
`AUTH_WINDOW_MS` / `AUTH_MAX_REQUESTS` constants), then silently overwrote
them with duplicate object keys. Even after fixing the syntax error, this
would throw `ReferenceError: authWindowMs is not defined` the moment the
module evaluates `authLimiter`, and again in the `handler` callback on the
first rate-limited request.

### Fix

Fixes #4588 
- Moved `isSuspiciousForwardedHeader` to its own top-level function and gave
  `isRedisReady()` a real body: `return !!(redisClient && redisClient.status
  === 'ready')`.
- Removed the dead `authWindowMs` / `authMaxRequests` lines from
  `authLimiter` and pointed the `handler`'s `retryAfter` at `AUTH_WINDOW_MS`.

### Impact
The module now parses and loads correctly, `isRedisReady()` actually reports
Redis connection state (previously would have always fallen back to
`MemoryStore` even after fixing the parse error, since it had no logic at
all), and `authLimiter` no longer throws on construction or on first use.

### Testing
- `node --check src/middleware/rateLimiter.js` now passes.
- Manually hit an auth endpoint past `AUTH_MAX_REQUESTS` and confirmed the
  429 handler responds with the correct `retryAfter` instead of throwing.

### DIFF
--- a/backend/api/src/middleware/rateLimiter.js
+++ b/backend/api/src/middleware/rateLimiter.js
@@ -3,6 +3,10 @@ import { RedisStore } from 'rate-limit-redis';
 import { redisClient } from '../config/db.js';
 import logger from './logger.js';
 
+function isRedisReady() {
+  return !!(redisClient && redisClient.status === 'ready');
+}
+
 function isSuspiciousForwardedHeader(header) {
   if (!header || typeof header !== 'string') return false;
 
@@ -13,13 +17,6 @@ function isSuspiciousForwardedHeader(header) {
 
   // Reject obviously malformed values.
   return parts.some((ip) => ip.length === 0 || ip.includes('\n') || ip.includes('\r'));
-}
-
-function isRedisReady() {
-  return !!(
-    redisClient?.status === 'ready' &&
-    redisClient?.isOpen
-  );
 }
 
 /**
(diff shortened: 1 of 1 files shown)


ishitaajain22-tech:
> 

GSSOC

PR #4920, opened 2026-07-26
 Fix: naive/aware datetime comparison silently dropped valid loads

### LINKED ISSUE #4589: Bug: naive/aware datetime comparison can silently drop valid loads

ishitaajain22-tech:
### Problem
`arrival_dt` and `deadline_dt` were each parsed independently with
`datetime.fromisoformat(...)`. Whichever one carried a UTC offset (e.g. a
`pickup_deadline` ending in `Z` or `+05:30`) came back timezone-**aware**,
while the other (especially the `datetime.now()` fallback for a bad
`arrival_time`) stayed **naive**. Comparing `estimated_arrival > deadline_dt`
across a naive/aware pair raises `TypeError`.

Fixes #4589 

That `TypeError` was swallowed by the per-load `except Exception` block
(logged as a warning, then `continue`), so the failure was invisible: any
load with a timezone-aware `pickup_deadline` was silently excluded from
`find_return_loads()`'s recommendations whenever the driver's `arrival_time`
was naive.

### Fix
Added `_to_naive()`, which strips `tzinfo` if present, and applied it right
after parsing both `arrival_dt` and `deadline_dt`. Both datetimes are now
guaranteed to be naive before any comparison or subtraction happens.

### Impact
Loads with tz-aware deadlines are no longer silently dropped from return-load
recommendations — this was quietly starving drivers of legitimate matches
whenever upstream services sent offset-qualified timestamps.

### Testing
- Called `find_return_loads` with an aware `pickup_deadline`
  (`...+05:30`) and a naive `arrival_time`; confirmed the load is now
  scored instead of being silently skipped.
- Existing all-naive and all-aware input combinations still behave as before.


ishitaajain22-tech:
@KanishJebaMathewM pls assign me this

PR #5001, opened 2026-07-26
fix: check in-memory behavioralProfiles in getOrCreateProfile

### LINKED ISSUE #4997: bug: FraudDetectionService getOrCreateProfile Never Reads from In-Memory Map
## Description

In `backend/api/src/services/fraud/FraudDetectionService.js`, `trackBehavior` (line 62) writes to `this.behavioralProfiles.set(deviceId, ...)` when Redis is unavailable. However, `getOrCreateProfile` (lines 102-123) only queries Redis and the database — it **never checks** `this.behavioralProfiles`:

```javascript
// FraudDetectionService.js:102-123
async getOrCreateProfile(deviceId) {
    // Checks Redis...
    // Checks database...
    // Never checks this.behavioralProfiles Map
}
```

Profiles written by `trackBehavior` into the in-memory Map are effectively **write-only** — they can never be read back by the same process. This means fraud detection during Redis outages produces profiles that are immediately invisible to subsequent lookups.

## Expected Behaviour

`getOrCreateProfile` should check the in-memory `behavioralProfiles` Map as a fallback before hitting the database.

## Actual Behaviour

In-memory profiles are write-only. During Redis outages, fraud profiles are created but never retrievable.

## Proposed Fix

```javascript
// FraudDetectionService.js — Add in-memory fallback in getOrCreateProfile
async getOrCreateProfile(deviceId) {
    // ... existing Redis check ...

    // Check in-memory cache (written by trackBehavior during Redis outages)
    const inMemory = this.behavioralProfiles.get(deviceId);
    if (inMemory) return inMemory;

    // ... existing database check ...
}
```

ionfwsrijan:
## Summary
Fixes #4997 — `getOrCreateProfile` only checked Redis and the database but never the in-memory `behavioralProfiles` Map. Profiles written by `trackBehavior` during Redis outages were effectively write-only.

## Changes
- `FraudDetectionService.js`: Added `this.behavioralProfiles.get(userId)` check between Redis cache and database lookup.

## Test plan
- [ ] During Redis outage, profiles written by `trackBehavior` are retrievable by `getOrCreateProfile`.
- [ ] Normal flow (Redis available) is unaffected.

### DIFF
--- a/backend/api/src/services/fraud/FraudDetectionService.js
+++ b/backend/api/src/services/fraud/FraudDetectionService.js
@@ -126,6 +126,10 @@ class FraudDetectionService {
       return JSON.parse(cached);
     }
 
+    // Check in-memory cache (written by trackBehavior during Redis outages)
+    const inMemory = this.behavioralProfiles.get(userId);
+    if (inMemory) return inMemory;
+
     // Check database
     const { data } = await supabase
       .from('behavioral_profiles')


ionfwsrijan:

PR #5002, opened 2026-07-26
fix: cap completed array at 1000 entries to prevent memory leak

### LINKED ISSUE #4998: bug: RenderScheduler `completed` Array Grows Without Bound (Memory Leak)
## Description

In `backend/scheduler/RenderScheduler.js`, every completed task is pushed onto `this.completed` (line 271). The array is only ever cleared inside `reset()` (line ~157). Over a long-running process, the array grows unboundedly, accumulating stale task objects and eventually consuming significant memory.

```javascript
// RenderScheduler.js:271
this.completed.push(task);
// ... this.completed is never trimmed
```

In a production scheduler running for days/weeks, this array can accumulate thousands of task objects with their full payload data.

## Expected Behaviour

The `completed` array should be capped at a maximum size, with old entries evicted when the cap is reached.

## Actual Behaviour

Every completed task is appended forever. The array grows without limit until the process is restarted or `reset()` is called.

## Proposed Fix

```javascript
// RenderScheduler.js — Add a max completed size constant
const MAX_COMPLETED = 1000;

// After line 271:
this.completed.push(task);
if (this.completed.length > MAX_COMPLETED) {
    this.completed = this.completed.slice(-MAX_COMPLETED);
}
```

ionfwsrijan:
## Summary
Fixes #4998 — The `completed` array in `RenderScheduler` grew without bound, causing memory leaks in long-running processes.

## Changes
- `RenderScheduler.js`: Added a cap of 1000 entries after each `push()`. When exceeded, the oldest entries are evicted via `slice(-1000)`.

## Test plan
- [ ] `completed` array never exceeds 1000 entries.
- [ ] `getCompletedTasks()` still returns the most recent entries correctly.

### DIFF
--- a/backend/scheduler/RenderScheduler.js
+++ b/backend/scheduler/RenderScheduler.js
@@ -269,6 +269,9 @@ class RenderScheduler extends EventEmitter {
             
             this.running.delete(task.id);
             this.completed.push(task);
+            if (this.completed.length > 1000) {
+                this.completed = this.completed.slice(-1000);
+            }
             this.stats.completedTasks++;
             
             // Update stats


ionfwsrijan:

PR #5062, opened 2026-07-27
bugfix: fix live tracking map updates and de-race websocket auth

Nareshkumawat-star:
# Description

This PR fixes the bug where the Live Tracking truck marker stays fixed at its initial position as the driver moves.

---

### 1. WebSocket Server (`backend/api/src/sockets/tracker.js`)
- Fixed a race condition where the client sent `subscribe_tracking` requests immediately upon connecting, which were processed and rejected before asynchronous auth/Supabase lookups were completed.
- Ensured the `message` event listener is only registered once authentication has fully succeeded.

### 2. Flutter Customer App (`apps/customer/lib/screens/live_tracking_screen.dart`)
- Added the missing `MapController` to the `LiveTrackingScreenState`.
- Wired the controller into the `FlutterMap` widget options.
- Updated `_updateTruckPosition` to pan/move the map camera to keep the moving truck centered on the map.


Nareshkumawat-star:
> 

GSSOC

Issue #2612, opened 2026-07-07
[BUG]: Race condition in Escrow payout when driver and customer complete trip simultaneously


anuragbraveboy-sudo:
**Describe the bug**
If both the Driver and the Customer tap the "Complete Trip / Confirm Delivery" button at the exact same millisecond, the backend attempts to trigger the `releaseFunds` smart contract function twice. This results in the second transaction failing and throwing an unhandled Revert exception in the backend.

**To Reproduce**
1. Go to the active trip screen on both the Driver and Customer devices.
2. Click on 'Confirm Delivery' at the exact same time.
3. See error: The backend logs a severe Ethereum revert error `Transaction reverted: funds already released`, which bubbles up as an HTTP 500 error to one of the users.

**Expected behavior**
The backend should use a database-level lock or a strict state machine update (`findOneAndUpdate` with `status: 'IN_PROGRESS'`) to ensure that only the first received request triggers the blockchain transaction. The second request should gracefully return a `400 Bad Request` stating "Trip already completed."

**Screenshots**
*(N/A)*

**Environment (please complete the following information):**
 - OS: Node.js Backend
 - App Version: API v1
 - Browser: N/A

**Additional context**
Blockchain transactions are expensive; duplicate calls waste gas and bloat the backend error logs.

PR #5167, opened 2026-07-27
fix: Implement strict database-level lock to prevent double payout race condition on trip completion (#2612)

anuragbraveboy-sudo:
### Description
This PR resolves a race condition where simultaneous clicks of the "Confirm Delivery" button by both the driver and customer would attempt to trigger the blockchain `releaseFunds` transaction twice.

### Changes Included
* Replaced the flawed `updateOrderGuardStatus` check in `deliveryVerificationService.js` with a strict atomic state machine lock.
* The update explicitly ensures the trip status hasn't changed before continuing the payout flow.
* If a concurrent request fails to acquire the lock, it is gracefully caught and returns a `400 Bad Request: Trip already completed`, rather than throwing an unhandled Revert exception.

### Notes for Reviewers
* This PR consists of exactly **1 commit**.
* Fixed a leftover syntax error (`guardErr`) in the old verification block.


anuragbraveboy-sudo:
gssoc

Issue #5008, opened 2026-07-26
[Smart Contract] Escrow dispute resolution timeout logic in Polygon contract


anuragbraveboy-sudo:
**Labels:** `blockchain`, `smart-contracts`, `phase-4`

**Description:**
The Polygon smart contracts handle escrow for freight payments. We need a fallback timeout logic in case a dispute is raised but neither party responds or resolves it via the n8n automation pipeline.

**Tasks:**
- Add a time-lock modifier to the escrow release function in the Solidity contract.
- If a dispute is inactive for X days, default to refunding the customer or splitting the escrow based on the ML distance calculation.
- Write Hardhat/Foundry tests to verify the timeout logic.

PR #5181, opened 2026-07-27
feat: Add dispute timeout logic to escrow contract (#5008)

anuragbraveboy-sudo:
### Description
Prevents a scenario where funds are permanently locked in the Polygon escrow contract if a dispute is raised but the resolution pipeline fails or both parties go unresponsive.

### Changes Included
* **`TruxifyEscrow.sol`:** Added a 7-day `DISPUTE_TIMEOUT` time-lock modifier. If a dispute sits idle for 7 days, `resolveDisputeTimeout()` can be called to default to a customer refund.
* **`TruxifyEscrow.test.js`:** Added time-manipulation tests using Hardhat network helpers to fast-forward the EVM and verify both successful timeout triggers and pre-timeout reverts.

### Notes for Reviewers
* This PR is contained in exactly **1 commit**.
* Re-entrancy guards and Checks-Effects-Interactions (CEI) patterns are strictly maintained in the new resolution function.


anuragbraveboy-sudo:
gssoc

PR #5294, opened 2026-07-28
fix: prevent concurrent modification of listeners array during publish

### LINKED ISSUE #5268: bug: src/lib/eventBus.ts concurrent modification of listeners array during publish() iteration — stale callbacks fire
## Summary

In `src/lib/eventBus.ts:5-8, 18-27`, if a callback during `publish()` calls `unsubscribe()` for the same event, the listeners array is replaced with a new filtered array. The in-progress `forEach` continues iterating the OLD array. Remaining callbacks after the removed one in the OLD array still fire even though they should have been removed, and the new filtered array is never used by the current publish cycle.

## Evidence

```typescript
// src/lib/eventBus.ts
subscribe(event, callback) {
    this.listeners.get(event)!.push(callback);
}

unsubscribe(event, callback) {
    const filtered = this.listeners.get(event)!.filter(cb => cb !== callback);
    this.listeners.set(event, filtered);   // Replaces the array during iteration
}

publish(event, data) {
    this.listeners.get(event)!.forEach(cb => {
        cb(data);   // If cb calls unsubscribe(), the array is replaced mid-iteration
    });
}
```

## Affected files

- `src/lib/eventBus.ts`

## Reproduction steps

1. Subscribe callbacks A, B, C to the same event
2. Publish the event
3. During A's execution, A calls `unsubscribe(event, B)`
4. B's callback is removed from the listeners map
5. But the in-progress `forEach` still calls B (because it's iterating the old array)
6. C also fires on the old array

## Expected behavior

Unsubscribed callbacks should not fire if they unsubscribe during the current publish cycle.

## Actual behavior

Callbacks removed during iteration are still called because `forEach` holds a reference to the old array.

## Impact

Event bus subscribers may receive events after they've unsubscribed. This can cause unexpected behavior, memory leaks, or crashes in components that expect to be idle after unsubscribing.

## Suggested fix

Use a defensive copy during iteration:

```typescript
publish(event, data) {
    const listeners = this.listeners.get(event);
    if (listeners) {
        [...listeners].forEach(cb => cb(data));
    }
}
```

## GSSoC 2026

vipul674:
Fixes #5268

## GSSoC 2026

### DIFF
--- a/src/lib/eventBus.ts
+++ b/src/lib/eventBus.ts
@@ -17,7 +17,7 @@ export class EventBus<T = any> {
   }
   publish(event: string, data: T) {
     if (this.listeners.has(event)) {
-      this.listeners.get(event)!.forEach(cb => {
+      [...this.listeners.get(event)!].forEach(cb => {
         try {
           cb(data);
         } catch (error) {


vipul674:

vipul674:

PR #5297, opened 2026-07-28
fix: add await to transaction.commit() in GlobalStore.js

### LINKED ISSUE #5220: bug: GlobalStore.js transaction.commit() called without await in synchronous code path — race condition on shared state
## Summary

In `backend/store/GlobalStore.js:224-226`, `transaction.commit()` is an async function but is called without `await` in the synchronous code path (when `fn(transaction)` does not return a promise). The returned Promise is silently dropped, and `this.activeTransaction` is set to `null` before `commit` finishes. If another `transaction()` call arrives, it sees no active transaction and starts a new one — both transactions operate on shared state concurrently.

## Evidence

```javascript
// backend/store/GlobalStore.js, lines 224-226
// Sync result
transaction.commit();   // MISSING 'await'
this.activeTransaction = null;
```

The `commit()` method at lines 55-76 is declared `async` and performs state modifications. Without `await`, the commit runs in the background while `activeTransaction` is immediately cleared.

## Affected files

- `backend/store/GlobalStore.js`

## Reproduction steps

1. Call `transaction()` with a synchronous callback
2. Before the async commit completes, call `transaction()` again
3. The second call sees `activeTransaction = null` and starts a new transaction
4. Both transactions concurrently modify `this.state`

## Expected behavior

The async commit should complete before `activeTransaction` is cleared. The second `transaction()` call should either queue or wait.

## Actual behavior

The commit runs in the background. The next transaction can interleave with the in-flight commit, causing race conditions on shared state.

## Impact

Race condition on `this.state` — two transactions can interleave, causing data corruption. The GlobalStore is the foundation for state management across several microservices.

## Suggested fix

Add `await` before `transaction.commit()` in the synchronous branch.

## GSSoC 2026

vipul674:
Fixes #5220

## GSSoC 2026

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -180,7 +180,6 @@ import {
 import { getEscrowBookingId } from '../services/escrow.js';
 import { getRouteEstimate, getRouteGeometry, buildStraightLineGeometry } from '../services/osrm.js';
 import { computeOrderPricing } from '../lib/pricing.js';
-import { getEscrowBookingId } from '../services/escrow.js';
 
 const router = express.Router();
 router.use(userLimiter);
--- a/backend/store/GlobalStore.js
+++ b/backend/store/GlobalStore.js
@@ -222,7 +222,7 @@ class GlobalStore extends EventEmitter {
             }
             
             // Sync result
-            transaction.commit();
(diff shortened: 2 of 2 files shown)


vipul674:

vipul674:

PR #5315, opened 2026-07-28
fix: replace counter-based product ID with UUID to prevent race conditions

### LINKED ISSUE #5225: bug: trace.service.js product ID retrieved via counter instead of event log — race condition on concurrent creates
## Summary

In `backend/traceability/trace.service.js:47`, `createProduct` fetches the new product ID by calling `getProductCount()` after `tx.wait()`. If two transactions are processed simultaneously, the counter could advance between `tx.wait()` and this call, returning a wrong ID. Contrast with `createShipment` (lines 80-94) which correctly parses the event log.

## Evidence

```javascript
// trace.service.js, line 47 — uses counter (buggy)
const productId = await this.contract.getProductCount();
```

vs `createShipment` which correctly uses event parsing.

## Affected files

- `backend/traceability/trace.service.js`

## Reproduction steps

1. Send two concurrent `createProduct` requests
2. Both blockchain transactions confirm
3. Both `getProductCount()` calls may return the same value (before either is fully indexed)
4. Both products get the same ID, causing data collision

## Expected behavior

The product ID should be parsed from the transaction receipt event logs, not from a live counter query.

## Actual behavior

The live counter query is racy and can return incorrect values under concurrent load.

## Impact

Product IDs can collide, causing database constraint violations or incorrect traceability records under concurrent load.

## Suggested fix

Parse the product ID from `receipt.logs` using the event ABI, consistent with how `createShipment` generates shipment IDs.

## GSSoC 2026

vipul674:
Fixes #5225

## GSSoC 2026

### DIFF
--- a/backend/traceability/trace.service.js
+++ b/backend/traceability/trace.service.js
@@ -44,18 +44,18 @@ class TraceabilityService {
             );
             const receipt = await tx.wait();
 
-            const productId = await this.contract.getProductCount();
+            const productId = uuidv4();
 
             await this.storeProduct({
                 ...productData,
-                productId: productId.toString(),
+                productId,
                 txHash: receipt.hash
             });
 
             logger.info(`✅ Product created: ${productId}`);
             return {
                 success: true,
-                productId: productId.toString(),
+                productId,
                 productHash,
                 txHash: receipt.hash
             };


vipul674:

vipul674:

PR #5365, opened 2026-07-28
fix(core): release escrow reconciliation locks on completion to prevent order deadlock

### LINKED ISSUE #5364: bug(core): escrowReleaseReconciliation permanently deadlocks orders due to incorrect reconciled_by filter
**Describe the bug**
The background worker responsible for reconciling and retrying failed escrow payments (`backend/api/src/services/escrowReleaseReconciliation.js`) contains a critical concurrency flaw that permanently deadlocks orders and prevents drivers from being paid.

When the worker claims an order, the `claim_release_reconciliation` RPC sets `reconciled_by = instanceId`. However, the subsequent success update block attempts to update the order with the filter `.is('reconciled_by', null)`. Because the order is already claimed, this update affects 0 rows. The order remains in `release_failed` state in the database despite the blockchain transaction occurring.

Additionally, both the success and error blocks fail to reset `reconciled_by = null`. As a result, the order becomes permanently locked to that `instanceId` and will be silently ignored by all future reconciliation runs. The retry mechanism (up to `MAX_RETRIES`) is completely broken.

**To Reproduce**
1. Trigger a failed escrow release so an order is in `release_failed` state.
2. The `reconcilePendingEscrowReleases` background job picks it up.
3. The job successfully claims the order (`reconciled_by` is set).
4. The job calls the blockchain and attempts to update the database to `released`.
5. The update silently fails due to `.is('reconciled_by', null)`.
6. The job completes, leaving the order stuck forever with `reconciled_by` populated.

**Expected behavior**
The success update block must filter by `.eq('reconciled_by', instanceId)` instead of `.is('reconciled_by', null)`.
Both the success block and the error catch block must explicitly include `reconciled_by: null` in their update payloads so the lock is released for future retries.

**Environment:**
 - OS: All
 - Backend: Node.js (Express worker)
 - Module: `escrowReleaseReconciliation.js`

**Additional context**
This is a Level 3 Core Architecture bug involving distributed lock management and financial transaction state machines. It must be fixed to ensure drivers reliably receive their payouts.

Unnati1007:
## Summary
Patched a highly critical concurrency flaw in the background escrow reconciliation workers that caused permanently deadlocked orders and prevented driver payouts from being retried.

## Motivation
Closes #5364.

Previously, the `escrowReleaseReconciliation.js` and `escrowRefundReconciliation.js` background workers correctly claimed failed escrow transfers by setting `reconciled_by = instanceId` via an RPC. However, when the workers attempted to update the database after communicating with the blockchain, they incorrectly tried to filter the update with `.is('reconciled_by', null)`. Since the lock was held by the worker, the update silently affected 0 rows. Even worse, the worker never explicitly cleared `reconciled_by`, meaning the order became permanently locked and ignored by all future reconciliation runs. This permanently broke the retry mechanism.

## Changes
- Modified `backend/api/src/services/escrowReleaseReconciliation.js` and `backend/api/src/services/escrowRefundReconciliation.js`:
  - Replaced the erroneous `.is('reconciled_by', null)` filter with `.eq('reconciled_by', instanceId)` in the success block.
  - Added `reconciled_by: null` to the success update payloads so the lock is successfully released.
  - Added `reconciled_by: null` to the error catch-block payloads to guarantee the lock is released if the blockchain transaction fails, allowing the standard `MAX_RETRIES` exponential backoff mechanism to function.

## Acceptance Criteria
- [x] Orders are no longer permanently locked to an `instanceId` after a reconciliation run.
- [x] The success update correctly clears the lock.
- [x] The failure catch block correctly clears the lock.

## Impact & Side Effects
Strict enforcement of database state machine correctness. No negative impact. This ensures drivers get paid and refunds get issued properly when the Polygon RPC endpoints experience transient failures.

## Quality Checklist
- [x] Linter passed
- [x] Types checked
- [x] Unit tests passed (Note: existing CI tests may fail due to unrelated bugs in `main`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved escrow refund and release reconciliation to prevent updates from being applied to records managed by another processing instance.
  * Escrow records now correctly become available for subsequent processing after successful completion or a failed attempt.
  * Improved retry handling by reliably recording errors and releasing processing ownership.

### DIFF
--- a/backend/api/src/services/escrowRefundReconciliation.js
+++ b/backend/api/src/services/escrowRefundReconciliation.js
@@ -140,8 +140,9 @@ export async function reconcilePendingEscrowRefunds(orderRepository) {
           refund_tx_hash: receipt.hash ?? refundTxHash,
           escrow_refunded_at: refundedAt,
           escrow_refund_error: null,
+          reconciled_by: null,
           updated_at: refundedAt,
-        }, [{ op: 'in', column: 'escrow_status', value: ['refund_pending', 'refund_failed'] }], 'id');
+        }, [{ op: 'in', column: 'escrow_status', value: ['refund_pending', 'refund_failed'] }, { op: 'eq', column: 'reconciled_by', value: instanceId }], 'id');
 
         if (updateError) {
           logger.error(
@@ -154,6 +155,7 @@ export async function reconcilePendingEscrowRefunds(orderRepository) {
         await orderRepository.updateOrder(order.id, {
           escrow_refund_retry_count: newRetryCount,
           escrow_refund_error: err.message,
+          reconciled_by: null,
           updated_at: new Date().toISOString(),
         });
         logger.warn(
--- a/backend/api/src/services/escrowReleaseReconciliation.js
+++ b/backend/api/src/services/escrowReleaseReconciliation.js
@@ -104,11 +104,12 @@ export async function reconcilePendingEscrowReleases() {
             escrow_released_at: releasedAt,
             escrow_release_attempts: releaseAttempts,
             escrow_release_last_attempt_at: releaseAttemptedAt,
+            reconciled_by: null,
             updated_at: releasedAt,
           })
           .eq('id', order.id)
-      .in('escrow_status', ['release_failed', 'funded'])
-          .is('reconciled_by', null);
+          .in('escrow_status', ['release_failed', 'funded'])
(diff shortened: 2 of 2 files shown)


Unnati1007:
@KanishJebaMathewM Please review the PR, do lemme know if any changes needed and merge it under GSSoC'26

PR #5408, opened 2026-07-29
fix: move custom error handlers before generic handler to prevent dead code (#5267)

### LINKED ISSUE #5267: fix: error handlers for entity.too.large and SyntaxError registered after generic errorHandler - dead code
## Description

In `index.js`, the custom error handlers for `entity.too.large` and `SyntaxError` are registered **after** the generic `errorHandler` middleware. Since the generic handler always sends a 500 response and never calls `next(err)`, the custom handlers never execute.

## Current Code

```javascript
// backend/api/src/index.js

// Lines ~569-572: Generic handlers registered FIRST
app.use(sentryErrorHandler());
app.use(errorHandler);

// Lines ~701-738: Custom handlers registered AFTER
app.use((err, req, res, next) => {
    if (err.type === 'entity.too.large') {
        return res.status(413).json({ error: 'Payload too large' });
    }
    if (err instanceof SyntaxError) {
        return res.status(400).json({ error: 'Invalid JSON' });
    }
    next(err);
});
```

## Problem

Express processes middleware in registration order. When `errorHandler` catches an error, it sends a 500 response and does NOT call `next(err)`. The custom handlers at lines 701-738 are never reached, so:

- Large payload errors get a generic 500 instead of a proper 413
- Malformed JSON gets a generic 500 instead of a proper 400

## Expected Behavior

Custom error handlers should be registered BEFORE the generic catch-all handler.

## Suggested Fix

Move the custom error handlers to before `sentryErrorHandler()` and `errorHandler`:

```javascript
// Register custom handlers FIRST
app.use((err, req, res, next) => {
    if (err.type === 'entity.too.large') {
        return res.status(413).json({ error: 'Payload too large' });
    }
    if (err instanceof SyntaxError) {
        return res.status(400).json({ error: 'Invalid JSON' });
    }
    next(err);
});

// Then register generic handlers
app.use(sentryErrorHandler());
app.use(errorHandler);
```

## Impact

- Payload size limit errors produce misleading 500 instead of 413
- Malformed JSON produces misleading 500 instead of 400
- Debugging production issues becomes harder
- Client applications cannot distinguish error types

Akshita-2307:
## Description

Custom error handlers for entity.too.large and SyntaxError were registered AFTER the generic errorHandler. Since the generic handler always sends a 500 and never calls next(err), the custom handlers never executed.

## Fix

Moved the custom error handlers to before the sentryErrorHandler and errorHandler, so they run first and can provide proper 413/400 responses instead of generic 500s.

Closes #5267
GSSoC

### DIFF
--- a/backend/api/src/middleware/errorHandler.js
+++ b/backend/api/src/middleware/errorHandler.js
@@ -2,6 +2,28 @@ import logger from './logger.js';
 import { AppError } from '../utils/errors.js';
 
 export function errorHandler(err, req, res, next) {
+  if (err?.type === 'entity.too.large') {
+    logger.warn(
+      { requestId: req.requestId, ip: req.ip, method: req.method, path: req.originalUrl },
+      'Request payload exceeded configured limit'
+    );
+    return res.status(413).json({
+      success: false,
+      error: 'Payload too large'
+    });
+  }
+
+  if (err instanceof SyntaxError && err.status === 400 && 'body' in err) {
+    logger.warn(
+      { requestId: req.requestId, ip: req.ip, method: req.method, path: req.originalUrl },
+      'Malformed JSON payload received'
+    );
+    return res.status(400).json({
+      success: false,
+      error: 'Malformed JSON payload'
+    });
+  }
+
   if (err && err.name === 'MulterError') {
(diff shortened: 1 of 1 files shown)


KanishJebaMathewM:
@Akshita-2307 Resolve the conflicts so that i can merge this

Issue #5439, opened 2026-07-29
fix: destination_picker_screen catch(_) silently swallows reverse-geocoding errors with no logging


Xenon010101:
## Bug Description

`apps/driver/lib/screens/destination_picker_screen.dart` line 158 has a `catch (_)` block that silently swallows reverse-geocoding errors:

```dart
} catch (_) {
  if (!mounted) { return; }
  setState(() {
    _selectedAddress = 'Pinned location (${point.latitude.toStringAsFixed(5)}, ${point.longitude.toStringAsFixed(5)})';
  });
}
```

When `GeocodeService.resolvePlace` fails (network error, API timeout), the actual error object is completely discarded. The screen falls back to showing raw coordinates, which is acceptable behavior, but the root cause is lost with no diagnostic logging.

### Expected
Replace `catch (_)` with `catch (e)` and add `debugPrint` to log the error while preserving the existing fallback behavior.

PR #5510, opened 2026-07-30
fix: destination_picker_screen catch(_) silently swallows reverse-geocoding errors with no logging (#5439)

Xenon010101:
GSSOC


Xenon010101:
Closing — duplicate of #5440

Issue #5432, opened 2026-07-29
fix: resilient_websocket catch(_) silently swallows connection errors with no logging


Xenon010101:
## Bug Description

`apps/customer/lib/core/offline/websocket/resilient_websocket.dart` line 54 silently swallows WebSocket connection errors with no diagnostic logging:

```dart
} catch (_) {
  _scheduleReconnect();
}
```

When `_connectOnce()` fails (e.g., network unreachable, DNS resolution failure, server down), the error is completely discarded with zero visibility. The only observable effect is a reconnection attempt with no indication of what went wrong.

### Impact
- Developers have no way to diagnose why WebSocket connections fail
- Reconnection loops without any logged root cause make debugging production issues much harder
- The error object contains valuable diagnostic information (network type, DNS status, HTTP status code) that's lost

### Expected
Replace `catch (_)` with `catch (e, st)` and add a `debugPrint` call:
```
} catch (e, st) {
  debugPrint('[ResilientWebSocket] Connection failed: $e\n$st');
  _scheduleReconnect();
}
```

PR #5513, opened 2026-07-30
fix: resilient_websocket catch(_) silently swallows connection errors with no logging (#5432)

Xenon010101:
GSSOC


Xenon010101:
Closing — duplicate of #5438

PR #5534, opened 2026-07-30
fix: sanitize userId in FraudDetectionService Supabase queries

### LINKED ISSUE #5528: bug: FraudDetectionService Supabase Query Injection via Unsanitized Dynamic Filter Construction
## Description
The `getUserConnections` and `getBatchUserConnections` methods in `backend/api/src/services/fraud/FraudDetectionService.js` build Supabase filter expressions by directly concatenating user-provided values into query strings without any sanitization or escaping.

**`getUserConnections` (lines 300–327):**
```js
const { data: orders, error } = await supabase
  .from('orders')
  .select('customer_id, driver_id')
  .or(`customer_id.eq.${userId},driver_id.eq.${userId}`);
```
`userId` is embedded directly into the `.or()` filter string. If `userId` contains a comma (`,`), it can alter the filter structure. If it contains a closing parenthesis followed by additional filter operators, an attacker can manipulate the query to bypass row-level security or enumerate other users' data.

**`getBatchUserConnections` (lines 329–365):**
```js
.or(userIds.map(id => `customer_id.eq.${id}`).join(',') + ',' 
  + userIds.map(id => `driver_id.eq.${id}`).join(','));
```
Multiple user IDs are mass-concatenated. A crafted userId like `12345),driver_id.eq.(select(*)from` could inject arbitrary filter logic into the PostgREST query.

An attacker could enumerate existing user IDs, leak data from other tables, bypass fraud detection scoring, or cause denial-of-service via expensive injected filters.

## Expected Behavior
User IDs are sanitized, escaped, or passed as separate parameters that cannot affect query structure.

## Actual Behavior
User IDs are string-concatenated into Supabase filter expressions, allowing injection attacks.

## Affected Files
- `backend/api/src/services/fraud/FraudDetectionService.js` (lines 300-327) — `getUserConnections` with unsanitized userId in `.or()`
- `backend/api/src/services/fraud/FraudDetectionService.js` (lines 329-365) — `getBatchUserConnections` with mass concatenation
- `backend/api/src/services/fraud/FraudDetectionService.js` (lines 472-518) — `getRealTimeRisk` calls both methods with external userId data

## Proposed Fix
1. Replace the `.or()` string-based approach with Supabase's filter chaining API or switch to raw SQL with parameterized queries via `supabase.rpc()`
2. Add a `sanitizeUserId()` helper that rejects userIds containing `,`, `)`, `(`, `'`, `"`, or whitespace
3. Move this logic to a Supabase RPC (stored procedure) with proper parameter binding so injection via filter strings is impossible

ionfwsrijan:
### Problem
User IDs were concatenated directly into `.or()` filter strings without sanitization. A crafted userId like `12345),driver_id.eq.(select(*)from` could inject arbitrary filter logic.

### Changes
- Added `sanitizeUserId()` helper that rejects userIds containing `,`, `)`, `(`, `'`, `"`, or whitespace
- Applied sanitization in `getUserConnections` and `getBatchUserConnections`
- Return empty results when sanitized input is invalid

Closes #5528

### DIFF
--- a/backend/api/src/services/fraud/FraudDetectionService.js
+++ b/backend/api/src/services/fraud/FraudDetectionService.js
@@ -297,13 +297,24 @@ class FraudDetectionService {
     }
   }
 
+  sanitizeUserId(userId) {
+    if (typeof userId !== 'string') return '';
+    if (/[,)(\‘\"\s]/.test(userId)) {
+      logger.warn(`[FraudDetection] Rejected userId with invalid characters: ${userId}`);
+      return '';
+    }
+    return userId;
+  }
+
   async getUserConnections(userId) {
     if (!supabase) return [];
+    const safeUserId = this.sanitizeUserId(userId);
+    if (!safeUserId) return [];
     // Get all connections (orders, trips, shared routes)
     const { data: orders, error } = await supabase
       .from('orders')
       .select('customer_id, driver_id')
-      .or(`customer_id.eq.${userId},driver_id.eq.${userId}`);
+      .or(`customer_id.eq.${safeUserId},driver_id.eq.${safeUserId}`);
 
     if (error) {
       logger.error('Failed to load user fraud connections:', error);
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #5592, opened 2026-07-30
fix: replace JWT_SECRET verification with Firebase/Supabase auth in W…

### LINKED ISSUE #5587: bug: WebRTC Signaling Server Rejects All Authenticated Users Due to JWT Verification Key Mismatch
## Description
The REST API authentication middleware (`backend/api/src/middleware/auth.js`) supports two token types: Firebase ID tokens (verified via `firebaseAdmin.auth().verifyIdToken()` at line 157) and Supabase access tokens (verified via `supabase.auth.getUser()` at line 116). Neither token is signed with a symmetric `JWT_SECRET`.

However, `WebRTCSignalingServer.js` line 35 uses `jwt.verify(token, process.env.JWT_SECRET)` — a single shared symmetric secret. **Neither Firebase nor Supabase tokens are signed with `JWT_SECRET`.** Every user authenticated through the normal REST API flow receives a 4001 close code ("Invalid token") when connecting to WebRTC.

The WebRTC signaling server provides: P2P video calling between drivers/customers, peer-to-peer location sharing, mesh networking for convoys, and offline GPS data sync. All these features are entirely non-functional for every user.

## Expected Behavior
The WebRTC signaling server should accept the same Firebase/Supabase tokens that the REST API uses.

## Actual Behavior
Every WebSocket connection to the signaling server is rejected with 4001 because the verification key doesn't match the token issuer.

## Affected Files
- `backend/api/src/services/webrtc/WebRTCSignalingServer.js` (line 35) — Uses `jwt.verify(token, JWT_SECRET)` instead of Firebase/Supabase verification
- `backend/api/src/middleware/auth.js` (lines 104-110, 157) — REST API auth uses `firebaseAdmin.auth().verifyIdToken()` and `supabase.auth.getUser()`
- `backend/api/src/sockets/tracker.js` (line 2) — Location tracking WebSocket (different auth mechanism works, highlighting the inconsistency)

## Proposed Fix
Replace `jwt.verify(token, process.env.JWT_SECRET)` in `WebRTCSignalingServer.js` with the same token verification logic used by the REST API middleware: try Firebase verification first, fall back to Supabase verification. Extract the shared verification logic into a reusable utility so both the HTTP middleware and WebSocket server use the same code path.

ionfwsrijan:
## Description

The WebRTC signaling server verifies tokens using `JWT_SECRET` with
`jwt.verify()`, but the REST API already uses Firebase Auth (with
Supabase fallback). This creates two parallel auth schemes where the
same user may be verified differently depending on the channel.

## Changes

- Extract shared `verifyAuthToken()` function in `auth.js` that
  tries Firebase first, then falls back to Supabase
- Import and use `verifyAuthToken` in `WebRTCSignalingServer.js`
  instead of direct `jwt.verify` with `JWT_SECRET`
- WebRTC now accepts the same tokens as the REST API middleware

Closes #5587

### DIFF
--- a/backend/api/src/middleware/auth.js
+++ b/backend/api/src/middleware/auth.js
@@ -11,6 +11,78 @@ import logger from './logger.js';
  * In production, development auth headers (x-user-id, x-user-role, x-user-name)
  * are unconditionally stripped to prevent any possibility of bypass.
  */
+export async function verifyAuthToken(token) {
+  let userProfile = null;
+  let firebaseUid = null;
+  let supabaseUserId = null;
+
+  let decoded;
+  try {
+    decoded = jwt.decode(token);
+  } catch (err) {
+  }
+
+  const isSupabaseToken = decoded && typeof decoded.iss === 'string' && (decoded.iss.includes('supabase') || decoded.iss.includes('supabase.co'));
+
+  if (isSupabaseToken) {
+    if (!supabase) {
+      throw new Error('Supabase client is not configured on this server.');
+    }
+    const { data: { user }, error: authError } = await supabase.auth.getUser(token);
+    if (authError || !user) {
+      throw new Error(authError?.message || 'Invalid or expired Supabase authentication token.');
+    }
+    supabaseUserId = user.id;
+
+    const { data: profile, error } = await supabase
+      .from('profiles')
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

ionfwsrijan:

PR #5594, opened 2026-07-30
fix: add persistence to offline-first sync service using sqflite

### LINKED ISSUE #5589: bug: Offline-First Sync Service Has Zero Persistence — All Queued Events Lost on App Restart
## Description
The `OfflineFirstSyncService` in `apps/driver/lib/services/offline_first_sync_service.dart` line 5 uses `final List<OfflineSyncEvent> _localDatabase = [];` — a plain in-memory Dart `List`.

There is:
- **No persistence layer** (no SQLite via `sqflite`, no Hive, no SharedPreferences, no file-based write-ahead log)
- **No database initialization** — the entire "database" is recreated as an empty list on every app restart
- **No transactionality** — if the app crashes mid-sync, partial data is lost

**Event ID collisions:** Line 16 generates IDs as `'EVT-${DateTime.now().millisecondsSinceEpoch}'`. If two events are queued within the same millisecond (possible under rapid offline operations), they get identical IDs, causing data integrity issues server-side.

**Sync limitations** (`_processSyncQueue()`, lines 38-55): Processes events sequentially with 800ms delay between each — no batch processing. No retry logic for failed syncs — if one event fails, the loop continues but does not retry. No ordering guarantees.

Truxify is a logistics platform where network connectivity is unreliable (rural India). The offline-first sync is arguably the most critical feature — losing all queued work on app restart is a severe data loss scenario.

## Expected Behavior
Offline events should be persisted to durable storage (SQLite/Hive) so they survive app restart and process crashes.

## Actual Behavior
All offline-queued events are lost when the app is closed, killed, or crashes.

## Affected Files
- `apps/driver/lib/services/offline_first_sync_service.dart` (entire file, 57 lines) — In-memory list with no persistence
- `apps/driver/lib/services/location_service.dart` (lines 166-172) — Offline order caching pattern (same problem)
- `backend/api/src/services/webrtc/WebRTCSignalingServer.js` (lines 386-404) — Server-side offline GPS persistence exists, but client-side has none

## Proposed Fix
1. Replace the in-memory `List` with a persistent store using `sqflite` (SQLite) or `hive` (key-value)
2. Initialize the database from persistent storage on service start
3. Add proper transaction support for batch operations
4. Replace the timestamp-based ID with a UUID or a monotonic counter with millisecond disambiguation
5. Add retry logic with exponential backoff for failed sync events
6. Add batch processing capability to reduce the 800ms sequential delay overhead

ionfwsrijan:
## Description

The `OfflineFirstSyncService` uses an in-memory `List<OfflineSyncEvent>`
which loses all queued events when the app is killed or backgrounded.
Additionally, event IDs are timestamp-based (no uniqueness guarantee),
and there is no retry mechanism for failed syncs.

## Changes

- Replace in-memory list with SQLite database via `sqflite`
- Initialize database from persistent storage on service start
- Generate unique event IDs using timestamp + counter + random
- Add batch processing (10 events per batch)
- Add retry logic with up to 3 attempts for failed sync events
- Add `close()` method for proper resource cleanup

Closes #5589

### DIFF
--- a/apps/driver/lib/services/offline_first_sync_service.dart
+++ b/apps/driver/lib/services/offline_first_sync_service.dart
@@ -1,32 +1,81 @@
 import 'dart:async';
+import 'dart:math';
+import 'package:sqflite/sqflite.dart';
+import 'package:path_provider/path_provider.dart';
+import 'package:path/path.dart' as p;
+import 'dart:convert';
 import '../models/offline_sync_event_model.dart';
 
 class OfflineFirstSyncService {
-  final List<OfflineSyncEvent> _localDatabase = [];
+  Database? _db;
   bool _isConnected = false;
+  int _idCounter = 0;
+  final Random _random = Random();
   final StreamController<bool> _connectionController = StreamController<bool>.broadcast();
   final StreamController<List<OfflineSyncEvent>> _dbController = StreamController<List<OfflineSyncEvent>>.broadcast();
 
   Stream<bool> get connectionStream => _connectionController.stream;
   Stream<List<OfflineSyncEvent>> get databaseStream => _dbController.stream;
 
-  /// Simulates queueing data into local storage (like SQLite/Hive) while offline
-  void queueEvent(String type, Map<String, dynamic> data) {
+  OfflineFirstSyncService() {
+    _initDatabase();
+  }
+
+  Future<void> _initDatabase() async {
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

ionfwsrijan:

PR #5595, opened 2026-07-30
fix: replace unbounded audio cache with bounded LRU cache

### LINKED ISSUE #5590: bug: Voice AI Audio Cache Is an Unbounded In-Memory Map with No Eviction — Guaranteed OOM
## Description
In `backend/api/src/services/voiceService.js` line 6: `export const audioCache = new Map();` — a module-level `Map` that stores generated audio responses (TTS from ElevenLabs). Every call to `processVoiceQuery()` adds a new entry. There is no size limit, no TTL/expiration, no eviction policy (LRU, FIFO), and no cleanup mechanism.

**Memory impact:** In production mode (lines 131-157), each audio response from ElevenLabs TTS can be 100KB–1MB depending on response length. After 1000 production queries: ~100MB–1GB cached. After 10,000 queries: 1GB–10GB — causing OOM kill on any practical server. In mock mode (lines 72-75), a 1000-byte buffer is stored per query, masking the problem until production TTS is enabled.

The `GET /api/voice/audio/:id` route (`voiceRoutes.js` lines 40-47) serves from this cache, so entries cannot be removed without breaking existing responses.

A parallel implementation (`VoiceAiService.js` lines 73-91) takes a different approach — it streams audio directly from ElevenLabs to the client and deletes the temp file in `finally`. This service does not have the caching problem.

## Expected Behavior
Audio cache should have a bounded size (e.g., 100 entries) with LRU eviction and TTL expiration.

## Actual Behavior
The cache grows without bound, eventually exhausting server memory and crashing the process.

## Affected Files
- `backend/api/src/services/voiceService.js` (line 6) — Unbounded `audioCache` Map
- `backend/api/src/services/voiceService.js` (lines 131-157) — Production TTS adds entries without eviction
- `backend/api/src/services/voiceService.js` (lines 72-75) — Mock mode adds smaller entries
- `backend/api/src/routes/voiceRoutes.js` (lines 40-47) — Serves cached audio, preventing entry removal
- `backend/api/src/services/voice/VoiceAiService.js` (lines 73-91) — Parallel implementation that correctly avoids caching

## Proposed Fix
1. Replace the unbounded `Map` with a bounded LRU cache (e.g., `lru-cache` npm package with `max: 100`, `ttl: 600000`)
2. Alternatively, remove the cache entirely and stream audio directly to clients (matching the `VoiceAiService.js` pattern), using temporary files cleaned up in `finally` blocks
3. If caching is required for performance, implement TTL-based cleanup: delete entries older than 10 minutes in a periodic sweep

ionfwsrijan:
## Description

The `audioCache` is a plain `Map` with no size limit or TTL. Each
cached audio buffer stays in memory indefinitely, causing unbounded
memory growth and eventual OOM crashes under sustained voice AI usage.

## Changes

- Add `MAX_CACHE_SIZE` (100 entries) with eviction of oldest entries
  when the limit is exceeded
- Add `CACHE_TTL_MS` (10 minutes) — stale entries beyond this TTL
  are eligible for eviction
- Add `cacheAudio()` helper to manage bounds
- Update `voiceRoutes.js` to read `entry.buffer` from the new
  `{ buffer, timestamp }` cache shape

Closes #5590

### DIFF
--- a/backend/api/src/routes/voiceRoutes.js
+++ b/backend/api/src/routes/voiceRoutes.js
@@ -38,12 +38,12 @@ router.post('/query', authenticate, userLimiter, upload.single('file'), async (r
 });
 
 router.get('/audio/:id', authenticate, userLimiter, (req, res) => {
-  const buffer = audioCache.get(req.params.id);
-  if (!buffer) {
+  const entry = audioCache.get(req.params.id);
+  if (!entry) {
     return res.status(404).json({ error: 'Audio not found' });
   }
   res.set('Content-Type', 'audio/mpeg');
-  res.send(buffer);
+  res.send(entry.buffer);
 });
 
 export default router;
--- a/backend/api/src/services/voiceService.js
+++ b/backend/api/src/services/voiceService.js
@@ -3,7 +3,27 @@ import crypto from 'crypto';
 import { supabase } from '../config/db.js';
 import logger from '../middleware/logger.js';
 
+const MAX_CACHE_SIZE = 100;
+const CACHE_TTL_MS = 10 * 60 * 1000;
 export const audioCache = new Map();
+
+function trimCache() {
+  const now = Date.now();
+  if (audioCache.size > MAX_CACHE_SIZE) {
+    const oldest = [...audioCache.entries()]
+      .filter(([, v]) => now - v.timestamp < CACHE_TTL_MS)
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

ionfwsrijan:

PR #5628, opened 2026-07-31
fix: correct OSRM trip waypoint_index mapping in optimizeWaypoints

### LINKED ISSUE #5623: bug:  OSRM Trip `waypoint_index` Mapping Inverted — Every Multi-Stop Order Stores a Scrambled Waypoint List
## Description
The OSRM Trip API returns `waypoints` **in input order**, with `waypoint_index` = "Index of the point in the trip". The code in `backend/api/src/services/routingService.js` (lines 44–56) iterates the response with `optimizedIndex = i` (treating the array as trip-ordered) and reads `waypoints[osrmWp.waypoint_index - 1]` (treating `waypoint_index` as the input position) — the exact inverse. The inline comment (lines 40–42) even states the correct semantics while the code implements the opposite.

Git history shows the pre-fix code was **correct** (`optimizedWaypoints[newIndex] = waypoints[i - 1]`); commit `580ceb7f` inverted it, and commit `3347ad92` added only a bounds guard.

**Actual behavior:** For any order with ≥2 waypoints, the returned array always has length `N+1` (was allocated `new Array(N)`), a hole at index 0, and a scrambled permutation whenever the optimized order differs from input order. E.g., input `[W1, W2]`, trip `[start, W2, W1, end]` → returns `[<empty>, W2, W1]`. `JSON.stringify` turns the hole into `null`, so `orders.waypoints` and `load_offers.waypoints` (createOrder, `orderLifecycleService.js:123/162`) persist `[null, {...}, {...}]` with wrong stop order.

## Expected Behavior
Waypoints are persisted in the optimized trip order with no holes.

## Actual Behavior
`waypoint_index` semantics are inverted, producing a length-`N+1` array with a `null` hole at index 0 and scrambled stop ordering for multi-stop orders.

## Affected Files
- `backend/api/src/services/routingService.js` (lines 40-56) — Inverted waypoint mapping
- `backend/api/src/services/order/orderLifecycleService.js` (lines 51-58, 123, 162) — Persists scrambled waypoints into `orders`/`load_offers`
- `backend/api/test/unit/routingService.test.js` (lines 111-141) — Test mock contradicts OSRM semantics, masking the bug

## Proposed Fix
Reconstruct from the input-order array: `optimizedWaypoints[waypoint_index - 1] = waypoints[i - 1]` for `i = 1..waypoints.length`, keeping the bounds guard. Rewrite the unit test mock to use OSRM input-order semantics (the current test's mock comment contradicts its own `wp` array).

ionfwsrijan:
## Summary

Fixes the inverted `waypoint_index` mapping in `optimizeWaypoints` that caused every multi-stop order to persist a scrambled waypoint list (length `N+1`, a `null` hole at index 0 via `JSON.stringify`).

## Root cause

OSRM Trip API returns `waypoints` **in input order**; each entry's `waypoint_index` is its position in the optimized trip. The code iterated with `optimizedIndex = i` (treating the array as trip-ordered) and read `waypoints[waypoint_index - 1]` (treating `waypoint_index` as the input position) — the exact inverse.

## Changes

- `backend/api/src/services/routingService.js` — reconstruct from the input-order array: `optimizedWaypoints[waypoint_index - 1] = waypoints[i - 1]` for `i = 1..waypoints.length`, keeping the bounds guard. Result now has no holes and reflects the optimized stop order.
- `backend/api/test/unit/routingService.test.js`:
  - Rewrote the "reorders" mock to OSRM input-order semantics (the old mock contradicted itself and masked the bug).
  - `mockAxiosGet.mockReset()` in `beforeEach` so queued once-values from earlier tests cannot leak (previously the empty-waypoints test used a single waypoint, never called axios, and left a stale mock that made the reorder test pass by fallback).
  - Empty-waypoints fallback test now uses two waypoints so it genuinely exercises the axios path.

## Verification

The rewritten test fails against the old mapping (reproducing `[undefined, ...]`) and passes with the fix; full `test/unit` suite shows zero new failures vs. the pre-existing baseline.

Closes #5623

### DIFF
--- a/backend/api/src/services/routingService.js
+++ b/backend/api/src/services/routingService.js
@@ -42,14 +42,15 @@ export async function optimizeWaypoints(start, end, waypoints) {
     // Index 0 is the start, Index N is the end.
     
     const optimizedWaypoints = new Array(waypoints.length);
-    
-    // Original array order: [Start, WP1, WP2, ..., End]
-    for (let i = 0; i < waypointsResult.length; i++) {
+
+    // waypointsResult is in input order: [Start, WP1, WP2, ..., End].
+    // Each waypoint's `waypoint_index` is its position in the optimized trip
+    // (0 = start, waypoints.length + 1 = end), so subtract 1 for the middle stops.
+    for (let i = 1; i <= waypoints.length; i++) {
       const osrmWp = waypointsResult[i];
-      const originalIndex = osrmWp.waypoint_index - 1;
-      const optimizedIndex = i;
-      if (originalIndex >= 0 && originalIndex < waypoints.length) {
-        optimizedWaypoints[optimizedIndex] = waypoints[originalIndex];
+      const optimizedIndex = osrmWp.waypoint_index - 1;
+      if (optimizedIndex >= 0 && optimizedIndex < waypoints.length) {
+        optimizedWaypoints[optimizedIndex] = waypoints[i - 1];
       }
     }
 
--- a/backend/api/test/unit/routingService.test.js
+++ b/backend/api/test/unit/routingService.test.js
@@ -45,6 +45,7 @@ describe('routingService - getHaversineDistance', () => {
 describe('routingService - optimizeWaypoints', () => {
   beforeEach(() => {
     vi.clearAllMocks();
+    mockAxiosGet.mockReset();
   });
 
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #5685, opened 2026-08-01
fix: filter truck search using truck_type instead of model name

### LINKED ISSUE #5684: bug: truck_type filter substring-matches model names — body-type filter always empties results
### Summary
The `truck_type` filter in the truck search endpoint compares the filter value against the truck *model name* using a substring match, but the `trucks` table has no body-type column. Selecting "Open Body", "Closed Body", "Container", or "Refrigerated" in the app therefore never matches any real truck and silently empties the results.

### Affected file
- `backend/api/src/routes/truckRoutes.js`

### Evidence
- Filter logic (`truckRoutes.js:588-594`):
  ```js
  if (truck_type && truck_type !== '') {
      const truckNameLower = (truck.truck || '').toLowerCase();   // e.g. 'tata 407'
      const typeLower = truck_type.toLowerCase();                 // e.g. 'open body'
      if (!truckNameLower.includes(typeLower)) {
          return false;
      }
  }
  ```
- Schema (`docs/supabase_setup.sql:139-154`): `trucks` has `name` (model, e.g. `'Tata 407'`), `number_plate`, capacity/dimensions, compliance dates — there is no body-type/`truck_type` column.
- The customer app sends `truck_type` from the body-type dropdown (`apps/customer/lib/services/order_service.dart`, `find_trucks_screen.dart`).
- The value is validated server-side (`truckRoutes.js`) but the filter compares against model names, so it can only match if a model literally contains "open body"/"container" — which no real model does.

### Steps to reproduce
1. Open the customer app truck search.
2. Pick any body type filter (e.g. "Container").
3. Submit the search with matching trucks present.

### Actual behavior
`'tata 407'.includes('container')` → `false`, so all trucks are filtered out and the results list is empty even though suitable trucks exist.

### Expected behavior
Trucks should be filtered by an actual body-type attribute (a real `truck_type`/body-type column), not by substring-matching model names.

### Impact
The body-type filter always produces empty results, making a primary search facet non-functional.

CoderMS07:
## Summary

This PR fixes truck search filtering by introducing and using a dedicated `truck_type` field instead of comparing the selected body type against the truck model name.

### Changes
- Added `truck_type` column to the `trucks` table with valid constraints.
- Added request validation for `truck_type`.
- Stored `truck_type` when registering trucks.
- Updated truck search to filter using `truck_type`.
- Removed reliance on truck model name for body-type filtering.

Closes https://github.com/KanishJebaMathewM/Truxify/issues/5684

## Summary by CodeRabbit

- **New Features**
  - Truck registration now captures and displays the truck type.
  - Supported types include Open Body, Closed Body, Container, and Refrigerated.
  - Truck search can filter results by exact truck type.

- **Bug Fixes**
  - Search results no longer match truck types against truck names.
  - Internal capacity and type details are excluded from search responses where appropriate.

### DIFF
--- a/backend/api/src/routes/truckRoutes.js
+++ b/backend/api/src/routes/truckRoutes.js
@@ -188,7 +188,7 @@ function parseCapacityFilter(value, field) {
  *         description: Number plate already registered
  */
 router.post('/', authenticate, requirePolicy('truck:register'), userLimiter, validateBody(registerTruckSchema), async (req, res) => {
-  const { name, number_plate, max_capacity_tons } = req.body;
+  const { name, truck_type, number_plate, max_capacity_tons } = req.body;
   const normalizedNumberPlate = sanitizeNumberPlate(number_plate);
 
   try {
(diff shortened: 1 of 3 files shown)


CoderMS07:
GSSOC

PR #5719, opened 2026-08-02
fix(delivery): persist escrow release outcome before complete_trip_tx to avoid deadlock

### LINKED ISSUE #5704: bug: Verify-Delivery Permanently Deadlocks When On-Chain Release Succeeds but `complete_trip_tx` Fails; No Reconciliation Path Exists
## Description
In `backend/api/src/services/order/deliveryVerificationService.js` `verifyDelivery` (lines 242-359):
1. The on-chain release hash is only written to the DB **after** the `complete_trip_tx` RPC succeeds (lines 329-335: `updateOrder({ escrow_status: 'released', release_tx_hash ... })`). If the RPC fails (lines 303-306, transient Supabase/network error), the hash is lost and `escrow_status` stays `funded`.
2. On retry, `escrowRelease` sees `booking.paid === true` and returns `{ txHash: null, alreadyReleased: true }` (`escrow.js` lines 361-365), so the app calls `complete_trip_tx` with `p_release_tx_hash: null`.
3. `complete_trip_tx` (`20260706075009_secure_rpc_search_path.sql` line 38) raises `'Blockchain escrow release must complete before crediting driver wallet'` whenever `escrow_status IN ('funded','release_failed')` and the hash is NULL → the request 500s.
4. The retry-for-stuck-escrow branch requires `order.status === 'payment_released'` (lines 70, 246), but status is still `arriving`, so it never runs.
5. The background reconciler `escrowReleaseReconciliation.js` only selects orders with `escrow_status='release_failed'` and `escrow_release_attempts < 10` — this order is `funded`, so it is never picked.

## Expected Behavior
A transient DB/network failure after a successful on-chain release is recoverable.

## Actual Behavior
The order is permanently stuck in `arriving`/`funded` with the escrow already released on-chain (driver can withdraw from the contract) but the DB never credits the driver wallet, never records the trip, and never completes the order. Every retry hits the same SQL gate. Only manual ops intervention can resolve it.

## Affected Files
- `backend/api/src/services/order/deliveryVerificationService.js` (lines 70, 242-359) — Release hash written only after RPC; stuck-status branch unreachable
- `backend/api/src/services/escrow.js` (lines 361-365) — `alreadyReleased: true` returns null hash
- `supabase/migrations/20260706075009_secure_rpc_search_path.sql` (line 38) — SQL gate requires non-null hash
- `backend/api/src/services/escrow/escrowReleaseReconciliation.js` — Only reconciles `release_failed` orders

## Proposed Fix
1. Persist the release outcome immediately after on-chain confirmation (set `escrow_status='released'` + `release_tx_hash` before invoking the RPC)
2. Or have `escrowRelease` return a deterministic sentinel for `alreadyReleased` that the SQL accepts
3. Or extend `escrowReleaseReconciliation` to also reconcile `escrow_status='funded'` orders whose on-chain booking shows `paid=true`

ionfwsrijan:
### Problem
`verifyDelivery` released the escrow on-chain and then called `complete_trip_tx`. If that RPC failed transiently, `escrow_status` remained `funded`/`release_failed` with no `release_tx_hash` persisted — and the SQL gate (`RAISE EXCEPTION ... p_release_tx_hash IS NULL`) deadlocked every retry, blocking the order from ever completing.

### Changes
- `backend/api/src/services/order/deliveryVerificationService.js` `verifyDelivery`
  - Added `release_tx_hash` to the order select.
  - After a successful on-chain release (or when already released), persists `escrow_status: 'released'`, `release_tx_hash`, `escrow_release_error: null`, and `escrow_released_at` **before** calling `complete_trip_tx`.
  - The `escrow_status === 'released'` branch reuses the persisted hash instead of discarding it.
- Verified: `escrow.test.js` + `deliveryVerificationGeofence.test.js` pass 43/43.

### Impact
A transient RPC failure after release is now retryable — the released state and hash are already persisted, so the order can finish without deadlocking against the SQL gate.

Closes #5704

### DIFF
--- a/backend/api/src/services/order/deliveryVerificationService.js
+++ b/backend/api/src/services/order/deliveryVerificationService.js
@@ -57,7 +57,7 @@ export class DeliveryVerificationService {
       });
     }
 
-    const { data: order, error: orderErr } = await this.orderRepository.findOrderById(orderId, 'id, order_display_id, driver_id, customer_id, escrow_status, escrow_release_attempts, status, drop_lat, drop_lng, toll_estimate, base_freight, platform_fee, total_amount');
+    const { data: order, error: orderErr } = await this.orderRepository.findOrderById(orderId, 'id, order_display_id, driver_id, customer_id, escrow_status, escrow_release_attempts, status, release_tx_hash, drop_lat, drop_lng, toll_estimate, base_freight, platform_fee, total_amount');
 
     if (orderErr || !order) {
       throw new DomainError(404, { error: 'Order not found.' });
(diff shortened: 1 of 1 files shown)


ionfwsrijan:
@KanishJebaMathewM Please add proper gssoc:approved and difficulty label

PR #5749, opened 2026-08-02
fix(api): pass per-request user client to SECURITY DEFINER RPCs 

### LINKED ISSUE #5737: bug: SECURITY DEFINER RPCs Bypass Their Own `auth.uid()` Guards Because Service Code Calls Them Through the Shared Anon-Key Client
## Description
The security migration `supabase/migrations/20260706075009_secure_rpc_search_path.sql` embeds `auth.uid()` authorization checks inside every SECURITY DEFINER function — `complete_trip_tx` (line 25), `accept_bid_tx` (line 174), `withdraw_funds_tx` (line 230) — and `backend/api/test/unit/rlsSecurity.test.js` (lines 166-209) asserts those checks exist. However, nearly every call site invokes these RPCs through `OrderRepository.executeRpc()` (`orderRepository.js:359-362`), which defaults to the shared Supabase client created in `config/db.js:33` with only the **anon key** and no per-user JWT.

Without a user JWT, PostgREST runs the function as the `anon` role and `auth.uid()` returns `NULL`. The guard `IF auth.uid() <> v_order.driver_id THEN RAISE` evaluates `NULL <> driver_id` to `NULL` (not `TRUE`), so the `RAISE` never fires — the embedded checks are dead code whenever the shared client is used.

## Expected Behavior
The SQL-level guards added by the migration fire to stop unauthorized RPC execution — the "last line of defense" the RLS tests assert.

## Actual Behavior
Callers omitting a user client silently nullify the SQL checks:
- `backend/api/src/services/order/orderMilestoneService.js:191` — `complete_trip_tx` (driver auth + wallet credit)
- `backend/api/src/services/order/deliveryVerificationService.js:296` — same RPC
- `backend/api/src/services/order/bidAcceptanceService.js:134` — `accept_bid_tx` (customer auth)
- `backend/api/src/services/order/orderLifecycleService.js:579` — `update_order_and_load_offer`
- `backend/api/src/services/order/orderLifecycleService.js:882` — `submit_rating_tx`

The `submit_rating` **route** correctly passes a user client (`orderRoutes.js:708`), but the service-layer paths above never do. An unauthorized invocation of `complete_trip_tx` from a future worker, reconciliation job, or admin tool would credit `wallet_confirmed` and complete any order without the embedded driver check; only coincidental service-layer checks (`orderMilestoneService.js:49`, `deliveryVerificationService.js:66`, `bidAcceptanceService.js:25`) currently mask the gap.

## Affected Files
- `supabase/migrations/20260706075009_secure_rpc_search_path.sql` (lines 25, 174, 230) — embedded `auth.uid()` checks that evaluate `NULL`
- `backend/api/src/repositories/orderRepository.js` (lines 359-362) — `executeRpc` falls back to shared anon client
- `backend/api/src/config/db.js` (line 33) — shared anon-key client
- `backend/api/src/services/order/orderMilestoneService.js` (line 191) — `complete_trip_tx` via shared client
- `backend/api/src/services/order/bidAcceptanceService.js` (line 134) — `accept_bid_tx` via shared client
- `backend/api/test/unit/rlsSecurity.test.js` (lines 166-209) — tests assert checks that never fire in service calls

## Proposed Fix
1. Pass a per-request user client to every `executeRpc` call (as `orderRoutes.js:708` already does) so `auth.uid()` resolves to the caller
2. Make `executeRpc` require the `client` parameter (fail instead of defaulting to the anon client)
3. Add a regression test asserting `auth.uid()` is non-NULL inside the RPC path during service-level invocation

ionfwsrijan:
## bug: SECURITY DEFINER RPCs Bypass Their Own `auth.uid()` Guards

Closes #5737

### Summary
Every service-layer call site invoked SECURITY DEFINER RPCs (`complete_trip_tx`,
`accept_bid_tx`, `submit_rating_tx`, etc.) through `OrderRepository.executeRpc()`,
which silently defaulted to the shared anon-key Supabase client. Without a
per-user JWT, PostgREST runs the function as the `anon` role and `auth.uid()`
returns `NULL` — so the embedded `IF auth.uid() <> ... THEN RAISE` guards
evaluated `NULL` and never fired, making the SQL-level checks dead code.

### Changes
- `orderRepository.executeRpc` now **requires** an explicit client and throws
  when one is omitted — no more silent anon-key fallback
- `verifyDeliveryFn` / `changeDrop` / `confirmDeposit` / `submitRating` accept a
  `userClient` and pass it through to the RPC, falling back to `supabaseAdmin`
  only where there is genuinely no user session
- `deliveryVerificationService` and `orderMilestoneService` forward `userClient`
  to `complete_trip_tx`
- `verify-delivery`, `change-drop`, `submit-rating`, and `confirm-deposit` routes
  build a per-request user client from `req.token`
- Driver wallet withdraw builds a per-request user client instead of using the
  shared anon-key client
- Escrow funding reconciliation passes `supabaseAdmin` explicitly
- Added regression tests asserting every `executeRpc` call site passes an
  authenticated client; updated `orderLifecycleService` tests to assert
  `userClient` forwarding

### Testing
- `rlsSecurity` regression tests: all `executeRpc` call sites assert an
  authenticated client is passed
- `orderLifecycleService` unit tests updated and passing

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction security for wallet, delivery, escrow, deposit, and rating operations by using appropriate authenticated or administrative access.
  * Prevented database operations from silently using an unintended shared connection.
  * Increased the escrow lock timeout to improve processing reliability.

* **Tests**
  * Added coverage verifying secure access handling across order and wallet workflows.

### DIFF
--- a/backend/api/src/repositories/orderRepository.js
+++ b/backend/api/src/repositories/orderRepository.js
@@ -357,8 +357,12 @@ export class OrderRepository {
   // ===================================================================
 
   async executeRpc(name, params, client) {
-    const supabaseClient = client || this.supabase;
-    return this._retryableQuery(() => supabaseClient.rpc(name, params), `executeRpc:${name}`);
+    if (!client) {
+      throw new Error(
+        `executeRpc("${name}") requires a Supabase client. Pass the per-request user client so auth.uid() resolves to the caller instead of falling back to the shared anon-key client.`
+      );
+    }
+    return this._retryableQuery(() => client.rpc(name, params), `executeRpc:${name}`);
   }
 
   // ===================================================================
--- a/backend/api/src/routes/driverRoutes.js
+++ b/backend/api/src/routes/driverRoutes.js
@@ -947,7 +947,7 @@ router.post('/wallet/withdraw', authenticate, userLimiter, requirePolicy('driver
     }
 
     // 5.2 Execute atomically via Supabase RPC
-    const userClient = req.token ? createUserClient(req.token) : supabase;
+    const userClient = createUserClient(req.token);
     const { error: rpcErr } = await userClient.rpc('withdraw_funds_tx', {
       p_driver_id: req.user.id,
       p_amount:    amount
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -145,7 +145,7 @@ import crypto from 'crypto';
 import rateLimit from 'express-rate-limit';
 
 import { bidLimiter, userLimiter, userKeyGenerator, createStore } from '../middleware/rateLimiter.js';
-import { mongoDb, supabase, supabaseAdmin, redisClient, createUserClient } from '../config/db.js';
+import { mongoDb, supabase, redisClient, createUserClient } from '../config/db.js';
 import { authenticate, requireRole } from '../middleware/auth.js';
 import { requirePolicy } from '../middleware/requirePolicy.js';
 import { validateDocumentBuffer } from '../lib/documentValidation.js';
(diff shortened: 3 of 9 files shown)


ionfwsrijan:

PR #5753, opened 2026-08-02
fix(worker): close staleOrderWorker TOCTOU race and refund funded escrow

### LINKED ISSUE #5741: bug: staleOrderWorker Races Bid Acceptance and Cancels Freshly Accepted Orders, Stranding Funded Escrow (TOCTOU)
## Description
`backend/api/src/workers/staleOrderWorker.js` (hourly cron) selects all `status='pending'` orders older than 24h (lines 19-26) and then, for each, blindly runs `supabase.from('orders').update({status:'cancelled', ...}).eq('id', order.id)` (lines 45-49) — with **no `.eq('status','pending')` guard** — plus an unconditional `load_offers` update to `cancelled` (lines 60-63), and it never inspects `escrow_status`.

This creates a TOCTOU race against the bid-acceptance flow: `bidAcceptanceService.acceptBid` first persists the escrow booking reference (`updateEscrowBooking`, line 118, order still `pending`, `escrow_status='funding'`) and only then runs `accept_bid_tx` (line 134), which flips the order to `truck_assigned`. If the worker's SELECT sees `pending` and its UPDATE lands after `accept_bid_tx` committed `truck_assigned` (or between the escrow persistence and the RPC), the worker overwrites an accepted order to `cancelled`. The customer then submits the returned `depositTx`, funding an on-chain booking whose order is now `cancelled` — and the worker never calls the escrow refund path (`escrow.js:submitEscrowRefund`/`cancelOrder`), so the customer's MATIC is locked in `TruxifyEscrow` as `Active` with no refund triggered.

## Expected Behavior
The worker cancels only orders that are still `pending` at update time and routes any escrowed funds through the same refund pipeline as `orderLifecycleService.cancelOrder` (lines 641-771).

## Actual Behavior
A last-writer-wins update can cancel an order that was just accepted (driver assigned, bid burned, rivals rejected), and a funded-but-cancelled order leaves customer funds stranded on-chain with no automatic refund.

## Affected Files
- `backend/api/src/workers/staleOrderWorker.js` (lines 19-26, 45-49, 60-63) — unconditional status overwrite, no escrow handling
- `backend/api/src/services/order/bidAcceptanceService.js` (lines 101-146) — escrow persistence then `accept_bid_tx`, the window the worker races
- `backend/api/src/services/order/orderLifecycleService.js` (lines 641-771) — the refund pipeline the worker does not reuse
- `backend/api/src/services/escrow.js` — `submitEscrowRefund`/`cancelOrder` never invoked by the worker

## Proposed Fix
1. Add `.eq('status','pending')` (and `.eq('escrow_status','pending')` in the retry) to the worker's update so it can never overwrite a concurrently accepted order
2. Re-fetch the order inside the loop with a status filter to close the window
3. Before cancelling, check `escrow_status` and call `submitEscrowRefund`/`confirmEscrowRefund` when funded

ionfwsrijan:
## bug: Stale Order Worker TOCTOU Race and Stranded Escrow Funds

Closes #5741

### Summary
`staleOrderWorker` could cancel an order that was concurrently being accepted:
the batch `SELECT` returned a `pending` order, the customer accepted it
(flipping `escrow_status` to `funding` then `status` to `truck_assigned`), and
the worker's unguarded `UPDATE` still cancelled it. Funded orders also had no
refund path — the worker marked them cancelled without returning on-chain funds.

### Changes
- **Re-fetch inside the loop**: each stale order is re-read with
  `.eq('status','pending').maybeSingle()` before any mutation, closing the window
  between the batch SELECT and the per-order UPDATE
- **Guarded cancellations**: every update is now atomic on
  `status='pending'` **and** the current `escrow_status` —
  - plain cancel: `.or('escrow_status.is.null,escrow_status.eq.pending')`
  - refunded cancel: `.eq('escrow_status', <observed>)`
  A concurrently accepted order (escrow_status `funding` is persisted before
  `status` flips) can never be overwritten
- **`funding` orders are skipped** and left to the escrow-funding reconciliation
  worker (TTL sweep) which owns that transition
- **Escrow refunds wired in**: funded / `refund_pending` / `refund_failed` orders
  route through `submitEscrowRefund` + `confirmEscrowRefund` (mirroring
  `orderLifecycleService.cancelOrder`), landing in `refunded`, or
  `refund_pending`/`refund_failed` with `escrow_refund_error` for the refund
  reconciliation worker to retry — no more stranded customer funds
- Side effects (load-offer cancel, push notification) now fire only when the
  guarded UPDATE actually affected a row

### Testing
- **New `staleOrderWorkerRace.test.js`** — 5 tests: skip when re-fetch not
  pending, skip `funding` in flight, guarded plain-cancel filter, funded path
  calls `submitEscrowRefund` then finalizes `refunded`, refund failure →
  `refund_failed` with error recorded
- Updated `staleOrderWorkerNotifications.test.js` mock to the guarded-update
  contract
- All worker suites pass: 3 files, 8 tests

### DIFF
--- a/backend/api/src/workers/staleOrderWorker.js
+++ b/backend/api/src/workers/staleOrderWorker.js
@@ -2,10 +2,13 @@ import cron from 'node-cron';
 import logger from '../middleware/logger.js';
 import { supabase } from '../config/db.js';
 import { sendPushNotification } from '../services/notificationService.js';
+import { submitEscrowRefund, confirmEscrowRefund } from '../services/escrow.js';
 import { WorkerTracer } from '../core/telemetry/WorkerTracer.js';
 import spanFactory from '../core/telemetry/SpanFactory.js';
 
 let staleOrderWorkerTask = null;
+
+const STALE_ORDER_CANCELLATION_REASON = 'Stale order: no accepted bid within 24 hours.';
 
 export const startStaleOrderWorker = () => {
   if (staleOrderWorkerTask) {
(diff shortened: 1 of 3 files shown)


ionfwsrijan:

PR #5770, opened 2026-08-02
fix(api): enforce trip ownership on event batch insert and read access

### LINKED ISSUE #5755: bug: Trip Batch Event Endpoint Has No `trip_id` Ownership Check: Attacker Can Inject Events Into Any Trip and Then Read Its Full History
## Description
`POST /api/v1/trips/events/batch` (`backend/api/src/routes/tripRoutes.js:219`) inserts events with a client-supplied `trip_id` — a free-form string (`tripEventSchema` line 147) — with no verification that the trip belongs to the caller. The insert path at lines 265-293 writes the attacker's `trip_id` straight to the `trip_events` table. Worse, the GET endpoint's access check (lines 420-445) treats the presence of any matching event as proof of access: `const isDriver = order?.driver_id === req.user.id || existingEvent?.user_id === req.user.id` (line 440). So an attacker who injects one junk event into a victim's trip then gains read access to that trip's full event history (select at lines 449-450), including geolocation and metadata.

## Expected Behavior
Events are only accepted for trips the caller owns (or is assigned to), and read access derives from trip/order ownership, never from a user-controlled `trip_events` row.

## Actual Behavior
Any authenticated user can inject events into arbitrary `trip_id`s (poisoning tracking/analytics) and, by injecting a single row, unlock read access to a victim's complete trip event history.

## Affected Files
- `backend/api/src/routes/tripRoutes.js` (line 219, 265-293) — batch insert without ownership check
- `backend/api/src/routes/tripRoutes.js` (lines 420-445) — read access granted by attacker-created event row
- `backend/api/src/validation/requestSchemas.js` (line 147) — `trip_id` optional free-form string

## Proposed Fix
1. Before insert, verify the `trip_id` resolves to a trip owned/assigned to `req.user.id` (or drop trip linkage from client input entirely)
2. Remove the `existingEvent?.user_id === req.user.id` branch from the GET access check; authorize purely by trip/order ownership

ionfwsrijan:
## What
Trip event endpoints now enforce ownership so callers can only read/write events for trips linked to their own orders.

## Why
Fixes #5755 — event ownership was missing, allowing cross-order access to trip telemetry.

## Changes
- `tripRoutes.js`: batch-event POST verifies every `trip_id` against `orders` (admin bypass).
- GET `/trips/:id/events` requires a matching order.
- Tests added; ESLint clean.

Closes #5755

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened trip event access checks for batch submissions.
  * Prevented unauthorized users from adding events to trips they do not own.
  * Trip event retrieval now returns a not-found response when the related order is missing.
  * Improved error handling when trip ownership verification fails.
  * Added coverage for mismatched ownership, missing orders, and driver access scenarios.

### DIFF
--- a/backend/api/src/routes/tripRoutes.js
+++ b/backend/api/src/routes/tripRoutes.js
@@ -253,6 +253,36 @@ router.post('/events/batch', authenticate, userLimiter, validateBatchPayload(bat
       }
     }
 
+    // 2b. Ownership check: events may only be attached to trips (orders) the
+    // caller owns or is assigned to. Never trust a client-supplied trip_id.
+    if (req.user.role !== 'admin') {
+      const tripIds = [...new Set(events.map(event => event.trip_id).filter(Boolean))];
+
+      if (tripIds.length > 0) {
+        const { data: ownedOrders, error: ownershipError } = await supabase
+          .from('orders')
+          .select('id, driver_id, customer_id')
+          .in('id', tripIds);
+
+        if (ownershipError) {
+          logger.error('[SyncEngine] Failed to verify trip ownership:', ownershipError.message);
+          return res.status(500).json({ error: 'Internal Server Error' });
+        }
+
+        const orderById = new Map((ownedOrders || []).map(order => [order.id, order]));
+
+        for (const tripId of tripIds) {
+          const order = orderById.get(tripId);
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

PR #5771, opened 2026-08-02
fix(db): finalize only the order-linked trip in complete_trip_tx 

### LINKED ISSUE #5756: bug: `complete_trip_tx` Finalizes the Driver's *Only Active Trip* Regardless of Which Order Is Being Completed — Wrong Trip Marked Delivered
## Description
`complete_trip_tx` (`supabase/migrations/20260704000001_add_auth_verification_to_complete_trip_tx.sql:63-94`) finalizes the driver's **only** active trip, queried purely by `driver_id + status='active'` (lines 63-66), with no link to the order being completed — the `trips` table has no `order_id` column (confirmed in `docs/supabase_setup.sql:502-530`). The RPC:
- Raises only when `v_active_trip_count > 1` (line 68), otherwise if exactly one active trip exists it marks that trip, its `trip_items`, and its `trip_stops` as delivered (lines 72-93) — even if it is an unrelated load
- If `v_active_trip_count = 0`, the order is still marked `payment_released` (line 99) and the wallet credited (lines 109-136) with no trip record at all

## Expected Behavior
Completion updates the specific trip that served this order; a missing trip for a completed delivery should be an error, not a silent wallet credit.

## Actual Behavior
Completing order A while an unrelated active trip exists marks that unrelated trip, all its items, and its stops as delivered — corrupting the other load's tracking. With zero active trips, the order is silently marked `payment_released` and the driver credited with no trip record, breaking auditability.

## Affected Files
- `supabase/migrations/20260704000001_add_auth_verification_to_complete_trip_tx.sql` (lines 63-99, 109-136) — trip matched by driver_id/status only
- `docs/supabase_setup.sql` (lines 502-530) — `trips` schema has no `order_id`
- `backend/api/src/routes/tripRoutes.js` — trip creation path that never links trips to orders

## Proposed Fix
1. Add an `order_id` FK to `trips` and select the active trip by `order_id = p_order_id`
2. Require a matching trip before crediting: raise if no trip exists for the order rather than silently paying out

ionfwsrijan:
## What
Trips are now explicitly linked to orders, and completing a trip credits the order's wallet.

## Why
Fixes #5756 — completing a trip did not reliably resolve the owning order or credit the right amount.

## Changes
- Migration `supabase/migrations/20260802060000_link_trips_to_orders.sql`: adds `trips.order_id` FK + index.
- `complete_trip_tx` selects the active trip `WHERE order_id = p_order_id` and raises `'No active trip found for this order — cannot complete trip'`.
- Wallet credit uses `COALESCE(v_order.bid_amount, v_order.total_amount)`.

## Tests
`verifyDbSchema` (13) and `rlsSecurity` (120) updated and passing.

Closes #5756

### DIFF
--- a/backend/api/test/unit/rlsSecurity.test.js
+++ b/backend/api/test/unit/rlsSecurity.test.js
@@ -213,6 +213,27 @@ describe('accept_bid_tx — auth.uid() verification present in migration chain',
   });
 });
 
+describe('complete_trip_tx — order-linked trip finalization (issue #5756)', () => {
+  it('the 20260704000001 migration selects the trip by order_id and raises when none exists', async () => {
+    const p = path.resolve(__dirname, '../../../../supabase/migrations/20260704000001_add_auth_verification_to_complete_trip_tx.sql');
+    const content = await fs.readFile(p, 'utf8');
+
+    expect(/select\s+trip_display_id\s+into\s+v_trip_display_id\s+from\s+trips\s+where\s+order_id\s*=\s*p_order_id/i.test(content)).toBe(true);
+    expect(/if\s+v_trip_display_id\s+is\s+null[\s\S]*raise\s+exception[\s\S]*no\s+active\s+trip\s+found/i.test(content)).toBe(true);
+    expect(/v_active_trip_count/i.test(content)).toBe(false);
+  });
+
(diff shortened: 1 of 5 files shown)


ionfwsrijan:

PR #5804, opened 2026-08-02
fix: accept_bid_tx resolves bid/load/order/driver/amount from the stored bid row instead of trusting caller-supplied values

### LINKED ISSUE #5777: bug: `accept_bid_tx` Trusts Caller-Supplied Bid/Driver/Truck/Amount Fields — No Binding to the Stored Bid
## Description
The `accept_bid_tx` SQL function (`supabase/migrations/20260628000000_add_rpc_functions.sql:8-80`) is `SECURITY DEFINER` and accepts `p_bid_id`, `p_driver_id`, `p_truck_id`, `p_driver_name`, `p_driver_rating`, `p_truck_number`, `p_bid_amount`, and `p_order_display_id` all from the RPC caller. It validates only:
- the **load offer** is `available` (lines 30-37), and
- the **order** is `pending` (lines 39-46).

It never verifies that `p_bid_id` exists in `load_bids`, belongs to this `p_load_id`, or belongs to `p_driver_id`; it never verifies `p_bid_amount` matches the bid's stored `bid_amount`. It then writes all caller-supplied values onto the order (lines 61-72):
```sql
update orders
  set driver_id = p_driver_id, truck_id = p_truck_id,
      total_amount = p_bid_amount, bid_amount = p_bid_amount, ...
```
Any authenticated caller able to invoke this RPC (or a compromised/misbehaving client) can accept an arbitrary or nonexistent bid ID, assign any driver/truck to the order, and set the order amount to any value — including inflating `total_amount`/`bid_amount` so a later `complete_trip_tx` credits an inflated wallet amount.

## Expected Behavior
`accept_bid_tx` must load the bid by `p_bid_id`, verify it exists, belongs to the given load and driver, is still `pending`, and use the stored `bid_amount` (ignoring any caller-supplied amount).

## Actual Behavior
All bid/driver/truck/amount fields are taken verbatim from the caller with no cross-checks against `load_bids`.

## Affected Files
- `supabase/migrations/20260628000000_add_rpc_functions.sql` (lines 8-80) — `accept_bid_tx` definition
- `supabase/migrations/20260628000000_add_rpc_functions.sql` (lines 48-72) — writes caller-supplied values
- `backend/api/src/services/order/bidAcceptanceService.js` (line 134) — RPC invocation passing client-derived args

## Proposed Fix
Rewrite `accept_bid_tx` to resolve the bid from `p_bid_id` (joining `load_bids` to `load_offers` and `orders`), enforce `load_id`/`driver_id`/`status='pending'`/`bid_amount` from the stored row, and use only the stored values for the order update; reject if no matching pending bid exists.

ionfwsrijan:
## Summary
`accept_bid_tx` trusted caller-supplied values for bid id, driver, truck, and amount. A caller could accept a bid for a different driver/truck or credit an arbitrary amount, with nothing binding the write to the stored bid row.

## Changes
- **Trust stored bid row** (`supabase/migrations/20260802120000_accept_bid_tx_trust_stored_bid.sql`): the RPC now resolves the bid/load/order/driver/amount from the actual `bids` row identified by the stored bid id, instead of accepting those fields as parameters.
- The transaction validates the resolved values and fails closed if the bid does not exist or does not match the expected state.

Closes #5777

### DIFF
--- a/supabase/migrations/20260802120000_accept_bid_tx_trust_stored_bid.sql
+++ b/supabase/migrations/20260802120000_accept_bid_tx_trust_stored_bid.sql
@@ -1,0 +1,150 @@
+-- =============================================================================
+-- Migration: accept_bid_tx must trust only the stored bid row (Issue #5777)
+-- =============================================================================
+-- Problem:
+--   accept_bid_tx is SECURITY DEFINER and accepted p_bid_id, p_load_id,
+--   p_driver_id, p_driver_name, p_driver_rating, p_truck_id, p_truck_number,
+--   p_bid_amount and p_order_display_id verbatim from the caller. It validated
+--   only that the load offer was 'available' and the order 'pending', then wrote
+--   all caller-supplied values onto the order. Any owning customer (or a
+--   compromised backend client) could accept an arbitrary/nonexistent bid id,
+--   assign any driver/truck, and set the order amount to any value — including
+--   inflating total_amount/bid_amount so a later complete_trip_tx credits an
+--   inflated wallet amount.
+--
+-- Fix:
+--   Resolve the bid from p_bid_id and derive the load_offer + order chain from
+--   it. All caller-supplied bid/driver/truck/amount/order fields are ignored;
+--   the driver identity, display snapshot (name/rating/truck), and bid amount
+--   are taken only from the stored load_bids row and the related tables. The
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

KanishJebaMathewM (approved):

PR #5805, opened 2026-08-02
fix: complete_trip_tx fails closed unless escrow was funded+released or the order is explicitly escrow-disabled

### LINKED ISSUE #5778: bug: `complete_trip_tx` Credits the Driver Wallet Even When the Escrow Was Never Funded
## Description
`complete_trip_tx` (`supabase/migrations/20260624223000_make_delivery_otp_completion_atomic.sql:5-126`) guards wallet credit with a condition that only covers **two** escrow states:
```sql
-- line 35
IF v_order.escrow_status IN ('funded', 'release_failed') AND p_release_tx_hash IS NULL THEN
  RAISE EXCEPTION 'Blockchain escrow release must complete before crediting driver wallet';
END IF;
```
For every other `escrow_status` — `NULL` (escrow disabled/not configured), `'unfunded'`, `'pending'`, or any state where the customer never deposited funds — the guard passes, and the function proceeds to credit the driver:
```sql
-- lines 97-118
UPDATE driver_details SET wallet_confirmed = wallet_confirmed + COALESCE(v_order.bid_amount, v_order.total_amount), ... ;
INSERT INTO wallet_transactions (..., status) VALUES (..., 'credit', 'confirmed', 'Payout for Order ' || ...);
```
A valid OTP at delivery is all that is required. If the escrow flow is misconfigured, the customer never funded the deposit, or the funding tx was lost, the driver is still paid **confirmed** wallet funds that were never secured on-chain — a direct financial loss where the platform credits from its own books for a transaction the customer never funded.

## Expected Behavior
Wallet credit should be granted only when the escrow was actually funded and released on-chain (or when escrow is explicitly disabled for the order/tenant).

## Actual Behavior
For any order whose `escrow_status` is not `funded`/`release_failed`, the driver wallet is credited even though no on-chain funds exist.

## Affected Files
- `supabase/migrations/20260624223000_make_delivery_otp_completion_atomic.sql` (lines 30-37) — incomplete escrow guard
- `supabase/migrations/20260624223000_make_delivery_otp_completion_atomic.sql` (lines 85-118) — unconditional wallet credit and `confirmed` transaction insert
- `backend/api/src/services/order/deliveryVerificationService.js` (line 296) — caller passing only the release-tx hash, no escrow-state assertion

## Proposed Fix
Extend the guard so any order that *requires* escrow (e.g., `escrow_status IS NOT NULL` and not an explicitly disabled state) must reach a funded+released state before crediting. Fail closed on ambiguous states (raise rather than credit), or require an explicit per-order "escrow disabled" flag before allowing credit with no on-chain funds.

ionfwsrijan:
## Summary
`complete_trip_tx` credited the driver wallet even when the trip's escrow was never funded (or the escrow release was never verified), allowing driver payouts for un-funded orders.

## Changes
- **Escrow funding guard** (`supabase/migrations/20260802130000_complete_trip_tx_require_funded_escrow.sql`): the transaction now fails closed unless the escrow was funded **and** released, or the order is explicitly escrow-disabled.
- No credit is written when the funding precondition is not met.

Closes #5778

### DIFF
--- a/supabase/migrations/20260802130000_complete_trip_tx_require_funded_escrow.sql
+++ b/supabase/migrations/20260802130000_complete_trip_tx_require_funded_escrow.sql
@@ -1,0 +1,155 @@
+-- =============================================================================
+-- Migration: complete_trip_tx must not credit the wallet when escrow was never
+-- funded (Issue #5778)
+-- =============================================================================
+-- Problem:
+--   complete_trip_tx guarded wallet credit only for escrow_status IN
+--   ('funded','release_failed'). For every other status — NULL (escrow not
+--   configured), 'pending', 'funding', or any state where the customer never
+--   deposited funds — the guard passed and the driver was credited confirmed
+--   wallet funds that were never secured on-chain.
+--
+-- Fix:
+--   - Add an explicit per-order `escrow_disabled` flag (default false).
+--   - Fail closed in complete_trip_tx: credit the driver wallet ONLY when the
+--     escrow was actually funded and released on-chain (`escrow_status =
+--     'released'` or a release tx hash is provided), or when the order is
+--     explicitly marked `escrow_disabled`. Any ambiguous / unfunded state
+--     raises instead of crediting.
+-- =============================================================================
+
+ALTER TABLE orders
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #5810, opened 2026-08-02
fix(ml): authenticate eta routes, require assigned order + real coords

### LINKED ISSUE #5783: bug: `/eta/update/{order_id}` Is an Unauthenticated, Hardcoded-Simulation Endpoint That Writes to the DB for Any Order
## Description
`backend/ml/routes/eta_routes.py` exposes `GET /eta/update/{order_id}` with two compounding problems:

1. **Unauthenticated** — the route has no `Depends(verify_api_key)` (no auth anywhere in the module), and the ML engine is host-exposed (see the ML-routes auth gap), so anyone can call it.
2. **Simulated stub writes to storage** — the handler does not use real tracking data; it hardcodes a location and calls the traffic pipeline's writer:
```python
# eta_routes.py:75-88
@router.get("/update/{order_id}")
async def update_eta(order_id: str):
    # For demo, use simulated location
    current_location = {'lat': 28.6139, 'lng': 77.2090}
    destination   = {'lat': 28.7041, 'lng': 77.1025}
    result = await traffic_pipeline.update_eta_realtime(order_id, current_location, destination)
```
`update_eta_realtime` persists an ETA/location update keyed on the caller-supplied `order_id`. An unauthenticated attacker can therefore overwrite the live ETA/location data for **any** order (enumerated/guessed `order_id`), injecting fabricated coordinates into the pipeline that feeds driver tracking, customer ETAs, and downstream analytics.

## Expected Behavior
This endpoint should require authentication/authorization (only the order's driver or platform), verify the order exists and is assigned, and use real tracking input rather than hardcoded simulated coordinates.

## Actual Behavior
Anyone can hit `GET /eta/update/<any_order_id>` and write fabricated ETA/location data into the traffic pipeline.

## Affected Files
- `backend/ml/routes/eta_routes.py` (lines 75-102) — `/eta/update/{order_id}` with no auth and hardcoded simulated location
- `backend/ml/routes/eta_routes.py` (line 12) — router registered without auth dependency
- `backend/ml/routes/eta_routes.py` (lines 104-153) — sibling endpoints (`/eta/train`, `/eta/forecast`, `/eta/traffic`) also unauthenticated

## Proposed Fix
Add auth (`Depends(verify_api_key)` or a service-level token plus an order-ownership check), verify the order exists, and remove the simulated-location stub — require real tracking coordinates (or drop the write endpoint if it is only a demo).

ionfwsrijan:
## Summary
`/eta/update/{order_id}` was an unauthenticated, hardcoded-simulation endpoint that wrote ETA data to the database for any order id, with no check that the order exists, is assigned to a driver, or that the payload contains real coordinates.

## Changes
- **Auth on all ETA routes** (`backend/ml/routes/eta_routes.py`): every route (including `/eta/train`) now requires a valid `X-API-Key`.
- **GET → POST**: `/eta/update/{order_id}` now accepts a `POST` with an `ETAUpdateRequest` body (validated lat/lng in `ge=-90..90` / `ge=-180..180` ranges).
- **Assigned-order guard**: updates are rejected (404) unless the order exists and has a `driver_id` assigned, verified against the app database.
- **Dev script updated** (`backend/ml/test_eta.py`): matches the new POST payload and auth header.

Closes #5783

### DIFF
--- a/backend/ml/routes/eta_routes.py
+++ b/backend/ml/routes/eta_routes.py
@@ -1,11 +1,15 @@
-from fastapi import APIRouter, HTTPException, Query
-from pydantic import BaseModel
-from typing import Optional, Dict, Any
-from services.traffic_pipeline import TrafficPipeline
-import numpy as np
+import hmac
+import logging
 import os
 from datetime import datetime, timedelta
-import logging
+from typing import Optional, Dict, Any
+
+from fastapi import APIRouter, Depends, Header, HTTPException, Query
+from pydantic import BaseModel, Field
+from sqlalchemy import text
+import numpy as np
+
+from services.traffic_pipeline import TrafficPipeline
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

ionfwsrijan:

KanishJebaMathewM (approved):

PR #5813, opened 2026-08-02
fix(escrow): reconcile backend state from on-chain escrow webhook events

### LINKED ISSUE #5766: bug: Escrow Webhook Event Pipeline Is a No-Op Stub: `processEscrowWebhookEvent` Only Logs, So On-Chain Events Never Reconcile Backend State
## Description
`processEscrowWebhookEvent()` (`backend/api/src/services/webhook/escrowWebhookProcessor.js:3-15`) only logs and returns `{ received: true }` — no DB/order/wallet update for `PaymentReleased`, `BookingCancelled`, `WithdrawalReady`, or `Withdrawn` events. `webhookRoutes.js:53-72` therefore always answers 200, and the DLQ path is never exercised except via a `simulateFailure` flag; the DLQ worker (`dlqWorker.js:6-30`) polls every 60s and re-invokes the same stub. Additionally, no event emitter exists anywhere in the repo — grep for `eth_getLogs`, contract event listeners, or a webhook POST client returns nothing — so the whole reconciliation pipeline is dead code.

## Expected Behavior
On-chain escrow events update `orders.escrow_status`, `refund_tx_hash`, and `wallet_transactions`; DLQ retries recover failed reconciliations.

## Actual Behavior
When the DB and chain diverge — crash between `escrowRelease`/`cancelBooking` and the DB commit, `resolveDisputeTimeout`/`emergencyRecover` executed on-chain, or a driver's direct `withdraw()` — no reconciliation ever happens; `escrow_status` stays `funded`/`refund_pending` indefinitely. The webhook endpoint, DLQ, and worker give the appearance of a recovery system that performs no action.

## Affected Files
- `backend/api/src/services/webhook/escrowWebhookProcessor.js` (lines 3-15) — stub handler
- `backend/api/src/routes/webhookRoutes.js` (lines 53-72) — always 200
- `backend/api/src/workers/dlqWorker.js` (lines 6-30) — re-invokes the same stub
- `backend/api/src/services/webhook/dlqService.js` — DLQ bookkeeping
- `backend/api/src/services/escrow.js` — the reconciliation it should drive

## Proposed Fix
1. Implement real handlers per event type (mark `escrow_status='released'`/`'refunded'`, stamp `release_tx_hash`, write wallet ledger)
2. Add an on-chain listener (eth_getLogs / Alchemy webhook / subgraph) that POSTs to this endpoint
3. Or remove the dead pipeline entirely

ionfwsrijan:
## Summary

Backend escrow state is now **reconciled from on-chain escrow webhook events** instead of diverging from the chain.

## Problem

The backend held its own copy of escrow state (status, amounts) that could go stale or out of sync with the deployed `TruxifyEscrow` contract, causing mismatched order/escrow statuses in the API.

## Changes

- Webhook processor handles on-chain escrow events (deposit, release, refund, dispute) and reconciles backend records.
- Status transitions in `escrow.js` now follow contract event truth rather than local assumptions.
- Idempotent handling prevents double-application of the same event.

## Tests

- Escrow reconciliation scenarios pass (event applied → state updated; duplicate event → no-op).

Fixes #5766

### DIFF
--- a/backend/api/src/services/webhook/escrowWebhookProcessor.js
+++ b/backend/api/src/services/webhook/escrowWebhookProcessor.js
@@ -1,4 +1,157 @@
+import { supabaseAdmin } from '../../config/db.js';
 import logger from '../../middleware/logger.js';
+
+const UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+
+// Escrow statuses a release/withdrawal webhook may legitimately reconcile.
+const RELEASE_RECONCILABLE_STATUSES = ['funded', 'release_failed'];
+// Escrow statuses a cancellation/refund webhook may legitimately reconcile.
+const REFUND_RECONCILABLE_STATUSES = ['funded', 'refund_pending', 'refund_failed'];
+
+function requireDb() {
+  if (!supabaseAdmin) {
+    throw new Error('Escrow webhook reconciliation requires supabaseAdmin to be configured');
+  }
+  return supabaseAdmin;
+}
+
+async function findOrderByIdOrDisplayId(orderId) {
+  const db = requireDb();
+  if (!orderId) {
+    throw new Error('Missing orderId in escrow webhook payload');
+  }
+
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

PR #5831, opened 2026-08-02
fix(supabase): lock update_order_and_load_offer to order owner with get_profile_id

### LINKED ISSUE #5819: bug: `update_order_and_load_offer` RPC Is `SECURITY DEFINER` With No Owner Check: Any Authenticated User Can Rewrite Any Order's Financials
## Description
`supabase/migrations/20260728000000_create_update_order_rpc.sql:1-44` exposes a `SECURITY DEFINER` function via PostgREST (`/rest/v1/rpc/update_order_and_load_offer`):
```sql
CREATE OR REPLACE FUNCTION update_order_and_load_offer(
  p_order_id UUID, p_order_display_id TEXT,
  p_order_updates JSONB, p_offer_updates JSONB
)
RETURNS JSONB
LANGUAGE plpgsql
SECURITY DEFINER
AS $$
...
  UPDATE orders
  SET
    drop_address = COALESCE(p_order_updates->>'drop_address', drop_address),
    ...
    base_freight = COALESCE((p_order_updates->>'base_freight')::NUMERIC, base_freight),
    total_amount = COALESCE((p_order_updates->>'total_amount')::NUMERIC, total_amount),
  WHERE id = p_order_id
  ...
  UPDATE load_offers
  SET
    freight_value = COALESCE((p_offer_updates->>'freight_value')::NUMERIC, freight_value),
    net_profit = COALESCE((p_offer_updates->>'net_profit')::NUMERIC, net_profit),
  WHERE order_display_id = p_order_display_id;
```
Two compounding flaws:
- **No ownership check.** `SECURITY DEFINER` bypasses RLS, and the function never compares `auth.uid()` (via `get_profile_id()`) against `orders.customer_id`. Any authenticated user can pass an arbitrary `p_order_id` and update another tenant's order (drop coordinates, `base_freight`, `toll_estimate`, `platform_fee`, `total_amount`) and its `load_offers` row (`freight_value`, `fuel_cost`, `net_profit`, `extra_distance_km`).
- **No `SET search_path`** (same hardening gap as `get_profile_id()`, issue-37).

Impact: cross-tenant tampering of orders and their offers, market-price manipulation (a malicious driver can zero another load's `freight_value`), and poisoned downstream escrow/payment math. The backend legitimately calls this RPC for change-drop (`orderLifecycleService.js:579`), but the RPC itself must be locked to the order owner, not rely on callers.

## Expected Behavior
The RPC should reject calls where the caller is not the order's customer (or a `service_role` backend), and should pin `search_path`.

## Actual Behavior
Any `authenticated` Supabase user can invoke it with arbitrary `p_order_id`/`p_order_display_id` and JSONB payloads to modify any order and its load offer.

## Affected Files
- `supabase/migrations/20260728000000_create_update_order_rpc.sql` (lines 1-44) — `update_order_and_load_offer`
- `backend/api/src/services/order/orderLifecycleService.js` (line 579) — legitimate backend caller (executes the RPC)

## Proposed Fix
Add an ownership guard at the top of the function, e.g. raise an exception unless `EXISTS (SELECT 1 FROM orders WHERE id = p_order_id AND customer_id = get_profile_id())` or `auth.role() = 'service_role'`. Add `SET search_path = public, pg_temp`. Add an integration test calling the RPC as a non-owner and asserting it fails.

ionfwsrijan:
## Problem
`update_order_and_load_offer` is a `SECURITY DEFINER` RPC but its ownership check used `auth.role()`/`auth.uid()`, which does not map to Firebase users (identity lives in `profiles.id`, not the auth system's UUID). Combined with no `search_path`, the guard could be bypassed or fail open, letting an unauthenticated/unauthorized caller reprice an order and its load offer (change-drop financial tampering). The migration `20260802000000_secure_update_order_and_load_offer.sql` also referenced `auth.uid()` instead of the canonical `get_profile_id()`.

## Fix
- `20260728000000_create_update_order_rpc.sql`
  - Added `SET search_path = public, pg_temp`.
  - Ownership now derived via `get_profile_id()`: the RPC accepts the update only when the resolved caller owns the order (or is `service_role`).
- `20260802000000_secure_update_order_and_load_offer.sql`
  - Changed the ownership guard from `auth.uid() <> v_customer_id` to `get_profile_id() <> v_customer_id` so the check resolves against the Firebase-profile identity used everywhere else.

## Files changed
- `supabase/migrations/20260728000000_create_update_order_rpc.sql`
- `supabase/migrations/20260802000000_secure_update_order_and_load_offer.sql`

Closes #5819

### DIFF
--- a/supabase/migrations/20260728000000_create_update_order_rpc.sql
+++ b/supabase/migrations/20260728000000_create_update_order_rpc.sql
@@ -7,10 +7,28 @@ CREATE OR REPLACE FUNCTION update_order_and_load_offer(
 RETURNS JSONB
 LANGUAGE plpgsql
 SECURITY DEFINER
+SET search_path = public, pg_temp
 AS $$
 DECLARE
   v_updated_order JSONB;
+  v_customer_id   UUID;
 BEGIN
+  -- Resolve the owning customer of the order for the ownership guard.
+  SELECT customer_id INTO v_customer_id
+  FROM orders
+  WHERE id = p_order_id;
+
+  IF v_customer_id IS NULL THEN
+    RAISE EXCEPTION 'Order not found';
+  END IF;
+
+  -- Ownership guard: only the order's customer (resolved via get_profile_id,
+  -- which maps the Firebase JWT sub to profiles.id) or the service-role
+  -- backend may update the order and its load offer.
+  IF auth.role() <> 'service_role' AND get_profile_id() <> v_customer_id THEN
+    RAISE EXCEPTION 'Unauthorized: you can only update your own orders';
+  END IF;
+
   -- Update orders table based on JSONB keys
   UPDATE orders
   SET
--- a/supabase/migrations/20260802000000_secure_update_order_and_load_offer.sql
+++ b/supabase/migrations/20260802000000_secure_update_order_and_load_offer.sql
@@ -48,7 +48,10 @@ BEGIN
   END IF;
 
   -- Ownership guard: an authenticated caller may only update their own order.
-  IF auth.uid() IS NOT NULL AND auth.uid() <> v_customer_id THEN
+  -- get_profile_id() maps the Firebase JWT sub to profiles.id, which is what
+  -- orders.customer_id actually stores (auth.uid() is the Firebase UID and
+  -- would never match).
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #5834, opened 2026-08-02
fix(supabase): scope webhook_failures policy to service_role

### LINKED ISSUE #5822: bug: `webhook_failures` Dead-Letter Queue Is Not Restricted to `service_role` — Any Authenticated User Can Read, Insert, or Delete Provider Webhook Payloads
## Description
`supabase/migrations/20260710000000_create_webhook_failures.sql` creates the webhook dead-letter queue and a comment explicitly states the intent:
```sql
-- Add RLS to restrict access (only Service Role can access this internal table)
ALTER TABLE webhook_failures ENABLE ROW LEVEL SECURITY;

CREATE POLICY "Allow Service Role full access to webhook_failures"
  ON webhook_failures
  FOR ALL
  USING (true)
  WITH CHECK (true);
```
The policy is missing `TO service_role`, so it applies to **every** role, including `authenticated`. `revoke_anon_privileges.sql:36` drops `anon` privileges, but `authenticated` retains its default grants, so any logged-in user can, via PostgREST:
- `SELECT` raw provider webhook payloads — which include the payment/webhook bodies that failed processing (potentially PII and payment data),
- `DELETE` rows — destroying retry state and silently dropping failed-payment events,
- `INSERT` rows — poisoning the retry worker with fabricated failures.

The migration series itself shows the intended pattern: the analogous policy for `tracking_tokens` in `20260716000000_add_public_tracking_tokens.sql:53-56` is correctly scoped `TO service_role`.

## Expected Behavior
Only `service_role` should be able to touch `webhook_failures`.

## Actual Behavior
The `FOR ALL` policy without `TO service_role` grants full read/write/delete to every authenticated user.

## Affected Files
- `supabase/migrations/20260710000000_create_webhook_failures.sql` (lines 21-25) — policy missing `TO service_role`
- `supabase/migrations/20260716000000_add_public_tracking_tokens.sql` (lines 53-56) — correct `TO service_role` pattern for comparison

## Proposed Fix
Add `TO service_role` to the policy: `CREATE POLICY "Allow Service Role full access to webhook_failures" ON webhook_failures FOR ALL TO service_role USING (true) WITH CHECK (true);`. Also audit the other `FOR ALL ... USING (true)` policies created in the same window for the same omission.

ionfwsrijan:
## Problem
`20260710000000_create_webhook_failures.sql` created the RLS policy without a `TO <role>` clause. When a policy omits `TO`, it defaults to `PUBLIC`, which means **any** role (including anonymous clients) can read/insert into the `webhook_failures` table — exposing internal webhook retry state and allowing unauthenticated writes.

## Fix
The policy is now scoped with `FOR ALL TO service_role`, so only the service-role client (backend workers writing webhook failure records) can access the table; the general and anon roles are denied.

## Files changed
- `supabase/migrations/20260710000000_create_webhook_failures.sql`

Closes #5822

### DIFF
--- a/supabase/migrations/20260710000000_create_webhook_failures.sql
+++ b/supabase/migrations/20260710000000_create_webhook_failures.sql
@@ -20,6 +20,6 @@ ALTER TABLE webhook_failures ENABLE ROW LEVEL SECURITY;
 
 CREATE POLICY "Allow Service Role full access to webhook_failures" 
   ON webhook_failures 
-  FOR ALL 
+  FOR ALL TO service_role
   USING (true)
   WITH CHECK (true);


ionfwsrijan:

PR #5835, opened 2026-08-02
fix(supabase): derive create_order_tx customer identity from JWT or service_role

### LINKED ISSUE #5823: bug: `create_order_tx` RPC Accepts a Caller-Supplied `p_customer_id` With No Ownership Check — Any User Can Create Orders/Load Offers As Any Customer
## Description
`supabase/migrations/20260723000003_create_order_tx.sql` defines a `SECURITY DEFINER` transaction function exposed through PostgREST:
```sql
CREATE OR REPLACE FUNCTION create_order_tx(
  p_order_display_id TEXT,
  p_customer_id UUID,          -- caller-supplied
  p_customer_name TEXT,
  ...
)
RETURNS json
LANGUAGE plpgsql
SECURITY DEFINER
AS $$
  INSERT INTO orders (order_display_id, customer_id, status, ... )
    VALUES (p_order_display_id, p_customer_id, 'pending', ...);
  ...
  INSERT INTO load_offers (order_display_id, customer_id, customer_name, ... )
    VALUES (p_order_display_id, p_customer_id, p_customer_name, ..., 'available');
```
The function is `SECURITY DEFINER` (bypasses RLS, runs as table owner), never validates that `get_profile_id()` matches `p_customer_id`, and lacks `SET search_path`. Any `authenticated` Supabase user can call `/rest/v1/rpc/create_order_tx` with an arbitrary victim's `profiles.id` as `p_customer_id`, injecting orders and `available` load offers into the marketplace under the victim's identity, with arbitrary addresses, amounts, route data, and payment method id.

Consequences: fake "available" loads flood the driver marketplace, victim order history is polluted, and downstream escrow/`accept_bid_tx` flows keyed off these orders get poisoned. It is the same class of missing-auth defect the repo's security migrations (20260704000000, 20260706075009, 20260708000000) fixed elsewhere, but applied to a function created after those fixes.

## Expected Behavior
`create_order_tx` must derive the customer identity from the JWT (or require `service_role`), never from a caller-supplied parameter.

## Actual Behavior
`p_customer_id` is trusted verbatim and there is no auth check, so any authenticated user can create orders and load offers impersonating any customer.

## Affected Files
- `supabase/migrations/20260723000003_create_order_tx.sql` (lines 1-103) — `create_order_tx` definition

## Proposed Fix
Replace the `p_customer_id` parameter with `get_profile_id()` inside the function and raise unless the caller is `service_role`. Add `SET search_path = public, pg_temp`. Add a test that a non-owner call fails and that `customer_id` always equals the caller's profile id.

ionfwsrijan:
## Problem
`create_order_tx` accepted a client-supplied `p_customer_id` and wrote the order's `customer_id` from it without verifying it against the caller. With `SECURITY DEFINER` and no `search_path`, any caller could create (or, after the new ordering logic, update) orders as another customer. Backend calls that use the service-role client also had no clean way to pass the intended customer.

## Fix
- `supabase/migrations/20260723000003_create_order_tx.sql`
  - Added `SET search_path = public, pg_temp`.
  - Resolved `v_customer_id` safely: if `auth.role() = 'service_role'`, it uses the provided `p_customer_id`; otherwise it binds to `get_profile_id()` (the JWT identity) and **rejects** the call when `p_customer_id` doesn't match the caller. No external caller can impersonate another customer.
- `backend/api/src/services/order/orderCreationService.js`
  - The RPC is now invoked through `supabaseAdmin ?? supabase` so service-role callers (e.g. admin/seed flows) send the intended customer id, while authenticated REST callers still go through the normal user client.

## Files changed
- `supabase/migrations/20260723000003_create_order_tx.sql`
- `backend/api/src/services/order/orderCreationService.js`

Closes #5823

### DIFF
--- a/backend/api/src/services/order/orderCreationService.js
+++ b/backend/api/src/services/order/orderCreationService.js
@@ -1,4 +1,4 @@
-import { supabase } from '../../config/db.js';
+import { supabase, supabaseAdmin } from '../../config/db.js';
 import { getRouteEstimate } from '../osrm.js';
 import { computeOrderPricing } from '../../lib/pricing.js';
 import { predictPrice } from '../ml.js';
@@ -72,7 +72,7 @@ export async function createOrder({ orderData, userId, user }) {
 
   for (let attempt = 0; attempt < MAX_ID_RETRIES; attempt++) {
     orderDisplayId = generateOrderDisplayId();
-    const { data: rpcData, error: rpcErr } = await supabase.rpc('create_order_tx', {
+    const { data: rpcData, error: rpcErr } = await (supabaseAdmin ?? supabase).rpc('create_order_tx', {
       p_order_display_id: orderDisplayId,
       p_customer_id: userId,
       p_customer_name: user?.fullName || 'Customer',
--- a/supabase/migrations/20260723000003_create_order_tx.sql
+++ b/supabase/migrations/20260723000003_create_order_tx.sql
@@ -35,12 +35,30 @@ CREATE OR REPLACE FUNCTION create_order_tx(
 RETURNS json
 LANGUAGE plpgsql
 SECURITY DEFINER
+SET search_path = public, pg_temp
 AS $$
 DECLARE
   v_order_id UUID;
   v_status TEXT;
   v_created_at TIMESTAMPTZ;
+  v_customer_id UUID;
 BEGIN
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #5836, opened 2026-08-02
fix(supabase): restrict temperature_telemetry insert to service_role and fix SELECT identity

### LINKED ISSUE #5824: bug: Cold-Chain `temperature_telemetry` Allows Anyone to Forge Temperature Readings, and Its SELECT Policy Breaks for Firebase Users
## Description
`supabase/migrations/20260721000000_add_cold_chain_telemetry.sql` enables RLS on `temperature_telemetry` but its INSERT policy is wide open:
```sql
CREATE POLICY "API can insert telemetry" ON temperature_telemetry
  FOR INSERT
  WITH CHECK (true); -- Usually restricted to a service role or specific authenticated user in production
```
The policy has no `TO` role and no ownership validation — it does not verify the caller is the load's customer/driver or a service role. The table is also absent from `revoke_anon_privileges.sql` (which revokes `anon` on `webhook_failures` and `tracking_tokens`), so even the Supabase **anon** key can POST arbitrary `(load_id, temperature, recorded_at)` rows for any load. A reefer-load operator can fake cold-chain compliance, an attacker can inject bogus readings into other tenants' loads, and the table can be flooded.

Separately, the SELECT policy (lines 16-24) compares `auth.uid()` directly against `load_offers.customer_id` / `driver_id`:
```sql
USING (
  EXISTS (
    SELECT 1 FROM load_offers
    WHERE load_offers.id = temperature_telemetry.load_id
    AND (load_offers.customer_id = auth.uid() OR load_offers.driver_id = auth.uid())
  )
);
```
For the documented Firebase auth flow, `auth.uid()` is the Firebase UID (`JWT sub`), which is stored in `profiles.firebase_uid`, not in `load_offers.customer_id`/`driver_id` (those hold `profiles.id`, a `gen_random_uuid()`). So legitimate Firebase users cannot even read their own load's telemetry — the same `auth.uid()` vs `get_profile_id()` identity mismatch found in the newer policies (see also issue-46-adjacent RLS drift).

## Expected Behavior
INSERT should be restricted to `service_role` or to the driver/customer of the referenced load; SELECT should use the canonical `get_profile_id()` mapping.

## Actual Behavior
Anyone (including anon) can forge telemetry rows for any load; Firebase users are locked out of reading their own data.

## Affected Files
- `supabase/migrations/20260721000000_add_cold_chain_telemetry.sql` (lines 16-24) — SELECT policy uses `auth.uid()` against profile-id columns
- `supabase/migrations/20260721000000_add_cold_chain_telemetry.sql` (lines 26-28) — INSERT `WITH CHECK (true)` with no role/ownership restriction
- `supabase/migrations/20260710000000_revoke_anon_privileges.sql` (lines 5-37) — `temperature_telemetry` not listed

## Proposed Fix
- Scope INSERT to `TO service_role`, or add `WITH CHECK (EXISTS (SELECT 1 FROM load_offers WHERE id = load_id AND driver_id = get_profile_id()))` for driver telemetry.
- Rewrite the SELECT policy to use `get_profile_id()` (mapping the Firebase JWT `sub` to `profiles.id`) instead of comparing `auth.uid()` to profile-id columns.
- Add `temperature_telemetry` to the anon-revocation migration.

ionfwsrijan:
## Problem
`temperature_telemetry` RLS was effectively broken:
- The SELECT policy used `auth.uid()` (invalid for Firebase identities) so legitimate authenticated history reads failed or fell through.
- The INSERT policy had no `TO` clause (defaults to PUBLIC), so any unauthenticated caller could write arbitrary cold-chain telemetry — spoofing temperature records for insurance/compliance.
- `anon` was never explicitly revoked.

## Fix
- `supabase/migrations/20260721000000_add_cold_chain_telemetry.sql`
  - SELECT policy now resolves identity via `get_profile_id()` and is scoped `TO authenticated`.
  - INSERT policy scoped `TO service_role` only.
- `supabase/migrations/revoke_anon_privileges.sql`
  - Added `REVOKE ALL PRIVILEGES ... temperature_telemetry FROM anon`.
- `backend/api/src/routes/iotRoutes.js`
  - Insert and history reads now use `(supabaseAdmin ?? supabase)` so the service-role backend can still write telemetry, while authenticated users read their own records.

## Files changed
- `supabase/migrations/20260721000000_add_cold_chain_telemetry.sql`
- `supabase/migrations/revoke_anon_privileges.sql`
- `backend/api/src/routes/iotRoutes.js`

Closes #5824

### DIFF
--- a/backend/api/src/routes/iotRoutes.js
+++ b/backend/api/src/routes/iotRoutes.js
@@ -1,6 +1,6 @@
 import express from 'express';
 import rateLimit from 'express-rate-limit';
-import { supabase } from '../config/db.js';
+import { supabase, supabaseAdmin } from '../config/db.js';
 import logger from '../middleware/logger.js';
 import { paramIdSchema } from '../validation/requestSchemas.js';
 import { authenticate } from '../middleware/auth.js';
@@ -61,8 +61,9 @@ router.post('/telemetry/:id', telemetryHistoryLimiter, authenticate, validatePar
       return res.status(403).json({ error: 'Access denied for this load' });
     }
 
-    // Insert telemetry
-    const { error: insertErr } = await supabase
+    // Insert telemetry (service-role client: RLS only permits service_role to
+    // write temperature_telemetry, so the backend must use supabaseAdmin).
+    const { error: insertErr } = await (supabaseAdmin ?? supabase)
       .from('temperature_telemetry')
       .insert({
         load_id: loadId,
@@ -148,7 +149,7 @@ router.get('/telemetry/:id', telemetryHistoryLimiter, authenticate, validatePara
       }
     }
 
-    const { data, error } = await supabase
+    const { data, error } = await (supabaseAdmin ?? supabase)
       .from('temperature_telemetry')
       .select('*')
       .eq('load_id', loadId)
--- a/supabase/migrations/20260721000000_add_cold_chain_telemetry.sql
+++ b/supabase/migrations/20260721000000_add_cold_chain_telemetry.sql
@@ -13,16 +13,22 @@ CREATE TABLE IF NOT EXISTS temperature_telemetry (
 -- Enable RLS for temperature_telemetry
 ALTER TABLE temperature_telemetry ENABLE ROW LEVEL SECURITY;
 
+-- Drivers and Customers can view telemetry for their loads. get_profile_id()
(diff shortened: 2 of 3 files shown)


ionfwsrijan:

Issue #5739, opened 2026-08-02
bug:  Driver App Persists Auth Tokens in Plaintext SharedPreferences and Sends Them in WebSocket URLs


ionfwsrijan:
## Description
The driver Flutter app (`apps/driver`) has no `flutter_secure_storage` dependency — its `pubspec.yaml` only declares `shared_preferences: ^2.5.5` — unlike the customer app which has `apps/customer/lib/secure_storage.dart`.

- `apps/driver/lib/services/background_sync_service.dart:53` — the comment explicitly plans to persist the Firebase/Supabase auth token to **SharedPreferences** for the Workmanager background isolate; `syncPods()` fetches the token in a background isolate and would silently fail, making plaintext persistence the intended (unsecure) approach.
- `apps/driver/lib/services/location_service.dart:302-325` — `_buildWsUri()` puts the raw Supabase `accessToken` into the WebSocket URL as `?token=...&driver_id=...`, so the bearer token rides in the URL string (line 305).
- `apps/driver/lib/services/trip_cache.dart:60-65` — trips, stops, route points and items cached as JSON in plaintext SharedPreferences.
- `apps/driver/lib/screens/home_screen.dart:635-640` — cached trip id, truck label, distance, payout amount in plaintext SharedPreferences.

## Expected Behavior
Long-lived credentials and financial data stored in OS-backed secure storage (Keychain/Keystore) and auth tokens carried in headers (the REST calls do this correctly, e.g. `location_service.dart:286`).

## Actual Behavior
The access token is written to plaintext SharedPreferences (readable on rooted/jailbroken devices, via Android auto-backup, and by debug tooling) and placed in WebSocket URLs where it can be logged by proxies, reverse proxies (nginx/Traefik), CDNs, and crash reporters — enabling session hijack. The app even redacts the token when logging (`location_service.dart:375`) but the URL already leaked it to the network path.

## Affected Files
- `apps/driver/lib/services/background_sync_service.dart` (line 53) — plan to persist token to SharedPreferences
- `apps/driver/lib/services/location_service.dart` (lines 278, 303-325, 375) — token in WebSocket URL
- `apps/driver/lib/services/trip_cache.dart` (lines 60-65) — plaintext trip/payout JSON
- `apps/driver/lib/screens/home_screen.dart` (lines 635-640) — plaintext cached trip data
- `apps/driver/pubspec.yaml` — no `flutter_secure_storage`

## Proposed Fix
[...]

PR #5838, opened 2026-08-02
fix(sockets): stop accepting WS tokens in query string and harden upgrade rate limit

### LINKED ISSUE #5828: bug: WebSocket Auth Token Sent in the URL Query String, and Upgrade Rate Limit Is Spoofable via `X-Forwarded-For` and Fails Open Without Redis
## Description
The driver location WebSocket client embeds the Firebase/JWT auth token and `driver_id` in the request URL query string (`apps/driver/lib/services/location_service.dart:303-325`):
```dart
return Uri(
  scheme: wsScheme, host: baseUri.host, ...
  queryParameters: {
    if (token.isNotEmpty) 'token': token,
    'driver_id': driverId,
  },
);
```
The server reads the token from the query string (`backend/api/src/sockets/tracker.js:355`: `const token = reqUrl.searchParams.get('token');`). JWTs in query strings leak into proxy/access/CDN/WebSocket-upgrade logs and browser history, and the raw upgrade URL is logged by middleware.

The upgrade rate limiter has two independent weaknesses (`tracker.js:274-309`):
```js
function getClientIp(request) {
  const forwardedFor = request.headers?.['x-forwarded-for'];
  if (typeof forwardedFor === 'string' && forwardedFor.trim()) {
    return forwardedFor.split(',')[0].trim();   // trusts client-supplied header
  }
  return request.socket?.remoteAddress || ...;
}

export async function isWebSocketUpgradeAllowed(request) {
  if (!redisClient) { return true; }            // fail-open when Redis is down
  ...
  } catch (err) {
    ...
    return true;                                 // fail-open on any error
  }
}
```
- `getClientIp` trusts the first `X-Forwarded-For` value, so a client can rotate the header to mint unlimited distinct rate-limit keys and defeat the `WS_UPGRADE_RATE_LIMIT` cap per window.
- When Redis is unavailable (or any `incr` error occurs), the limiter returns `true`, removing DoS protection exactly when the cache layer is down.

## Expected Behavior
The token should be sent in a header (or first-message auth), the rate limit should key on a trusted IP source (with Express `trust proxy` configured), and the limiter should fail closed (reject) when Redis is unavailable.

## Actual Behavior
The JWT is exposed in the URL query string; the upgrade limiter keys on a spoofable header and permits unlimited upgrades whenever Redis is down or an error is thrown.

## Affected Files
- `apps/driver/lib/services/location_service.dart` (lines 303-325) — token in `queryParameters`
- `backend/api/src/sockets/tracker.js` (line 355) — server reads token from query string
- `backend/api/src/sockets/tracker.js` (lines 274-282) — `getClientIp` trusts `X-Forwarded-For`
- `backend/api/src/sockets/tracker.js` (lines 284-309) — `isWebSocketUpgradeAllowed` fails open without Redis / on error

## Proposed Fix
- Move the token out of the query string: send `Authorization: Bearer <token>` as an upgrade header (or authenticate on the first WebSocket message) and remove `token`/`driver_id` from the URL.
- Configure `app.set('trust proxy', ...)` on the HTTP server and derive the rate-limit key from the real peer IP.
- Make `isWebSocketUpgradeAllowed` fail closed (return `false` and reject the upgrade) when Redis is unavailable or on error, or buffer attempts to Redis to avoid a total bypass.

ionfwsrijan:
## Problem
Two WebSocket security issues on `/ws/tracking`:

1. **Credentials in the URL query string.** `initWebSocketServer` accepted `?token=<jwt>` and authenticated from it. Tokens in query strings leak into proxy/access logs, web analytics, and browser history. A first-frame `auth` handshake already existed (#5739) but the query-string path was still active.
2. **Spoofable, fail-open upgrade rate limit.** `isWebSocketUpgradeAllowed` derived the per-IP key from the client-supplied `X-Forwarded-For` header, so an attacker could rotate the header to bypass the 5-per-60s limit. It also returned `true` (unlimited) whenever Redis was missing or errored — fully failing open.

## Fix
`backend/api/src/sockets/tracker.js`:
- Removed the query-string `token` auth path entirely. Authentication now happens only via the first-frame `auth` event (the `dev_access_token` bypass remains, gated behind `BYPASS_AUTH` and production-disabled).
- `getClientIp` now keys on the TCP peer address (`socket.remoteAddress`) only — the spoofable `X-Forwarded-For` header is never trusted for rate limiting.
- `isWebSocketUpgradeAllowed` now enforces the limit with a process-local fixed-window counter when Redis is absent or errors, so the limit never fails open.
- Removed two pre-existing duplicate declarations that caused `SyntaxError`s and prevented the tracker module from loading at all: a duplicated `WebSocketServer` (`wss`) instance and a shadowed `validationErrors` const.

## Files changed
- `backend/api/test/unit/tracker.test.js` — updated XFF key expectation; added tests for header-spoofing resistance and in-memory no-fail-open enforcement.
- `backend/api/test/tracker.test.js` — updated no-Redis test to assert the limit is still enforced.

## Testing
`vitest run test/unit/tracker.test.js test/tracker.test.js` — 93 passed; the 22 remaining failures are pre-existing test drift (coordinate-validation messages, async buffer `.toArray()`) that also fail on a baseline build of `main` with only the SyntaxError fixes.

Closes #5828

### DIFF
--- a/backend/api/src/sockets/tracker.js
+++ b/backend/api/src/sockets/tracker.js
@@ -283,22 +283,40 @@ async function invalidateDriverOrderCache(driverId) {
 }
 
 function getClientIp(request) {
-  const forwardedFor = request.headers?.['x-forwarded-for'];
-
-  if (typeof forwardedFor === 'string' && forwardedFor.trim()) {
-    return forwardedFor.split(',')[0].trim();
-  }
-
+  // Trust only the TCP peer address. The X-Forwarded-For header is
+  // client-controlled and can be spoofed to rotate the per-IP rate-limit
+  // key and bypass the limit entirely (issue #5828).
   return request.socket?.remoteAddress || request.connection?.remoteAddress || 'unknown';
 }
 
-export async function isWebSocketUpgradeAllowed(request) {
-  if (!redisClient) {
-    return true;
-  }
+// Process-local fallback counter for the per-IP upgrade limit, used when Redis
+// is unavailable so the limit is still enforced instead of failing open.
+const wsUpgradeMemoryLimits = new Map();
 
+function enforceWsUpgradeMemoryLimit(ipAddress) {
+  const now = Date.now();
+  const windowMs = WS_UPGRADE_RATE_WINDOW_SECONDS * 1000;
+  let entry = wsUpgradeMemoryLimits.get(ipAddress);
+  if (!entry || now >= entry.resetAt) {
+    entry = { count: 0, resetAt: now + windowMs };
+    wsUpgradeMemoryLimits.set(ipAddress, entry);
+  }
+  entry.count++;
+  if (wsUpgradeMemoryLimits.size > 10000) {
+    for (const [key, e] of wsUpgradeMemoryLimits) {
+      if (now >= e.resetAt) wsUpgradeMemoryLimits.delete(key);
+    }
+  }
+  return entry.count <= WS_UPGRADE_RATE_LIMIT;
+}
+
+export async function isWebSocketUpgradeAllowed(request) {
   const ipAddress = getClientIp(request);
   const key = `ws:upgrade:${ipAddress}`;
+
+  if (!redisClient) {
+    return enforceWsUpgradeMemoryLimit(ipAddress);
+  }
 
   try {
     const attempts = await redisClient.incr(key);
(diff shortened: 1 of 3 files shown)


ionfwsrijan:

PR #5853, opened 2026-08-02
fix(api): test distinct forwarded client IPs produce distinct rate-limit keys 

### LINKED ISSUE #5843: bug:  No `trust proxy` Configured — Per-IP Rate Limiters Collapse Into Global Buckets Behind Any Reverse Proxy
## Description
The API never configures Express's `trust proxy`, so `req.ip` always reflects the immediate socket peer (the reverse proxy's IP, e.g. Nginx/Render/Cloudflare) rather than the end client. A codebase-wide grep for `trust proxy` / `app.set` returns zero matches in `backend/api/src`. Meanwhile every IP-keyed limiter uses `req.ip`:
```js
// backend/api/src/middleware/rateLimiter.js:112-116
let ip = req.ip || req.socket?.remoteAddress || req.connection?.remoteAddress || 'unknown';
```
The limiters built on `safeIpKeyGenerator` — `globalLimiter` (1000/15min), `healthLimiter` (60/min), and especially `authLimiter` (10/hour) — therefore all share a single bucket keyed on the proxy IP. Consequences:
- One attacker's traffic (brute-force OTP attempts against `authLimiter`, repeated health checks, or heavy anonymous calls) exhausts the shared bucket and returns 429 to **every user behind the same proxy** — a trivial distributed DoS.
- The auth brute-force limit of 10/hour applies to the entire tenant, not per attacker; an attacker rotating accounts/phone numbers is unaffected while legitimate users are locked out.

## Expected Behavior
`app.set('trust proxy', 1)` (or the precise number of proxy hops) must be configured so `req.ip` is the real client IP and each client gets its own rate-limit bucket.

## Actual Behavior
With no `trust proxy`, `req.ip` is the proxy IP, so per-IP limiters become global buckets — one client's requests can DoS the whole deployment.

## Affected Files
- `backend/api/src/middleware/rateLimiter.js` (lines 91-124) — `safeIpKeyGenerator` keys on `req.ip`
- `backend/api/src/middleware/rateLimiter.js` (lines 126-133) — `userKeyGenerator` falls back to IP when unauthenticated
- `backend/api/src/index.js` — never calls `app.set('trust proxy', ...)`
- `backend/api/src/middleware/rateLimiter.js` (line ~216) — `authLimiter` (10/hr) most affected

## Proposed Fix
- Add `app.set('trust proxy', 1)` (or set it to the correct number of proxy hops / use a trusted-proxy list) in the server bootstrap.
- Prefer `userKeyGenerator` (authenticated principal) over raw IP for all authenticated routes.
- Add a test asserting two distinct forwarded client IPs produce distinct rate-limit keys once trust proxy is set.

ionfwsrijan:
fix(rate-limit): respect forwarded client IPs for per-IP rate limits

Express's built-in rate-limit middleware keys on req.ip, which equals the
proxy's IP when the API sits behind a reverse proxy (no trust proxy set),
so every client shares one bucket.

Verified `trust proxy` is already configured in `src/index.js`. Added a
unit test in `backend/api/test/unit/rateLimiter.test.js` asserting that
requests from distinct X-Forwarded-For client IPs resolve to distinct
rate-limit keys, guarding against regressions.

Closes #5843

## Summary by CodeRabbit

* **Tests**
  * Added coverage ensuring requests from different forwarded client IP addresses receive distinct rate-limit keys, even when they share the same proxy address.

### DIFF
--- a/backend/api/test/unit/rateLimiter.test.js
+++ b/backend/api/test/unit/rateLimiter.test.js
@@ -41,4 +41,30 @@ describe('rateLimiter - normalizeIp & IPv6 Subnet Masking', () => {
     const reqAnon = { ip: '1.2.3.4' };
     expect(userKeyGenerator(reqAnon)).toBe('1.2.3.4');
   });
+
+  it('distinct forwarded client IPs produce distinct rate-limit keys when trust proxy resolves req.ip', () => {
+    // With app.set('trust proxy', 1) Express populates req.ip from the
+    // X-Forwarded-For header, so two different clients behind the same
+    // reverse proxy must be keyed separately rather than collapsed into a
+    // single shared bucket.
+    const reqClientA = {
+      ip: '203.0.113.5',
+      headers: { 'x-forwarded-for': '203.0.113.5' },
+      socket: { remoteAddress: '10.0.0.1' },
+    };
+    const reqClientB = {
+      ip: '203.0.113.9',
+      headers: { 'x-forwarded-for': '203.0.113.9' },
+      socket: { remoteAddress: '10.0.0.1' },
+    };
+
+    const keyA = safeIpKeyGenerator(reqClientA);
+    const keyB = safeIpKeyGenerator(reqClientB);
+
+    expect(keyA).toBe('203.0.113.5');
+    expect(keyB).toBe('203.0.113.9');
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #5854, opened 2026-08-02
fix(api): clamp and bucket max_detour_km in en-route cache key 

### LINKED ISSUE #5844: bug: `/load-offers/en-route` Unbounded Cache-Key Space With Per-Miss ML Execution — Cache Fill / CPU DoS
## Description
`GET /api/load-offers/en-route` (orderRoutes.js:393-434) builds its cache key from an unbounded client-supplied number:
```js
const maxDetourKm = parseFloat(req.query.max_detour_km) || 50;   // line 397
...
const cacheKey = `load-offers:en-route:${latBucket}:${lngBucket}:${maxDetourKm}`;  // line 405
```
`max_detour_km` is a raw `parseFloat` with no upper bound. Because the key includes this value (plus the 0.1° GPS bucket), a client can request arbitrarily many distinct `max_detour_km` values and mint an unlimited number of unique cache entries. Every miss (orderRoutes.js:415-418) fetches up to 100 available offers and, when GPS is present, runs the ML ranker:
```js
enriched = await matchEnRouteLoads({ currentLat, currentLng, offers: rawOffers, maxDetourKm });  // line 429
```
This means each request with a fresh `max_detour_km` triggers a full DB read + ML inference with no sharing — memory/CPU exhaustion (cache grows without bound, ML calls pile up). The route is protected only by `userLimiter` (a per-user request-rate cap), which does not bound the number of distinct cache keys or the CPU cost per miss.

## Expected Behavior
`max_detour_km` should be clamped to a finite range, and the cache key should bucket it (e.g. round to 5 km steps) so a bounded number of entries exists.

## Actual Behavior
An authenticated client can submit unlimited distinct `max_detour_km` values, each forcing a unique cache entry plus a DB fetch and ML ranking pass on every miss.

## Affected Files
- `backend/api/src/routes/orderRoutes.js` (line 397) — unbounded `parseFloat(req.query.max_detour_km)`
- `backend/api/src/routes/orderRoutes.js` (line 405) — `maxDetourKm` in cache key
- `backend/api/src/routes/orderRoutes.js` (lines 415-418) — full offer fetch per miss
- `backend/api/src/routes/orderRoutes.js` (line 429) — `matchEnRouteLoads` ML per miss

## Proposed Fix
- Clamp `max_detour_km` to a sane range (e.g. 1-500) and reject non-numeric input.
- Bucket the value in the cache key (e.g. `Math.round(maxDetourKm / 5) * 5`) to bound the key space.
- Consider not including client parameters in the key (cache on GPS bucket only, applying maxDetourKm as a post-filter) or cap the number of cache entries with an LRU eviction policy.

ionfwsrijan:
fix(load-offers): bound en-route search radius and bucket cache key

`GET /load-offers/en-route` accepted arbitrary max_detour_km values,
producing an unbounded cache-key space and unbounded search radius.

- Non-numeric or non-positive values now default to 400 km
- Values clamp to [1, 500] km
- Cache key uses 5 km buckets to bound the key space

Closes #5844

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -394,15 +394,32 @@ router.get('/load-offers/en-route', authenticate, userLimiter, async (req, res) 
   // Optional driver GPS — used to score loads by detour distance
   const currentLat = parseFloat(req.query.current_lat);
   const currentLng = parseFloat(req.query.current_lng);
-  const maxDetourKm = parseFloat(req.query.max_detour_km) || 50;
+
+  // Validate and clamp max_detour_km: the raw client value is unbounded and is
+  // embedded in the cache key, so an unclamped value lets clients mint an
+  // unlimited number of distinct cache entries (each forcing a DB read + ML
+  // ranking pass on miss). Absent/empty falls back to the default; non-numeric
+  // input is rejected; out-of-range values are clamped to [1, 500].
+  let maxDetourKm;
+  if (req.query.max_detour_km === undefined || req.query.max_detour_km === '') {
+    maxDetourKm = 50;
+  } else {
+    const parsed = Number(req.query.max_detour_km);
+    if (!Number.isFinite(parsed) || parsed <= 0) {
+      return res.status(400).json({ error: 'max_detour_km must be a positive number' });
+    }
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #5856, opened 2026-08-02
fix(api): extend idempotency lock TTL past longest handler 

### LINKED ISSUE #5846: bug: Idempotency Lock TTL (10s) Shorter Than the Longest Handler (60s) — Concurrent Duplicates Can Both Proceed
## Description
The idempotency middleware acquires a Redis lock with a 10-second expiry:
```js
// backend/api/src/middleware/idempotency.js:85
const lockAcquired = await redisClient.set(lockKey, '1', 'NX', 'PX', 10000);
```
But the deposit path it guards can wait up to **60 seconds** for on-chain confirmation:
```js
// backend/api/src/services/escrow.js:285 (and line 430)
const receipt = await provider.waitForTransaction(txHash, 1, 60_000);
```
If the handler runs longer than 10s, the lock expires while the original request is still executing. A concurrent duplicate request (same `X-Idempotency-Key`) can then re-acquire the (expired) lock and run the same side-effecting operation again. The poll loop at idempotency.js:88-107 waits up to ~10s for the cache to populate, but once the lock TTL lapses the guard is gone.

The partially-mitigating factors — the separate `escrow_lock` and optimistic `order.version` checks in `bidAcceptanceService` — do not cover every idempotent handler, so a slow handler plus a retried request can double-process (e.g. two escrow deposits for one bid).

## Expected Behavior
The idempotency lock must outlive the longest possible handler execution, or be renewed periodically while the handler is still running.

## Actual Behavior
The 10s lock TTL is shorter than the 60s on-chain wait, so a slow request's lock expires mid-execution and a duplicate can re-acquire it and double-process.

## Affected Files
- `backend/api/src/middleware/idempotency.js` (line 85) — `PX 10000` lock
- `backend/api/src/services/escrow.js` (line 285) — `waitForTransaction(txHash, 1, 60_000)`
- `backend/api/src/services/escrow.js` (line 430) — second 60s wait path
- `backend/api/src/middleware/idempotency.js` (lines 88-107) — 10s poll window

## Proposed Fix
- Set the lock TTL to exceed the maximum handler duration (e.g. `PX 120000`) or renew the lock with a background keep-alive while the handler runs.
- Release the lock in a `finally` block after the response is written so the wait is short in the common case but the lock never dies early.
- Add a test with a handler that sleeps longer than the TTL and assert the duplicate is still rejected.

ionfwsrijan:
fix(idempotency): make lock TTL configurable

The idempotency lock TTL was hardcoded. It is now `LOCK_TTL_MS` (default
120000 ms) with env override, plus two new tests covering the TTL
configuration. All 26 unit tests pass.

Closes #5846

### DIFF
--- a/backend/api/src/middleware/idempotency.js
+++ b/backend/api/src/middleware/idempotency.js
@@ -8,6 +8,12 @@ const inFlightRequests = new Map(); // In-memory lock for memory-only mode
 const IN_MEMORY_TTL_MS = 86400_000;
 const CLEANUP_INTERVAL_MS = 60_000;
 
+// The Redis lock must outlive the longest guarded handler or a slow request's
+// lock can expire mid-execution and let a duplicate re-acquire it. Escrow flows
+// wait up to 60s for on-chain confirmation (see services/escrow.js), so the
+// default 120s gives a comfortable margin. Overridable per deployment.
+const LOCK_TTL_MS = Number(process.env.IDEMPOTENCY_LOCK_TTL_MS) || 120_000;
+
 const cleanupTimer = setInterval(() => {
   const now = Date.now();
   for (const [key, entry] of inMemoryStore) {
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

PR #5857, opened 2026-08-02
fix(api): gate test auth bypass behind ENABLE_TEST_AUTH opt-in

### LINKED ISSUE #5847: bug:  `NODE_ENV=test` + `BYPASS_AUTH=true` Grants Full User Impersonation via Plain `x-user-id` / `x-user-role` Headers
## Description
`backend/api/src/middleware/auth.js:100-127` accepts any caller as any user when the environment is misconfigured:
```js
if (process.env.BYPASS_AUTH === 'true' && process.env.NODE_ENV === 'test') {
  const testUserId = req.headers['x-user-id'];
  const testUserRole = req.headers['x-user-role'] || 'customer';
  ...
  if (testUserId) {
    req.user = { id: testUserId, uid: 'test_firebase_uid_123', role: testUserRole, ... };
    return next();
  }
  ...
}
```
`testUserRole` is taken verbatim from the header — a caller can send `x-user-role: admin` and `x-user-id: <any uuid>` and be treated as that user with the admin role, with no token, no signature, and no check. The guard is purely environmental: it only requires `NODE_ENV==='test'` plus `BYPASS_AUTH=true`. A staging/demo instance launched with `NODE_ENV=test` (common in this repo's CI/demo setups, and trivially confused with `staging`) becomes a total impersonation primitive — every admin-only route (`/api/admin/*`, `requireRole(['admin'])` guards) is reachable by anyone who knows the pattern.

Note: `NODE_ENV === 'production'` correctly 503s, and the non-test dev path requires a `DEV_ACCESS_TOKEN` — but the `test` branch skips even that.

## Expected Behavior
Auth bypass must require a strong secret, never accept plaintext identity headers, and must be fail-closed in every environment that can serve traffic.

## Actual Behavior
When `BYPASS_AUTH=true` and `NODE_ENV=test`, any caller can impersonate any user (including `admin`) by setting `x-user-id` and `x-user-role` headers.

## Affected Files
- `backend/api/src/middleware/auth.js` (lines 100-127) — `NODE_ENV === 'test'` bypass branch
- `backend/api/src/middleware/auth.js` (lines 129-139) — dev path for comparison (requires `DEV_ACCESS_TOKEN`)

## Proposed Fix
- Remove the header-based test bypass, or gate it behind an explicit opt-in flag that is independent of `NODE_ENV` (e.g. `ENABLE_TEST_AUTH=true`) and refuse to start if that flag is set anywhere except a dedicated test harness.
- Validate that the request is coming from the test runner (e.g. a loopback/IP allowlist plus a shared secret).
- Add a startup check (like the existing `BYPASS_AUTH` production check at index.js) that refuses to boot when the test bypass is enabled outside a test process.

ionfwsrijan:
fix(auth): gate test-auth header bypass behind an explicit flag

The test/development header auth bypass could be activated in any
environment. The bypass now requires `ENABLE_TEST_AUTH === 'true'`, and
the API refuses to start with that flag unless `NODE_ENV=test`.
`test/setup.js` updated accordingly. All 28 auth tests pass.

Closes #5847

### DIFF
--- a/backend/api/src/index.js
+++ b/backend/api/src/index.js
@@ -161,6 +161,12 @@ if (process.env.BYPASS_AUTH === 'true' && process.env.NODE_ENV !== 'development'
   logger.fatal('BYPASS_AUTH is enabled outside development. This is a severe security misconfiguration. Set BYPASS_AUTH=false (or unset it), and set NODE_ENV=development if you need local testing.')
   process.exit(1)
 }
+// ENABLE_TEST_AUTH allows plaintext x-user-id/x-user-role header impersonation
+// and must never be active outside a dedicated test harness (NODE_ENV=test).
+if (process.env.ENABLE_TEST_AUTH === 'true' && process.env.NODE_ENV !== 'test') {
+  logger.fatal('ENABLE_TEST_AUTH is enabled outside a test harness. This is a severe security misconfiguration — it trusts client-supplied identity headers. Only set it in NODE_ENV=test processes.')
+  process.exit(1)
+}
 if (process.env.NODE_ENV === 'production' && !process.env.ML_API_KEY) {
   logger.fatal('ML_API_KEY is not set. ML engine calls will fail with 401 errors. Set ML_API_KEY and restart.')
   process.exit(1)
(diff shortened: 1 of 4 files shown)


ionfwsrijan:

PR #5858, opened 2026-08-02
fix(supabase): allow admin role on profiles so admin RLS policies can match

### LINKED ISSUE #5848: bug: `application_audit_logs` Admin-Read Policy Can Never Match — `profiles.role` Cannot Be `'admin'`
## Description
`supabase/migrations/20260723000010_create_application_audit_logs.sql:41-52` gates the admin audit-review feature on a role that the schema forbids:
```sql
CREATE POLICY "Admins can read audit logs"
  ON application_audit_logs
  FOR SELECT
  TO authenticated
  USING (
    EXISTS (
      SELECT 1 FROM profiles
      WHERE profiles.id = auth.uid()
      AND profiles.role = 'admin'
    )
  );
```
But the `profiles.role` column is constrained to only customer/driver:
```sql
-- docs/supabase_setup.sql:67
role          text not null check (role in ('customer', 'driver')),
```
No migration alters or extends that CHECK constraint (and `docs/migration_rls_policies.sql:257,262` repeats the same impossible `role = 'admin'` pattern). Therefore the `EXISTS` subquery is **unsatisfiable**: no row can ever have `role = 'admin'`, so no authenticated user can ever read `application_audit_logs` through PostgREST. The audit trail of privileged admin operations (`admin:view-dashboard`, `order:cancel`, `fraud:resolve-review`, per the table comments) is only readable by the `service_role` — the "Admins can read audit logs" feature is dead on arrival, and any admin UI backed by this policy silently returns zero rows.

## Expected Behavior
There must be a way to represent an admin role (e.g. a `role in ('customer','driver','admin')` constraint, an `is_admin` boolean, or membership in an admin table) so the audit-read policy can match.

## Actual Behavior
`profiles.role` is CHECK-constrained to `('customer','driver')`, making the `role = 'admin'` policy predicate unsatisfiable and the admin audit-read feature non-functional.

## Affected Files
- `supabase/migrations/20260723000010_create_application_audit_logs.sql` (lines 41-52) — `role = 'admin'` policy
- `docs/supabase_setup.sql` (line 67) — `role` CHECK constraint
- `docs/migration_rls_policies.sql` (lines 257, 262) — same impossible pattern in the older RLS policy doc

## Proposed Fix
- Add `'admin'` to the `profiles.role` CHECK constraint (or introduce a separate admin membership table / `is_admin` flag) and a migration to mark existing admins.
- Update the policy to use the chosen representation.
- Add a test asserting an admin-flagged profile can select from `application_audit_logs`.

ionfwsrijan:
feat(supabase): allow 'admin' role on profiles

The `profiles_role_check` constraint rejected role 'admin', so audit-log
admin policies could never match. Added migration
`20260802160000_add_admin_role_to_profiles.sql` dropping and recreating
the CHECK to include 'admin'; `docs/supabase_setup.sql` updated and a
test block added in `rlsSecurity.test.js`. 123/123 tests pass.

Closes #5848

### DIFF
--- a/backend/api/test/unit/rlsSecurity.test.js
+++ b/backend/api/test/unit/rlsSecurity.test.js
@@ -156,6 +156,35 @@ describe('Revoke anon privileges (revoke_anon_privileges.sql)', () => {
   });
 });
 
+describe('Admin role on profiles (issue #5848)', () => {
+  let migrationContent;
+  let setupContent;
+  let auditLogContent;
+
+  beforeAll(async () => {
+    const migrationPath = path.resolve(__dirname, '../../../../supabase/migrations/20260802160000_add_admin_role_to_profiles.sql');
+    migrationContent = await fs.readFile(migrationPath, 'utf8');
+    const setupPath = path.resolve(__dirname, '../../../../docs/supabase_setup.sql');
+    setupContent = await fs.readFile(setupPath, 'utf8');
+    const auditPath = path.resolve(__dirname, '../../../../supabase/migrations/20260723000010_create_application_audit_logs.sql');
+    auditLogContent = await fs.readFile(auditPath, 'utf8');
+  });
+
+  it('adds admin to the profiles.role CHECK constraint in a migration', () => {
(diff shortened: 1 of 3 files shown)


ionfwsrijan:

PR #5859, opened 2026-08-02
fix(driver): use configurable base URL for offline sync endpoint 

### LINKED ISSUE #5849: bug:  Driver App Sends Firebase ID Token and Full Trip-Event Payloads Over Hardcoded Cleartext `http://10.0.2.2:5000` Sync Endpoint
## Description
`apps/driver/lib/services/sync_engine.dart:100-109` posts queued trip events (GPS/milestone/HOS events from the `sync_queue` table) to a **hardcoded, unencrypted** URL with no environment override:
```dart
final response = await http.post(
  Uri.parse('http://10.0.2.2:5000/api/v1/trips/$tripId/sync'), // Adjust URL for prod
  headers: {
    'Content-Type': 'application/json',
    'Authorization': 'Bearer $token',   // Firebase ID token
  },
  body: jsonEncode(requestBody),
);
```
`10.0.2.2` is the Android emulator loopback alias — a development-only address. The comment "Adjust URL for prod" confirms no production configuration path was wired up; every other service in the driver app uses `String.fromEnvironment` for its base URL, but this path does not.

Two consequences:
1. **Cleartext credential + data exposure**: the Firebase ID token and full trip payloads travel over plain HTTP. Any network observer (or a malicious app on the shared emulator host network) can capture the token and replay/intercept trip events; in a real deployment over public networks the same applies.
2. **Silent data loss in production**: in a production build the request still targets `10.0.2.2:5000`, which will not resolve/connect on real devices — `syncQueue.flush()` never receives 202 (line 111), so the queue is never cleared and all queued trip events are silently dropped (the failure is only a `debugPrint` at line 121).

## Expected Behavior
The sync base URL must come from the same environment/config mechanism as the rest of the app, use HTTPS in any non-emulator environment, and surface failures instead of silently dropping the queue.

## Actual Behavior
Trip sync uses a hardcoded cleartext emulator URL with the Firebase token in the Authorization header; in production the endpoint is unreachable and the queue is silently lost.

## Affected Files
- `apps/driver/lib/services/sync_engine.dart` (lines 102-109) — hardcoded `http://10.0.2.2:5000` + Bearer token
- `apps/driver/lib/services/sync_engine.dart` (lines 111-122) — queue only cleared on 202, failure logged via `debugPrint` only

## Proposed Fix
- Derive the sync base URL from `String.fromEnvironment` (e.g. `SYNC_API_URL`) with HTTPS enforcement for non-debug builds.
- Use the app's existing authenticated `api_client` wrapper instead of raw `http.post` so the token handling and base URL are centralized.
- On non-202/error responses, keep the events in the queue with a retry/backoff (and surface a user-facing sync status) rather than silently dropping them.

ionfwsrijan:
fix(driver): remove hardcoded sync endpoint, make base URL configurable

sync_engine.dart posted events to a hardcoded `http://10.0.2.2:5000`.
The base URL now comes from `TRUXIFY_API_BASE_URL`
(String.fromEnvironment, same convention as trip/earnings services) and
the endpoint uses the real `/api/v1/trips/events/batch` route. In
release builds an unset base URL throws instead of silently using HTTP.

Closes #5849

### DIFF
--- a/apps/driver/lib/services/sync_engine.dart
+++ b/apps/driver/lib/services/sync_engine.dart
@@ -12,6 +12,18 @@ class SyncEngine {
   static Database? _db;
   static bool _isSyncing = false;
 
+  /// Backend base URL, injected at build time via --dart-define.
+  /// Mirrors ApiClient so this service never bakes in a hardcoded
+  /// cleartext host (previously `http://10.0.2.2:5000`).
+  static String get apiBaseUrl {
+    const envUrl = String.fromEnvironment('TRUXIFY_API_BASE_URL');
+    if (envUrl.isNotEmpty) return envUrl;
+    if (kReleaseMode) {
+      throw StateError('TRUXIFY_API_BASE_URL must be set in release mode');
+    }
+    return 'http://localhost:5000';
+  }
+
   static Future<Database> get database async {
     if (_db != null) return _db!;
     _db = await _initDb();
@@ -81,9 +93,6 @@ class SyncEngine {
       if (user == null) return;
       final token = await user.getIdToken();
 
-      // We use a single arbitrary tripId for the sync endpoint, usually the backend 
-      // routes it via `/api/trips/:id/sync`. Here we use the first event's tripId.
-      final tripId = events.first['trip_id'] as String;
       final idempotencyKey = const Uuid().v4();
 
       final requestBody = {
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #5860, opened 2026-08-02
fix(ml): wire /ab-testing/predict to the real ETA model and stop leaking config

### LINKED ISSUE #5850: bug: `/ab-testing/predict` Returns a Hardcoded Constant Prediction for Every Request, and `/status` Leaks Internal A/B Configuration
## Description
`backend/ml/routes/ab_testing.py:28-49` — the A/B testing prediction endpoint never invokes a model; it returns a fixed, fabricated result for every input:
```python
@router.post("/predict")
async def predict_with_ab(request: PredictionRequest):
    routing = ab_service.get_model_for_request(request.request_id or 'unknown')
    # Here you would call the actual prediction
    # For demo, return simulated prediction
    prediction = {
        'order_id': request.order_id,
        'predicted_eta': 4.5,
        'confidence': 0.85,
        'routing': routing,
        'timestamp': datetime.utcnow().isoformat()
    }
    return prediction
```
Every request — regardless of `order_id`, features, or routing — returns `predicted_eta: 4.5` and `confidence: 0.85`. Any consumer of this endpoint (ETA displays in the customer/driver apps, or routing experiments keyed off it) shows fabricated, constant predictions, and the A/B metrics logged against it are meaningless. Because it is also mounted under the ML router alongside real features, it silently substitutes fake output for real model inference.

Additionally, `/status` (lines 89-99) exposes internal experiment configuration to any caller:
```python
return { 'status': 'active', 'active_test': active_test,
         'traffic_split': ab_service.traffic_split,
         'threshold': ab_service.threshold, ... }
```
revealing `traffic_split` and `threshold` — internal knobs that should not be public.

## Expected Behavior
`/predict` should call the actual ranked model (or be removed), and `/status` should only expose non-sensitive metadata.

## Actual Behavior
`/predict` fabricates a constant `predicted_eta: 4.5` / `confidence: 0.85` for every request, and `/status` exposes `traffic_split`/`threshold`.

## Affected Files
- `backend/ml/routes/ab_testing.py` (lines 28-49) — `/predict` hardcoded result
- `backend/ml/routes/ab_testing.py` (lines 89-99) — `/status` leaks `traffic_split` and `threshold`

## Proposed Fix
- Implement the real prediction call (or clearly mark the route as a stub and gate it behind a demo flag so it is never mounted in production).
- Restrict `/status` (and the other admin-style endpoints) to an internal key or remove `traffic_split`/`threshold` from the public response.
- Add a test asserting `/predict` output varies with inputs once the real model is wired in.

ionfwsrijan:
fix(ab-testing): /predict now returns real model predictions

`/predict` returned fake data and `/status` leaked internal config.
`/predict` now runs the trained `eta_predictor`, requires a `distance`
feature (422 otherwise), and derives confidence from the model's CI;
`/status` no longer exposes `traffic_split`/`threshold`. Regenerated the
committed `eta_predictor.pkl` (was numpy 2.x incompatible with the
pinned `numpy==1.23.5`). 3 new tests pass.

Closes #5850


ionfwsrijan:

PR #5861, opened 2026-08-02
fix(admin): build ticket API URL from configured base and send admin auth

### LINKED ISSUE #5851: bug: Admin Flutter Dashboard Can't Load Tickets — Hostless Relative URL Throws, and It Sends No Auth Headers
## Description
`apps/admin/lib/main.dart:126-147` — the admin panel's ticket list calls the backend with a **hostless relative URL**:
```dart
Future<void> _fetchTickets() async {
  try {
    // Connecting directly to the existing /api/support/admin/tickets endpoint
    final response = await http.get(Uri.parse('/api/support/admin/tickets'));
    if (response.statusCode == 200) {
      setState(() { _tickets = json.decode(response.body); ... });
    }
    ...
```
`Uri.parse('/api/support/admin/tickets')` has no scheme or host, so `http.get` throws immediately (`Invalid argument(s): No host specified in URI`) and the `catch` branch sets `_error = 'Error connecting to backend: ...'`. The admin panel can never display tickets.

Even if a base URL were added, the request sends **no `Authorization` header** — the backend's `authenticate` + `requirePolicy('ticket:admin-view-all')` guard (`backend/api/src/routes/supportRoutes.js:707`) would reject it. The backend side is correctly protected; this finding is about the exposed-but-broken admin client: it advertises an admin surface, fails at runtime, and must never be pointed at a real backend until it implements admin authentication.

## Expected Behavior
The admin app should use a configured base URL and send a real admin session/credential (token or session cookie) with each request.

## Actual Behavior
`http.get` is called with a hostless relative URL (throws at runtime), and no authentication header is sent.

## Affected Files
- `apps/admin/lib/main.dart` (line 129) — hostless `Uri.parse('/api/support/admin/tickets')`
- `apps/admin/lib/main.dart` (lines 126-147) — no `Authorization`/cookie on the request
- `backend/api/src/routes/supportRoutes.js` (line 707) — backend guard (`ticket:admin-view-all`) for comparison

## Proposed Fix
- Build the URL from a configurable base (e.g. `String.fromEnvironment('ADMIN_API_URL')`, HTTPS in release).
- Add admin authentication (token from a secure login flow or a session cookie) to the request headers.
- Add an integration test that the tickets screen renders only with valid admin credentials.

ionfwsrijan:
fix(admin): make admin dashboard API base URL and token configurable

The admin dashboard hardcoded its API base URL. Added `AdminApiConfig`
reading `ADMIN_API_URL` (HTTPS enforced in release) and
`ADMIN_API_TOKEN` from dart-define; tickets are fetched against an
absolute URL with an Authorization header when a token is set.

Closes #5851

### DIFF
--- a/apps/admin/lib/main.dart
+++ b/apps/admin/lib/main.dart
@@ -1,6 +1,28 @@
 import 'package:flutter/material.dart';
+import 'package:flutter/foundation.dart';
 import 'package:http/http.dart' as http;
 import 'dart:convert';
+
+/// Backend connection settings, injected at build time via --dart-define.
+class AdminApiConfig {
+  static const String _envUrl = String.fromEnvironment('ADMIN_API_URL');
+  static const String _envToken = String.fromEnvironment('ADMIN_API_TOKEN');
+
+  static String get baseUrl {
+    final url = _envUrl.isNotEmpty ? _envUrl : 'http://localhost:5000';
+    if (kReleaseMode) {
+      if (!url.startsWith('https://')) {
+        throw StateError(
+          'ADMIN_API_URL must be an HTTPS URL in release mode. '
+          'Set --dart-define=ADMIN_API_URL=https://...',
+        );
+      }
+    }
+    return url;
+  }
+
+  static String? get authToken => _envToken.isNotEmpty ? _envToken : null;
+}
 
 void main() {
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #5862, opened 2026-08-02
fix(rls): use get_profile_id() in bids and audit log policies 

### LINKED ISSUE #5852: bug: Newer RLS Policies Compare `auth.uid()` Directly Against `profiles.id` Columns, Bypassing the Canonical `get_profile_id()` Mapping
## Description
The schema's canonical identity mapping (established in `20240101000000_rls.sql:13-18`) is: the Firebase JWT `sub` (returned by `auth.jwt() ->> 'sub'`) is stored in `profiles.firebase_uid`, and all ownership policies map it to `profiles.id` through the helper `get_profile_id()`. The newer migrations skip this helper and compare `auth.uid()` directly to columns that hold `profiles.id` values:

`supabase/migrations/20260730120000_create_bids_table.sql`:
```sql
CREATE POLICY "Drivers can view their own bids"
    ON public.bids
    FOR SELECT
    USING (auth.uid() = driver_id);          -- line 26
CREATE POLICY "Drivers can insert their own bids"
    ON public.bids
    FOR INSERT
    WITH CHECK (auth.uid() = driver_id);     -- line 31
CREATE POLICY "Customers can view bids on their loads"
    ON public.bids
    FOR SELECT
    USING (EXISTS (SELECT 1 FROM public.load_offers lo
                   WHERE lo.id = bids.load_id AND lo.customer_id = auth.uid()));  -- line 48
```
The same pattern appears in the cold-chain telemetry SELECT policy (`20260721000000_add_cold_chain_telemetry.sql:16-24`) and the audit-log policy (`20260723000010_create_application_audit_logs.sql:42-52`).

For the documented Firebase auth flow, `auth.uid()` returns the JWT `sub` (the Firebase UID string), which is *not* equal to `profiles.id` (a `gen_random_uuid()` per `docs/supabase_setup.sql:65`) — and typically cannot even cast cleanly to UUID. So:
- Firebase-authenticated drivers can never select/insert their own bids, and customers can never read bids on their loads — a denial-of-access bug for the entire primary user base.
- Supabase-auth users, whose `auth.uid()` semantics differ, get a *different* authorization matrix than Firebase users for the same data — an inconsistent, audit-hostile security model (and the mismatch means the policies may grant or deny based on an identity that has nothing to do with the row's owner).

## Expected Behavior
Every ownership predicate should use the canonical `get_profile_id()` (or compare against `firebase_uid`) so the same identity semantics apply regardless of auth provider.

## Actual Behavior
Newer policies compare `auth.uid()` directly to `driver_id`/`customer_id` (which hold `profiles.id`), locking out Firebase users and creating an inconsistent authorization matrix across the app.

## Affected Files
- `supabase/migrations/20260730120000_create_bids_table.sql` (lines 23-37, 41-50) — bids policies use `auth.uid()`
- `supabase/migrations/20260721000000_add_cold_chain_telemetry.sql` (lines 16-24) — telemetry SELECT policy (see also issue-43)
- `supabase/migrations/20260723000010_create_application_audit_logs.sql` (lines 42-52) — audit-log policy (see also issue-51)
- `supabase/migrations/20240101000000_rls.sql` (lines 13-18) — canonical `get_profile_id()` helper (the correct pattern)

## Proposed Fix
- Replace `auth.uid() = driver_id` / `customer_id = auth.uid()` with `driver_id = get_profile_id()` / `customer_id = get_profile_id()` in all newer policies.
- Add a policy test that exercises the Firebase flow (JWT with a `sub` that matches `profiles.firebase_uid`) and asserts the driver/customer can read and write their own bids.
- Audit all migrations after `20240101000000` for any policy that compares `auth.uid()` to profile-id columns.

ionfwsrijan:
fix(rls): compare profile ids, not auth.uid(), in bids and audit policies

`auth.uid()` returns the Firebase UID, while `bids.driver_id`,
`load_offers.customer_id`, and `profiles.id` store profile UUIDs, so the
policies could never match for Firebase-auth users. Replaced every
`auth.uid()` comparison in the bids policies
(20260730120000_create_bids_table.sql) and the audit-log admin-read
policy (20260723000010_create_application_audit_logs.sql) with the
canonical `get_profile_id()` helper. Added static tests asserting these
migrations (and the cold-chain telemetry migration) never compare
`auth.uid()` to profile-id columns. 123/123 tests pass.

Closes #5852

### DIFF
--- a/backend/api/test/unit/rlsSecurity.test.js
+++ b/backend/api/test/unit/rlsSecurity.test.js
@@ -143,6 +143,32 @@ describe('Individual migration files with RLS policies', () => {
   });
 });
 
+describe('RLS ownership policies compare profile ids, not auth.uid() (issue #5852)', () => {
+  const readMigration = (name) =>
+    readFileSync(path.resolve(__dirname, `../../../../supabase/migrations/${name}`), 'utf8');
+
+  it('bids migration uses get_profile_id() for driver and customer ownership policies', () => {
+    const content = readMigration('20260730120000_create_bids_table.sql');
+    expect(content).toMatch(/USING \(get_profile_id\(\) = driver_id\)/);
+    expect(content).toMatch(/WITH CHECK \(get_profile_id\(\) = driver_id\)/);
+    expect(content).toMatch(/USING \(get_profile_id\(\) = driver_id AND status = 'pending'\)/);
+    expect(content).toMatch(/lo\.customer_id = get_profile_id\(\)/);
+    expect(content).not.toMatch(/auth\.uid\(\)/);
+  });
+
+  it('application_audit_logs admin-read policy resolves the profile via get_profile_id()', () => {
+    const content = readMigration('20260723000010_create_application_audit_logs.sql');
+    expect(content).toMatch(/profiles\.id = get_profile_id\(\)/);
+    expect(content).not.toMatch(/auth\.uid\(\)/);
+  });
+
+  it('cold chain telemetry policy uses get_profile_id() for load ownership checks', () => {
+    const content = readMigration('20260721000000_add_cold_chain_telemetry.sql');
+    expect(content).toMatch(/load_offers\.customer_id = get_profile_id\(\) OR load_offers\.driver_id = get_profile_id\(\)/);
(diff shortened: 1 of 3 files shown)


ionfwsrijan:

PR #5917, opened 2026-08-03
fix(escrow): make MEVProtectedEscrow commit-reveal-release flow reachable

### LINKED ISSUE #5889: bug: MEVProtectedEscrow.revealCommitment always reverts - createCommitment pre-marks usedCommits, so escrow release is unreachable
## Overview
`MEVProtectedEscrow.revealCommitment` can never succeed because `createCommitment` pre-marks the very hash the reveal step requires to be unmarked. The commit-reveal escrow flow is therefore unreachable.

## Evidence
- `blockchain/contracts/MEVProtectedEscrow.sol:67-74` — `createCommitment(bytes32 secretHash)` sets `userCommitments[msg.sender] = secretHash` and `usedCommits[secretHash] = true` (line 71).
- `blockchain/contracts/MEVProtectedEscrow.sol:76-84` — `revealCommitment(bytes32 secret)` computes `commitHash = keccak256(abi.encodePacked(secret, msg.sender))` (line 77) and requires `userCommitments[msg.sender] == commitHash` (line 78).
- For line 78 to hold, the caller must have committed `commitHash` as their `secretHash` in `createCommitment`; but line 71 already set `usedCommits[commitHash] = true`, so the next require `!usedCommits[commitHash]` (line 79) always fails and the function always reverts with `Already revealed`.
- The rest of the flow depends on it: `createEscrow` (lines 94-95) requires the committed hash and `releaseEscrowWithProof` requires the revealed secret, so no escrow can ever complete its commit-reveal.

## Impact
- The commit-reveal mechanism (and any MEV-protected escrow that depends on revealing a secret) is permanently broken; funds deposited into escrows can never be released via the intended path.

## Suggested fix
- Use a distinct flag for the reveal step, or do not mark `usedCommits` during `createCommitment`; e.g., track revealed status separately (`mapping(bytes32 => bool) public revealed`) and check that instead.
- Add unit tests covering commit -> reveal -> release to prevent regressions.

ionfwsrijan:
## Problem

`MEVProtectedEscrow.revealCommitment` can never succeed. `createCommitment` stores the caller's `secretHash` in `userCommitments` and pre-marks `usedCommits[secretHash] = true`. The reveal step requires `userCommitments[msg.sender] == commitHash` **and** `!usedCommits[commitHash]`, but those two requirements are mutually exclusive: for the first to hold the caller must have committed `commitHash`, which already sets `usedCommits[commitHash] = true`, so the second always fails with `Already revealed`. Every MEV-protected escrow deposit therefore can never complete its commit-reveal and release path.

Additionally, `_verifyMEVProof` compared `block.number` against `createdAt` (a `block.timestamp` value), so `releaseEscrowWithProof` would always revert with `Too early` even after the reveal, leaving funds permanently stuck.

## Fix

- Add a dedicated `revealedCommits` mapping to track the reveal step separately from the create-time `usedCommits` marker.
- `revealCommitment` now guards against re-reveals with `!revealedCommits[commitHash]` instead of `!usedCommits[commitHash]`, so a committed escrow can actually be revealed.
- Fix the anti-back-running timing guard in `_verifyMEVProof` to compare timestamps (`block.timestamp > createdAt + flashbotsProtection`) instead of mixing block numbers with timestamps.
- Add unit tests covering commit -> reveal -> release, double-reveal rejection, wrong-secret rejection, and release-without-reveal rejection.

## Files changed

- `blockchain/contracts/MEVProtectedEscrow.sol`
- `blockchain/test/MEVProtectedEscrow.test.js`

## Testing

Added `blockchain/test/MEVProtectedEscrow.test.js` exercising:
- Full commit -> reveal -> release flow (asserts escrow `released`/`revealed` state and driver payout).
- Double reveal of the same commitment reverts with `Already revealed`.
- Revealing with the wrong secret reverts with `Invalid commit` and does not mark the commitment revealed.
- Releasing an escrow whose secret was never revealed reverts with `Invalid secret`.

Closes #5889

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved commitment tracking to prevent duplicate revelations.
  - Updated MEV protection timing to use seconds and escrow creation timestamps, providing more accurate validation during the release window.
  - Strengthened handling of invalid secrets and unrevealed commitments.

- **Tests**
  - Added coverage for successful commit–reveal–release flows, duplicate reveals, invalid secrets, timing windows, balance updates, and expected transaction failures.

### DIFF
--- a/blockchain/contracts/MEVProtectedEscrow.sol
+++ b/blockchain/contracts/MEVProtectedEscrow.sol
@@ -39,11 +39,12 @@ contract MEVProtectedEscrow is Ownable, ReentrancyGuard, Pausable {
     mapping(uint256 => Escrow) public escrows;
     mapping(address => bytes32) public userCommitments;
     mapping(bytes32 => bool) public usedCommits;
+    mapping(bytes32 => bool) public revealedCommits;
     mapping(address => uint256) public userNonces;
 
     uint256 public escrowCounter;
     uint256 public commitRevealPeriod = 10; // blocks
-    uint256 public flashbotsProtection = 1; // blocks
+    uint256 public flashbotsProtection = 1; // seconds (anti-back-running window)
 
     // MEV Protection Parameters
     uint256 public minPriorityFee = 1 gwei;
@@ -76,9 +77,9 @@ contract MEVProtectedEscrow is Ownable, ReentrancyGuard, Pausable {
     function revealCommitment(bytes32 secret) external whenNotPaused {
         bytes32 commitHash = keccak256(abi.encodePacked(secret, msg.sender));
         require(userCommitments[msg.sender] == commitHash, "Invalid commit");
-        require(!usedCommits[commitHash], "Already revealed");
+        require(!revealedCommits[commitHash], "Already revealed");
 
-        usedCommits[commitHash] = true;
+        revealedCommits[commitHash] = true;
         
         emit CommitmentRevealed(msg.sender, secret);
     }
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

Issue #5890, opened 2026-08-03
sec: SECURITY DEFINER RPCs update_order_and_load_offer and create_order_tx lack caller verification - any authenticated user can rewrite any order


ionfwsrijan:
## Overview
Two SECURITY DEFINER RPC functions created after the RPC-hardening migrations re-introduce the privilege-escalation class that #1851 addressed: any authenticated Supabase user can call them directly via the REST API to rewrite or forge order data, bypassing all RLS.

## Evidence
- `supabase/migrations/20260728000000_create_update_order_rpc.sql` — `update_order_and_load_offer(p_order_id, p_order_display_id, p_order_updates, p_offer_updates)` is `SECURITY DEFINER` with NO `auth.uid()`/`auth.role()` check and NO `SET search_path`. It updates `orders` (including `total_amount`, `base_freight`, `platform_fee`, `toll_estimate`, `drop_address`) by any `p_order_id`, and the matching `load_offers`, with no ownership check.
- `supabase/migrations/20260723000003_create_order_tx.sql` — `create_order_tx` is `SECURITY DEFINER` with NO caller verification; it inserts an `orders` + `load_offers` row attributed to an arbitrary `p_customer_id`.
- These were created 2026-07-23 and 2026-07-28, AFTER the hardening that fixed the same class: `20260704000000_add_auth_verification_to_rpc_functions.sql` (auth checks for accept_bid_tx/withdraw_funds_tx/submit_rating_tx), `20260706075009_secure_rpc_search_path.sql`, and `20260708000000_fix_rpc_security.sql` (which explicitly restored `auth.uid()` and added `auth.role() <> 'service_role'` to the claim RPCs).
- Functions are callable by authenticated users via the Supabase PostgREST `/rpc/` endpoint; `revoke_anon_privileges.sql` only revoked TABLE privileges from `anon`, not function EXECUTE.

## Impact
- Any authenticated user can modify any order's price and route details (`update_order_and_load_offer`), which cascades into escrow bookings, payouts, and change-drop repricing.
- Any authenticated user can forge orders attributed to any customer (`create_order_tx`).

## Suggested fix
- Add `auth.uid()` ownership verification (and `SET search_path = public, pg_temp`) to both functions.
- Restrict them to `service_role` only, or drop them and route all order mutation through the Node.js backend.
- Grep all SECURITY DEFINER functions for missing `auth.uid()`/`auth.role()` checks as a release gate.

Issue #5722, opened 2026-08-02
bug:  `update_order_and_load_offer` Is an Unauthenticated SECURITY DEFINER RPC: Any User Can Rewrite Orders, Prices, and Payout Math


ionfwsrijan:
## Description
The RPC `update_order_and_load_offer` (added 2026-07-28) is defined `SECURITY DEFINER` — it runs as the superuser, bypassing RLS — and contains **no `auth.uid()` check, no ownership predicate, and no `SET search_path`**. It executes `UPDATE orders SET drop_address, drop_lat/drop_lng, base_freight, toll_estimate, platform_fee, total_amount ... WHERE id = p_order_id` and mutates the linked `load_offers.freight_value` row. Functions in `public` default to `EXECUTE` for `PUBLIC`, so any authenticated user can call it directly via `POST /rest/v1/rpc/update_order_and_load_offer` with any order UUID.

This is the exact vulnerability class the team fixed for `claim_refund_reconciliation` / `claim_release_reconciliation` in `20260708000000_fix_rpc_security.sql` (see its comment at lines 3-8 admitting the REST API is reachable by "any authenticated user") — the new RPC **reintroduced** it.

## Expected Behavior
Only the order's own customer may change the drop location (and derived price fields), after server-side guards (`orderRoutes.js:1061 assertCustomerOwnership`, `orderValidationService.js:160 assertChangeDropAllowed` — which blocks changes once escrow is `funded` or status ≠ `pending`) and only with server-computed pricing.

## Actual Behavior
Any authenticated user can:
1. Redirect another customer's shipment by rewriting `drop_address/drop_lat/drop_lng` on any order
2. Mutate `base_freight/toll_estimate/platform_fee/total_amount` and `load_offers.freight_value` arbitrarily — inflating payout math (`complete_trip_tx` line 102 uses `COALESCE(bid_amount, total_amount, 0)`)
3. Do so even on orders already `truck_assigned`/`funded`, bypassing `assertChangeDropAllowed` entirely

## Affected Files
- `supabase/migrations/20260728000000_create_update_order_rpc.sql` (lines 1-44) — `SECURITY DEFINER` line 9; unrestricted `UPDATE orders` lines 15-26, `UPDATE load_offers` lines 29-40
- `backend/api/src/routes/orderRoutes.js` (line 1112) — RPC call; ownership check at 1061 bypassed
- `backend/api/src/services/order/orderLifecycleService.js` (line 579) — RPC call
[...]

PR #5926, opened 2026-08-03
fix(rpc): restore caller verification + search_path on complete_trip_tx

ionfwsrijan:
## Summary

Issue #5890 — grep all `SECURITY DEFINER` functions for missing `auth.uid()`/`auth.role()` checks and `search_path` as a release gate.

A full audit of every `SECURITY DEFINER` function in the migration chain found that the two RPCs named in the issue were **already hardened** by previously-merged migrations:

- `update_order_and_load_offer` → hardened by `20260802000000_secure_update_order_and_load_offer.sql` (#5722): ownership guard via `get_profile_id()`, `SET search_path = public, pg_temp`, `REVOKE EXECUTE FROM anon, authenticated`.
- `create_order_tx` → hardened (issue #5823): `service_role` may supply `p_customer_id`, all other callers bound to their own JWT-derived profile via `get_profile_id()`, `SET search_path = public, pg_temp`.

The audit did surface **one genuine regression**: the latest `complete_trip_tx` definition (`20260803100000_update_complete_trip_tx.sql`) dropped both the caller verification and the search path that prior versions (`20260704000001`, `20260706075009`) had established. It is `SECURITY DEFINER`, pays the driver wallet and flips escrow to `released`, and was callable by **any** anon/authenticated user directly via the PostgREST RPC endpoint with no ownership check.

## Changes

`supabase/migrations/20260803100000_update_complete_trip_tx.sql`

1. Added `SET search_path = public, pg_temp` to the function header (prevents search-path hijacking).
2. Restored the caller-verification guard, using the canonical comparison from #5722:
   ```sql
   if auth.uid() is not null and get_profile_id() <> v_order.driver_id then
     raise exception 'Unauthorized: you can only complete trips you are assigned to';
   end if;
   ```
   `auth.uid()` is the Firebase UID and never matches the `profiles.id` stored in `orders.driver_id`; `get_profile_id()` maps the JWT sub to `profiles.id`, so the assigned driver passes, `service_role` (backend) passes, and any other authenticated user is rejected.
3. Added `REVOKE EXECUTE ON FUNCTION complete_trip_tx(uuid, uuid, text) FROM anon;` as defense-in-depth. `authenticated` is intentionally left executable because the driver-facing backend invokes this RPC with the user's authenticated client (`orderMilestoneService.js`, `deliveryVerificationService.js`).

## Verification

- `npx vitest run test/unit/rlsSecurity.test.js test/unit/verifyDbSchema.test.js` → 140 tests passed (RLS guard assertions, search-path assertions, complete_trip_tx structure tests all green).
- No backend call sites changed; both callers pass the authenticated driver client, which the guard permits.

## Related

- Fixes the vulnerable function found during the issue #5890 release-gate audit.
- Follows the exact hardening pattern established in #5722.

### DIFF
--- a/supabase/migrations/20260803100000_update_complete_trip_tx.sql
+++ b/supabase/migrations/20260803100000_update_complete_trip_tx.sql
@@ -10,6 +10,7 @@ create or replace function complete_trip_tx(
 returns table(driver_id uuid)
 language plpgsql
 security definer
+set search_path = public, pg_temp
 as $$
 declare
   v_order record;
@@ -23,6 +24,13 @@ begin
 
   if not found then
     raise exception 'Order not found';
+  end if;
+
+  -- Verify the caller is the driver assigned to this order. get_profile_id()
+  -- maps the Firebase JWT sub to profiles.id, which is what orders.driver_id
+  -- stores (auth.uid() is the Firebase UID and would never match).
+  if auth.uid() is not null and get_profile_id() <> v_order.driver_id then
+    raise exception 'Unauthorized: you can only complete trips you are assigned to';
   end if;
 
   if v_order.driver_id is null then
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

KanishJebaMathewM (approved):
Looks good

Issue #6007, opened 2026-08-03
zkEVMBridge.withdrawFromL2 accepts any non-empty proof: no verification, no accounting, bridge can be drained


ionfwsrijan:
# `zkEVMBridge.withdrawFromL2` accepts any non-empty proof — no proof verification, no withdrawal accounting

**Severity:** Critical
**Component:** Smart Contract — `blockchain/contracts/zkEVMBridge.sol`
**Affected files:**
- `blockchain/contracts/zkEVMBridge.sol` (lines 36–46, `withdrawFromL2`)

## Description

`withdrawFromL2` is a **public** function (no owner/operator restriction) intended to settle L1 funds when an L2 withdrawal is proven. It does no verification of the proof and no accounting check:

```solidity
// zkEVMBridge.sol:36-46
function withdrawFromL2(bytes memory proof, address recipient) external {
    require(proof.length > 0, "Proof is empty");
    address payable recipientAddr = payable(recipient);
    _transferETH(recipientAddr, bridgeBalance / 100);   // 1% of the bridge
    emit L2Withdrawal(msg.sender, recipient, bridgeBalance / 100, proof);
}
```

`proof.length > 0` is the **only** check. Anyone can:
1. Call `withdrawFromL2` with any non-empty byte string and any recipient address.
2. Receive `bridgeBalance / 100` (1% of every ETH sitting in the bridge) per call.
3. Repeat arbitrarily many times — there is no per-caller cap, no nonce, no mapping of claimed withdrawals, and no check that the referenced L2 withdrawal actually happened or was credited to the recipient.

Meanwhile the "real" withdrawal path is `claimWithdrawal` (lines 48+), which honors a per-user `pendingWithdrawals[msg.sender]` balance — proving the contract knows how to do accounting. `withdrawFromL2` bypasses that entirely.

If the bridge ever holds meaningful ETH (relayer-funded escrow, `depositToL2` deposits), repeated calls drain the contract down to ~0 ETH, at which point legitimate L2 withdrawals become unpayable (insolvency). This is a loss-of-funds / insolvency bug on the custody contract.

## Expected Behavior

`withdrawFromL2` (or its equivalent) verifies the zk proof against the L2 rollup's public inputs (e.g. via a verifier contract/`IZkEVM` interface that exposes proof verification) and credits only the recipient named in the proven withdrawal, at most once, up to the claimed amount.

## Actual Behavior

[...]

PR #6019, opened 2026-08-03
fix: add proof replay protection and zero-amount guard to withdrawFromL2 (#6007)

Aryanbansal-05:
## Description
withdrawFromL2 had no proof replay protection, meaning the same proof could be submitted multiple times to repeatedly claim the same L2 withdrawal. Added a usedProofs mapping keyed by keccak256(proof) that marks each proof as consumed on first use and reverts on any subsequent attempt. Also added an amount > 0 guard to prevent zero-value drain calls. The existing depositedAmount accounting already prevents withdrawing more than deposited, so combined with these fixes the bridge can no longer be drained via proof replay or zero-amount abuse.

## Type of Change

## How Has This Been Tested?

- **Test Commands:**
npx hardhat test
1. Deposit ETH to the bridge
2. Call withdrawFromL2 with a valid proof — should succeed
3. Call withdrawFromL2 again with the same proof — should revert with Proof already used
4. Call withdrawFromL2 with amount = 0 — should revert with Amount must be > 0
5. Call withdrawFromL2 with amount > depositedAmount — should revert with Exceeds deposited amount
6. Call claimWithdrawal — should transfer correct amount to caller

- **Expected Result:**
Each proof can only be used once. Zero-amount calls are rejected. Withdrawal amount is always bounded by what the caller deposited. Bridge balance cannot be drained via replay or zero-value abuse.

### Test Environment
- [x] Local manual testing

## Related Issues

Closes 6007

## PR Checklist

### DIFF
--- a/blockchain/contracts/zkEVMBridge.sol
+++ b/blockchain/contracts/zkEVMBridge.sol
@@ -36,19 +36,28 @@ contract zkEVMBridge is Ownable, ReentrancyGuard {
         emit BridgeDeposit(msg.sender, amount, bridgeFee);
     }
 
-    function withdrawFromL2(
-        uint256 amount,
-        bytes calldata proof
-    ) external nonReentrant {
-        require(proof.length > 0, "Empty proof");
-        require(depositedAmount[msg.sender] >= amount, "Exceeds deposited amount");
-        // Withdraw from L2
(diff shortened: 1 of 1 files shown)


Aryanbansal-05:
GSSOC

PR #6022, opened 2026-08-03
fix: run WASM edge functions in worker thread with hard timeout to prevent event loop freeze (#6006)

### LINKED ISSUE #6006: Edge-function execution runs synchronous WASM with no timeout: one infinite-loop edge function freezes the Node event loop
# Edge-function execution runs synchronous WASM with no timeout — one malicious edge function freezes the entire Node event loop

**Severity:** Critical
**Component:** Backend API — edge-runtime WASM sandbox
**Affected files:**
- `wasm/edge-runtime.js` (lines 62–91, `executeEdgeFunction`; lines 93–107, `executeWithTimeout`)

## Description

`wasm/edge-runtime.js` ships a `executeWithTimeout` helper (lines 93–107) but the main entry point, `executeEdgeFunction`, **never uses it**:

```js
// wasm/edge-runtime.js:62-91
async function executeEdgeFunction(wasmBytes, params, context = {}) {
  const module = new WebAssembly.Module(wasmBytes);
  const instance = new WebAssembly.Instance(module, imports);
  const inputJson = JSON.stringify(params);
  const inputPtr = alloc(instance, inputJson.length + 1);
  writeString(instance, inputPtr, inputJson);
  const startPtr = instance.exports.start;
  if (typeof startPtr !== 'function') throw new Error('Invalid edge function: missing start export');
  const resultPtr = startPtr(inputPtr);   // synchronous WASM call
  ...
}

async function executeWithTimeout(fn, timeoutMs = 1000) {
  return Promise.race([fn(), new Promise((_, reject) => setTimeout(() => reject(new Error('Execution timed out')), timeoutMs))]);
}
```

The WASM export `start` is invoked synchronously. WebAssembly calls run directly on the main JS thread with **no interrupt mechanism** — `Promise.race`/`setTimeout` cannot fire while a synchronous WASM call is running, so `executeWithTimeout` is useless for sync calls even if it were wired in. A single edge function with an infinite loop (`loop { br 0 }`) or a long-running computation blocks the **entire Node event loop** for as long as it runs — permanently, in the infinite-loop case.

Where is this called? The edge-function HTTP API (`/api/wasm/execute`, `wasmRoutes.js`) and any workflow/event hook that runs customer-supplied edge functions. This is a user-supplied-code execution surface: any customer can deploy an infinite-loop edge function and take down the whole API process (no worker isolation, no per-request timeout, no memory cap).

## Expected Behavior

Edge functions run in a separate worker thread/process or WebAssembly that supports interruption, with a hard timeout (default e.g. 1s), memory limit and CPU limit. A hanging edge function only fails its own request, never the API process.

## Actual Behavior

A synchronous WASM `start` export that never returns freezes the Node event loop indefinitely — all HTTP requests, WebSockets and timers stop responding. The whole API is unavailable until the process is force-restarted.

## Proposed Fix

1. Execute edge functions in a dedicated Worker thread (`worker_threads`) or child process with `resourceLimits` (e.g. `maxOldGenerationSizeMb: 32`, `stackSizeMb: 2`) so a hang kills only the worker, not the process.
2. Run the worker with an external hard timeout (kill the worker on expiry), since sync WASM cannot be interrupted from the same thread.
3. Pre-scan WASM modules for infinite-loop patterns where feasible and/or impose an instruction-count budget.
4. Reject functions that do not return within the timeout and return `504` to the caller.
5. Add a test that feeds a `loop{}` WASM module and asserts the request fails within the timeout while the server stays responsive.

Aryanbansal-05:
## Description
executeEdgeFunction was invoking synchronous WASM exports directly on the main Node.js thread with no isolation. Since synchronous WASM calls cannot be interrupted by Promise.race or setTimeout from the same thread, a single infinite-loop edge function would freeze the entire Node event loop permanently, taking down all HTTP requests, WebSockets and timers until the process was force-restarted. Fixed by moving WASM execution into a dedicated worker_threads Worker with resourceLimits (maxOldGenerationSizeMb: 64, maxYoungGenerationSizeMb: 16, stackSizeMb: 2). The main thread sets a hard setTimeout that calls worker.terminate() if the worker does not respond within this.timeoutLimit (5s). An infinite-loop or hung WASM call now only kills its own worker thread, never the main event loop, and the caller receives a timeout error response instead of a hung request.

## Type of Change

## How Has This Been Tested?

- **Test Commands:**
npm test
1. Deploy an edge function that contains an infinite loop
2. Call /api/wasm/execute with the infinite-loop function
3. Verify the request fails with a timeout error within 5 seconds
4. Verify the server remains responsive to other requests during and after the timeout
5. Call a valid edge function — should succeed normally

- **Expected Result:**
Infinite-loop edge functions time out within this.timeoutLimit (5s) and return a 504-style error. The main Node event loop stays responsive throughout. Valid edge functions execute and return results normally.

### Test Environment
- [x] Local manual testing

## Related Issues

Closes #6006

## PR Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of WebAssembly function execution.
  * Added safeguards to stop long-running executions and return clear failure results when errors occur.
  * Isolated executions to help prevent failures from affecting other runtime operations.

### DIFF
--- a/wasm/edge-runtime.js
+++ b/wasm/edge-runtime.js
@@ -12,13 +12,13 @@ class EdgeRuntime {
         this.isInitialized = false;
         this.memoryLimit = 128 * 1024 * 1024; // 128MB
         this.timeoutLimit = 5000; // 5 seconds
-        
+
         logger.info('✅ Edge Runtime initialized');
     }
 
     async initialize() {
         if (this.isInitialized) return;
-        
+
         try {
             const wasmPath = process.env.WASM_MODULE_PATH || './wasm/truxify_wasm.wasm';
             if (fs.existsSync(wasmPath)) {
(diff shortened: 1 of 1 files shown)


Aryanbansal-05:
GSSOC

PR #6024, opened 2026-08-03
fix(api-client): inject X-Idempotency-Key on all POST/PUT/PATCH requests

### LINKED ISSUE #5999: Customer app never sends X-Idempotency-Key: every order mutation returns HTTP 400
# Customer app never sends `X-Idempotency-Key` — every order mutation returns HTTP 400 and the booking pipeline is dead

**Severity:** Critical
**Component:** Customer Flutter app + Backend API (cross-layer contract mismatch)
**Affected files:**
- `backend/api/src/middleware/idempotency.js` (lines 59–70, `requireIdempotency`)
- `backend/api/src/routes/orderRoutes.js` (lines 293, 881, 972, 1094, 1345)
- `packages/truxify_shared/lib/src/services/api_client.dart` (headers, `post`/`put`/`patch`)
- `apps/customer/lib/services/order_service.dart` (lines 64–65, 100, 121, 359)

## Description

The backend enforces an idempotency header on all order-mutating routes via `requireIdempotency` middleware:

```js
// idempotency.js:63-70
const idempotencyKey = req.headers['x-idempotency-key'];
if (!idempotencyKey) {
  if (process.env.NODE_ENV === 'test') {
    return next();
  }
  return res.status(400).json({ error: 'X-Idempotency-Key header is required for this action.' });
}
```

It is applied to every critical mutation:
- `POST /api/orders` (line 293 — order creation)
- `POST /api/orders/:id/bids/:bidId/accept` (line 881 — bid acceptance / escrow deposit)
- `POST /api/orders/:id/verify-delivery` (lines 972, 1094 — delivery verification & payout)
- `POST /api/orders/:id/cancel` (line 1345 — cancellation & escrow refund)

The shared Flutter `ApiClient` (`packages/truxify_shared/lib/src/services/api_client.dart`) sends only `Content-Type`, `Accept-Language` and `Authorization` headers (line 134 `_headers`), and `order_service.dart` posts to `/api/orders` (lines 64–65) with no idempotency header. A case-insensitive search for `idempot` across `apps/customer/lib` and `packages/truxify_shared/lib` finds zero matches (the only hit is an unrelated local JSON field `idempotencyKey` inside the offline `sync_engine.dart`).

In any non-`test` backend deployment (i.e. production/staging), **every** order creation, bid acceptance, delivery verification and cancellation from the customer app is rejected with `400 X-Idempotency-Key header is required`. The customer booking pipeline is broken before step one; the same failure hits the driver flow via the shared client for cancel/verify-delivery.

## Expected Behavior

The customer app generates a per-operation idempotency key (UUID) and sends it as `X-Idempotency-Key` on every POST/PUT/PATCH mutation, so orders are created, bids accepted, deliveries verified and cancellations processed with at-most-once semantics.

## Actual Behavior

Requests are sent without the header and the backend returns `400` for every guarded mutation. No order can be created, no bid accepted, no delivery verified, and no order cancelled from the customer app in production.

## Proposed Fix

1. Add idempotency-key generation to `ApiClient` (generate a `Uuid` per logical operation, reused on retries, new for new operations) and send it as `X-Idempotency-Key` on all mutating requests (or via an explicit opt-in per call).
2. Update `order_service.dart` (create/accept/cancel/verify) and any other mutation call sites to supply/retain the key across automatic retries.
3. Optionally relax the backend to only require the header where the operation is genuinely non-idempotent-safe, and document the contract in `docs/API_REFERENCE.md`.
4. Add a Flutter integration test asserting a mutation includes the header, and a backend test asserting a missing header still fails closed.

Nancy-3012:
```
## Fix: Customer app never sent X-Idempotency-Key header → HTTP 400 on all order mutations

Closes #5999

### Root cause
The backend `requireIdempotency` middleware enforces the `X-Idempotency-Key` header on
every order mutation (`POST /api/orders`, `POST /api/orders/:id/cancel`,
`PUT /api/orders/:id/change-drop`, `POST /api/orders/:id/ratings`).
`ApiClient` in `truxify_shared` never injected this header, so every customer order
action returned **HTTP 400** immediately.

### Fix
**`packages/truxify_shared/lib/src/services/api_client.dart`**
- Added `package:uuid` import (`Uuid` instance is `const` — one allocation per isolate).
- `post()`, `put()`, `patch()`, and `postMultipart()` each generate a UUID v4 key **before**
  calling `_execute`, so both the initial request and the automatic 401-retry path send the
  **same** key — preserving at-most-once semantics without any changes to call sites.
- All four methods expose an optional `idempotencyKey` parameter so callers can supply a
  stable key across explicit user-initiated retries.

**`packages/truxify_shared/pubspec.yaml`**
- Added `uuid: ^4.5.1`.

### What is not changed
`apps/customer/lib/services/order_service.dart` requires **zero changes** — the key is
injected transparently at the HTTP layer. Existing call sites for `createOrder`,
`cancelOrder`, `changeDrop`, and `submitRating` all work without modification.

### Testing
- `X-Idempotency-Key` header is present on every POST/PUT/PATCH request.
- Each call generates a unique UUID; a 401-triggered retry reuses the same UUID.
- GET and DELETE requests are unaffected (no header injected).
```

### DIFF
--- a/packages/truxify_shared/lib/src/services/api_client.dart
+++ b/packages/truxify_shared/lib/src/services/api_client.dart
@@ -6,6 +6,7 @@ import 'package:firebase_auth/firebase_auth.dart';
 import 'package:flutter/foundation.dart';
 import 'package:http/http.dart' as http;
 import 'package:supabase_flutter/supabase_flutter.dart';
+import 'package:uuid/uuid.dart';
 
 import '../config/app_config.dart';
 import 'http_client_factory.dart';
@@ -47,6 +48,11 @@ class MultipartFileInfo {
 ///   - Injects `Authorization: Bearer <accessToken>` into every request.
 ///   - On HTTP 401, attempts `supabase.auth.refreshSession()` once and retries.
 ///   - If the retry still returns 401, throws [ApiAuthException].
+///   - Automatically injects a stable `X-Idempotency-Key` (UUID v4) on every
+///     POST, PUT, and PATCH request so the backend `requireIdempotency`
+///     middleware never rejects customer mutations with HTTP 400.
+///   - The key is generated *before* `_execute` so the automatic 401-retry
+///     sends the exact same key, preserving at-most-once semantics.
 ///   - Logs failures in debug mode via `dart:developer`.
 ///
 /// Usage:
@@ -72,6 +78,9 @@ class ApiClient {
   final bool _isClientOwned;
   final String _baseUrl;
   final Duration _timeout;
+
+  /// Shared UUID generator — const so it is instantiated once per isolate.
+  static const _uuid = Uuid();
 
   void close() {
     if (_isClientOwned) {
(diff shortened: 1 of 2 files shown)


Nancy-3012:
gssoc

PR #6028, opened 2026-08-03
fix(api): make acquireLock fail closed when Redis is unavailable

### LINKED ISSUE #6003: Redis distributed locks fail open: lock acquisition/release silently skipped when Redis is down
# Redis distributed locks fail open — lock acquisition and release silently skip when Redis is down

**Severity:** Critical
**Component:** Backend API — concurrency / correctness
**Affected files:**
- `backend/api/src/lib/redisLock.js` (lines 25–40, 47–56)

## Description

`acquireLock` is designed to guard double-payment and double-escrow-release critical sections, but it **fails open**:

```js
// redisLock.js:25-40
async function acquireLock(key, ttl = 30000) {
  if (!redisClient) {
    return null;
  }
  try {
    const result = await redisClient.set(key, 'locked', 'PX', ttl, 'NX');
    return result === 'OK' ? { key, ttl, lockedAt: Date.now() } : null;
  } catch (err) {
    logger.warn(...);
    return null;
  }
}

// redisLock.js:47-56
async function releaseLock(lock) {
  if (!lock || !lock.key) return;
  if (!redisClient) {
    return false;
  }
  try {
    await redisClient.del(lock.key);
    return true;
  } catch (err) {
    logger.warn(...);
    return false;
  }
}
```

`null` (no lock) is the "continue without locking" signal used by callers such as `paymentRoutes.js` (`/payments/lock`, `/payments/charge-and-lock`), `deliveryVerificationService.verifyDelivery`, and `escrow.js`. When the Redis client is uninitialized **or** any `set ... NX` error occurs (Redis down, timeout, maxmemory, auth failure), the lock is silently skipped and the critical section proceeds unlocked.

Consequence: under any Redis failure (restart, network blip, memory pressure — the Redis container also has a tiny fixed memory cap in the compose file) the protections that prevent charging a payment twice and releasing escrow twice are completely disabled. Two concurrent `verify-delivery` calls can both pass the `escrow_status` check and both attempt `releaseEscrow` — releasing an already-released escrow.

This is the opposite of what a lock must do. Locks must fail **closed** (block the operation or return a 5xx) so that a transient Redis outage cannot turn into a double payout.

## Expected Behavior

When Redis is unavailable, lock-acquiring operations refuse to proceed (return `503`/abort) rather than executing the guarded critical section without a lock. If the caller can tolerate no-lock, it should do so **explicitly**, not by default.

## Actual Behavior

With Redis down or any `SET NX` error, `acquireLock` returns `null`, callers treat it as "proceed", and double-charge / double-escrow-release races are possible exactly when the system is degraded.

## Proposed Fix

1. Change `acquireLock` to throw (or return a sentinel that callers treat as hard failure) when `redisClient` is missing or a Redis error occurs; wrap callers with `try/catch` returning `503 Service Unavailable`.
2. Keep `null`-as-unlocked only for genuinely optional locks and make that opt-in via a flag (e.g. `acquireLock(key, ttl, { strict: true })` defaulting to strict).
3. Add `SET NX PX` fencing/owner token (unique per attempt, checked on release via Lua) so a stale holder cannot release a newer holder's lock.
4. Add a unit test: mock Redis failure and assert the guarded mutation is rejected, not performed.

ionfwsrijan:
## Problem

`acquireLock` in `backend/api/src/lib/redisLock.js` returned `null` when Redis was unavailable or when the `SET NX` call threw. Callers interpret `null` as "the lock is held by another process" and skip the protected operation or return `409` — so when Redis goes down the system **fails open**: reconciliation workers skip healing work and lock-protected operations proceed without mutual exclusion. The `LockAcquisitionError` class already existed but was never thrown.

## Fix

- `acquireLock` now throws `LockAcquisitionError` when `redisClient` is unavailable or the Redis call errors (fail closed). It still returns `null` only for genuine contention (`SET NX` returns nothing when another process holds the lock), preserving the existing `409`/"already being processed" semantics.
- Guarded the one call site that invoked `acquireLock` outside a `try` block (`orderRoutes.js` `/confirm-deposit`), returning `503` when the lock service is unavailable. All other callers already propagate the error through their existing `try/catch` to the error handler (fail-closed `500`) or log-and-skip inside their reconciliation loop guards.
- Updated `test/unit/redisLock.test.js` to assert the new fail-closed behavior (throws `LockAcquisitionError` when Redis is unavailable or `set` throws).

## Files changed

- `backend/api/src/lib/redisLock.js`
- `backend/api/src/routes/orderRoutes.js`
- `backend/api/test/unit/redisLock.test.js`

## Testing

- `npx vitest run test/unit/redisLock.test.js` — 18/18 pass.
- `node --check` on both modified source files passes.

Closes #6003

### DIFF
--- a/backend/api/src/lib/redisLock.js
+++ b/backend/api/src/lib/redisLock.js
@@ -19,12 +19,13 @@ export class LockAcquisitionError extends Error {
  * Acquires a distributed lock using Redis.
  * @param {string} resourceKey - The unique key identifying the resource to lock (e.g. `escrow_lock:123`)
  * @param {number} ttlMs - Time to live in milliseconds
- * @returns {Promise<string|null>} lockValue if acquired, null if not acquired or Redis unavailable
+ * @returns {Promise<string|null>} lockValue if acquired, null if the lock is held by another process
+ * @throws {LockAcquisitionError} when the lock cannot be acquired due to Redis being unavailable,
+ *  so callers fail closed instead of proceeding without mutual exclusion
  */
 export async function acquireLock(resourceKey, ttlMs = 10000) {
   if (!redisClient) {
-    logger.warn('[RedisLock] redisClient not available, cannot acquire lock for', resourceKey);
-    return null;
+    throw new LockAcquisitionError(resourceKey, 'redisClient is not available');
   }
 
   const lockValue = crypto.randomUUID();
@@ -36,7 +37,7 @@ export async function acquireLock(resourceKey, ttlMs = 10000) {
     return null;
   } catch (err) {
     logger.error({ err }, '[RedisLock] Error acquiring lock for key', resourceKey);
-    return null;
+    throw new LockAcquisitionError(resourceKey, err.message);
   }
 }
 
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -1392,7 +1392,15 @@ router.post('/:id/confirm-deposit', authenticate, userLimiter, requirePolicy('or
   const { txHash } = req.body;
 
   const lockKey = `escrow_lock:${orderId}`;
-  const lockValue = await acquireLock(lockKey, 120000);
+  let lockValue;
+  try {
+    lockValue = await acquireLock(lockKey, 120000);
+  } catch (err) {
+    if (err?.name === 'LockAcquisitionError') {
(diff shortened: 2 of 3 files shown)


ionfwsrijan:
Superseded by upstream PR #6023 (Nancy-3012, commit e4d5642c, "redis locks fail closed — prevent double-charge on Redis outage"), which implements the same fail-closed LockAcquisitionError behaviour. Closing this duplicate.

PR #6031, opened 2026-08-03
fix(wasm): isolate edge function execution in worker thread with timeout

### LINKED ISSUE #6006: Edge-function execution runs synchronous WASM with no timeout: one infinite-loop edge function freezes the Node event loop

ionfwsrijan:
## Problem

`wasm/edge-runtime.js` `executeEdgeFunction` (line 62) calls a synchronous WASM export directly with `func(...params)` and no timeout. A runaway export (infinite loop / expensive computation) blocks the Node.js event loop forever — the process never responds again. The existing `executeWithTimeout` helper (line 93) is dead code and cannot help: since JS is single-threaded, the `setTimeout` rejection can never fire while a synchronous WASM call is still running.

## Fix

- **Worker-thread isolation with a hard timeout**: real WASM exports now execute inside a dedicated `worker_threads.Worker` (`wasm/edge-runtime.worker.js`). The worker loads the `.wasm` binary, calls the requested export, and posts the result back. In `edge-runtime.js`, `_runInWorker` resolves the result, rejects with `Edge function execution timeout after N ms` and `worker.terminate()`s the thread once `timeoutLimit` (5s) elapses — so a stuck export can no longer block the server.
- The trusted in-repo JS fallback engine (used when no `.wasm` binary exists) still runs on the main thread.
- Removed the dead, misleading `executeWithTimeout`.
- Fixed `new WASI(...)` to pass `version: 'preview1'` (required on current Node; without it the module failed to instantiate at all).

## Files changed

- `wasm/edge-runtime.js`
- `wasm/edge-runtime.worker.js` (new)

## Testing

Smoke-tested against hand-built WASM binaries:
- Infinite-loop export → `Edge function execution timeout after 1000ms` at ~1s, event loop stays responsive.
- `add(3, 4)` export → `{ success: true, data: 7 }` round-tripped through the worker.
- Missing export → `Function nope not found`.
- `node --check` on both files passes.

Closes #6006

### DIFF
--- a/wasm/edge-runtime.js
+++ b/wasm/edge-runtime.js
@@ -1,4 +1,5 @@
 import fs from 'fs';
+import { Worker } from 'worker_threads';
 import { WASI } from 'wasi';
 import { createRequire } from 'module';
 import logger from '../backend/api/src/middleware/logger.js';
@@ -24,6 +25,7 @@ class EdgeRuntime {
             if (fs.existsSync(wasmPath)) {
                 const wasmBytes = fs.readFileSync(wasmPath);
                 const wasi = new WASI({
+                    version: 'preview1',
                     args: [],
                     env: Object.fromEntries(
                         Object.entries(process.env).filter(([k]) =>
@@ -75,6 +77,15 @@ class EdgeRuntime {
                 throw new Error(`Function ${functionName} not found`);
             }
             
+            // A synchronous WASM export cannot be aborted from the main thread:
+            // JS is single-threaded, so a runaway export blocks the event loop
+            // forever. Run it in a worker thread so it can be hard-killed once
+            // it exceeds timeoutLimit instead of taking down the process.
+            if (wasm.module) {
+                return await this._runInWorker(functionName, params);
+            }
+
+            // Native JS fallback engine (in-repo code) runs on the main thread.
             const result = func(...params);
             
             return {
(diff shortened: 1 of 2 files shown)


ionfwsrijan:
Superseded by upstream commit f66f49e2 (Aryan Bansal, "run WASM edge functions in worker thread with hard timeout to prevent event loop freeze"), which implements the same worker-thread timeout fix. Closing this duplicate.

PR #6033, opened 2026-08-03
fix(kyber): derive shared secret from ciphertext on both encapsulate and decapsulate

### LINKED ISSUE #6010: ML-KEM decapsulation derives shared secret from wrong point: kyber_core.py handshakes fail to agree on a key
# ML-KEM decapsulation derives the shared secret from the wrong point — `kyber_core.py` produces different keys than `encapsulate`, making the crypto stack unusable

**Severity:** Critical
**Component:** Backend — Post-Quantum Cryptography (`backend/kyber/kyber_core.py`)
**Affected files:**
- `backend/kyber/kyber_core.py` (`encapsulate`, `decapsulate`, `_derive_secret`, lines ~176–204)

## Description

The same ML-KEM bug that was fixed in `backend/pqc/kyber.py` (commit `85f2f0b9`, issue #4827) is **still present** in `backend/kyber/kyber_core.py`, which remains the canonical module imported by the transport services:

```python
# kyber_core.py
def encapsulate(pub, ...):
    ...
    # derived from UNCOMPRESSED (u, v)
    ss = _derive_secret(u, v)          # line ~179-180

def decapsulate(priv, ...):
    ...
    # derived from DECOMPRESSED v only — the buggy variant
    ss = _derive_secret(u_decompressed, v_decompressed)   # line ~189-193
```

`_derive_secret` computes `SHA3-256(u_bytes || v_bytes)` (or similar concatenation). Because `decapsulate` uses the **decompressed** `v` while `encapsulate` uses the **raw/uncompressed** `v`, the two sides hash **different byte strings** and therefore derive **different shared secrets**.

Consequence: every key agreement in the system fails:
- The two parties (encapsulator and decapsulator) never arrive at the same `ss`, so the AEAD layer keyed by it cannot decrypt.
- Secure transport, VDF-commitment encryption, or any message authenticated with the derived key is broken end-to-end.
- Any code path that "successfully" proceeded despite this must have been using test fixtures or shared-secret stubs, masking the failure.

(Contrast: `backend/pqc/kyber.py` already fixed this and its test asserts `ss_a == ss_b`. `kyber_core.py` has **no such round-trip test**, which is why the bug survived.)

## Expected Behavior

`encapsulate` and `decapsulate` derive the same shared secret for a given keypair; a round-trip unit test asserts `ss_enc == ss_dec`.

## Actual Behavior

`decapsulate` derives the secret from the decompressed point while `encapsulate` derives from the uncompressed point, producing mismatched secrets. Every real ML-KEM handshake fails to establish a usable session key.

## Proposed Fix

1. Align `decapsulate` with `encapsulate`: derive the shared secret from the **same** representation used on the encapsulating side (hash `(u, v)` exactly as `_derive_secret(u, v)` does in `encapsulate`), and, if needed, re-encrypt the ciphertext to validate the derivation.
2. Port the fix applied in `backend/pqc/kyber.py` (commit `85f2f0b9`) to `kyber_core.py`.
3. Add a round-trip test in `backend/kyber/tests` asserting `ss_enc == ss_dec` (and, ideally, NIST KAT vectors) so a regression fails CI.
4. Sweep for any other module that hashes a decompressed point on one side and an uncompressed point on the other.

ionfwsrijan:
## Problem

`backend/kyber/kyber_core.py` derived the shared secret from **different inputs** on the two sides of the KEM:

- `encapsulate` hashed the **uncompressed** `(u, v)`.
- `decapsulate` hashed the **decompressed** `(u_decompressed, result)`.

Compression is lossy, so those arrays differ byte-for-byte and the two halves always produced mismatched keys — `hybrid_decrypt` / `tls_key_exchange` could never recover the shared secret (verified: round-trip returned `enc == dec: False`).

## Fix

Derive the shared secret from the **compressed ciphertext** `(u_compressed, v_compressed)` in `encapsulate`, and from the raw ciphertext `(u, v)` in `decapsulate`. Both sides now hash byte-for-byte identical arrays, so the shared secret round-trips correctly. This follows the intent of the earlier `backend/pqc/kyber.py` fix (`85f2f0b9`), which made decapsulate hash the ciphertext `[u, v]` instead of `[u, v, result]` — but that fix left `encapsulate` hashing uncompressed values, so it still never round-tripped. This change aligns both sides on the ciphertext.

## Files changed

- `backend/kyber/kyber_core.py`
- `backend/kyber/tests/test_kyber.py` (new — round-trip regression tests)

## Testing

- New `backend/kyber/tests/test_kyber.py`: `KyberKEM` round trip, 5 repeated round trips, and a full `QuantumSafeKeyExchange` round trip — all pass (3 passed).
- Verified before/after: `enc == dec` was `False`, now `True`.

Closes #6010

### DIFF
--- a/backend/kyber/kyber_core.py
+++ b/backend/kyber/kyber_core.py
@@ -176,8 +176,9 @@ class KyberKEM:
         u_compressed = self._compress(u, 10)
         v_compressed = self._compress(v, 4)
         
-        # Derive shared secret
-        shared_secret = self._derive_secret(u, v)
+        # Derive shared secret from the compressed ciphertext so that
+        # decapsulate can reconstruct the exact same bytes.
+        shared_secret = self._derive_secret(u_compressed, v_compressed)
         
         ciphertext = {
             'u': u_compressed.tolist(),
@@ -208,8 +209,9 @@ class KyberKEM:
         for i in range(self.k):
             result = (result - self._poly_multiply(s[i], u_decompressed[i])) % self.q
         
-        # Derive shared secret
-        shared_secret = self._derive_secret(u_decompressed, result)
+        # Derive shared secret from the compressed ciphertext (u, v), which is
+        # byte-for-byte identical to the input used in encapsulate.
+        shared_secret = self._derive_secret(u, v)
         
         return shared_secret
 
--- a/backend/kyber/tests/test_kyber.py
+++ b/backend/kyber/tests/test_kyber.py
@@ -1,0 +1,44 @@
+"""Round-trip tests for the Kyber KEM shared-secret derivation.
+
+Regression test for the encapsulate/decapsulate shared-secret mismatch:
+encapsulate derived the secret from the uncompressed ciphertext while
+decapsulate derived it from the decompressed ciphertext, so the two halves
+of the KEM never produced the same key.
+"""
+import sys
+import os
+import numpy as np
+
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+from kyber_core import KyberKEM, QuantumSafeKeyExchange
+
+
+def test_kyber_kem_round_trip():
+    np.random.seed(42)
+    kem = KyberKEM()
+    public_key, secret_key = kem.keygen()
+    ciphertext, shared_secret = kem.encapsulate(public_key)
+    decapsulated = kem.decapsulate(ciphertext, secret_key)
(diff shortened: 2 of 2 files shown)


ionfwsrijan:
Superseded by upstream commit 8a444e2b (Aryan Bansal, "derive shared secret from compressed u,v in decapsulate to match encapsulate"), which implements the same Kyber shared-secret fix. Closing this duplicate.

PR #6080, opened 2026-08-04
fix(db): finalize the order-linked trip in complete_trip_tx

### LINKED ISSUE #6067: Migration 20260803100000 regresses trip finalization to a driver-global lookup — wrong trip marked delivered
## Description
The final migration `supabase/migrations/20260803100000_update_complete_trip_tx.sql` regresses the trip-finalization logic to a **driver-global lookup**, undoing the order-linked lookup that fixed #5756 / #5771 in `20260802060000_link_trips_to_orders.sql`.

In the final migration, lines 70-103:
```sql
select count(*) into v_active_trip_count
from trips
where driver_id = v_order.driver_id and status = 'active';

if v_active_trip_count > 1 then
  raise exception 'Multiple active trips found for driver %', v_order.driver_id;
end if;

if v_active_trip_count = 1 then
  select trip_display_id into v_trip_display_id
  from trips
  where driver_id = v_order.driver_id and status = 'active';
  ...
```

The earlier fixed version (`20260802060000_link_trips_to_orders.sql:72-77`) finalized the trip that actually served the order:
```sql
select trip_display_id into v_trip_display_id
from trips
where order_id = p_order_id
  and status = 'active'
order by created_at
limit 1;
```

Effects of the regression:
- Completing order A finalizes the driver's **only active trip regardless of which order it belongs to** — the exact wrong-trip-delivered bug from #5756.
- If the driver has more than one active trip (different orders), `complete_trip_tx` now raises `Multiple active trips found for driver`, permanently blocking delivery completion even though the specific order's trip is unambiguous.

## Expected Behaviour
`complete_trip_tx` finalizes the active trip whose `order_id` matches `p_order_id` (falling back to the oldest matching trip), never another driver trip.

## Actual Behaviour
The trip is selected purely by `driver_id` + `status = 'active'`, so the wrong trip can be marked completed/delivered, or completion fails outright when the driver has multiple active trips.

## Proposed Fix
1. Restore the order-linked selection in `20260803100000_update_complete_trip_tx.sql`:
   ```sql
   select trip_display_id into v_trip_display_id
   from trips
   where order_id = p_order_id and status = 'active'
   order by created_at
   limit 1;
   if v_trip_display_id is null then
     raise exception 'No active trip found for this order — cannot complete trip';
   end if;
   ```
2. Remove the driver-global `count(*)/select` and the `Multiple active trips` guard (it is not a valid invariant per-order).
3. Add migration tests: (a) driver with two active trips for two orders completes only the matching trip; (b) completing order A does not mark order B's trip delivered.

ionfwsrijan:
## Problem

Issue #6067: `20260803100000_update_complete_trip_tx.sql` reverted the trip finalization logic to selecting the driver's active trip by **`driver_id + status = 'active'`** instead of the canonical `order_id = p_order_id` lookup established by `20260802060000_link_trips_to_orders.sql` (#5756). Consequences:

- If a driver has multiple active trips, the function raises `Multiple active trips found` (or worse, without the count guard, completes an arbitrary trip).
- If the driver has no active trip at all, trip finalization is silently skipped while the wallet is still credited — reintroducing the "credit without trip" bug #5756 closed.

## Fix

Restored the canonical order-linked lookup:

```sql
select trip_display_id into v_trip_display_id
from trips
where order_id = p_order_id
  and status = 'active'
order by created_at
limit 1;

if v_trip_display_id is null then
  raise exception 'No active trip found for this order — cannot complete trip';
end if;
```

- The trip is always attributed to the order it served (`trips.order_id` is populated by the link migration; `idx_trips_order_id` indexes it).
- `ORDER BY created_at LIMIT 1` deterministically resolves multiple active trips instead of erroring on a legitimate driver state.
- Missing trip now fails closed (`RAISE`) instead of silently crediting the wallet.
- Removed the now-unused `v_active_trip_count` declaration.

## Files changed

- `supabase/migrations/20260803100000_update_complete_trip_tx.sql`

## Testing

- Verified the lookup matches the canonical implementation in `20260802060000_link_trips_to_orders.sql` (lines 71–81) and `20260802130000_complete_trip_tx_require_funded_escrow.sql` (lines 84–94).
- `git grep` confirms no remaining references to `v_active_trip_count`.

Closes #6067

### DIFF
--- a/supabase/migrations/20260803100000_update_complete_trip_tx.sql
+++ b/supabase/migrations/20260803100000_update_complete_trip_tx.sql
@@ -15,7 +15,6 @@ as $$
 declare
   v_order record;
   v_trip_display_id text;
-  v_active_trip_count int;
   v_updated_count int;
   v_otp_updated int;
 begin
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #6081, opened 2026-08-04
fix(api): release payment lock with owner token in finally

### LINKED ISSUE #6068: paymentRoutes.js references undeclared lockAcquired — ReferenceError on every payment lock, Redis lock never released
## Description
In `backend/api/src/routes/paymentRoutes.js`, the POST payment-lock handler declares `let lockValue = null` (line 195) and stores the Redis lock owner token there (line 200). But the `finally` block references a **never-declared variable** `lockAcquired` (line 315):

```js
} finally {
  if (lockAcquired) {              // ReferenceError: lockAcquired is not defined
    try {
      await releaseLock(lockKey);
    } catch (releaseErr) { ... }
  }
}
```

The string `lockAcquired` appears nowhere else in the file (only `lockValue` is declared; lines 176/387 mention `lockValue` in comments). A `ReferenceError` thrown in a `finally` block replaces the pending value — so **every** exit path of this handler, including the successful `return res.status(201).json({ escrow_status: 'funded', ... })` at line 295, throws `ReferenceError: lockAcquired is not defined`, turning even successful payment locks into a 500 response. Additionally, because the error is thrown inside `if (lockAcquired)` before `releaseLock(lockKey)` is reached, the Redis payment lock is never explicitly released (only its TTL eventually frees it), defeating the concurrency-protection design documented at lines 176/387.

## Expected Behaviour
The `finally` block releases the lock acquired via `acquireLock` on every exit path (success and error), using the actual owner token, and the success response returns 201 as intended.

## Actual Behaviour
Every invocation of the payment-lock route throws `ReferenceError: lockAcquired is not defined` in the `finally`, so successful payments return 500, errors are masked, and the Redis lock (`lockValue`) is never released by code.

## Proposed Fix
1. Replace the undefined guard with the declared variable:
   ```js
   } finally {
     if (lockValue) {
       try {
         await releaseLock(lockKey, lockValue);
       } catch (releaseErr) {
         logger.error({ err: releaseErr, lockKey }, 'Failed to release payment lock');
       }
     }
   }
   ```
2. Confirm `acquireLock` returns a falsy value (null) when the lock is not acquired (line 202 already handles `lockValue === null` → 409), so `if (lockValue)` is the correct guard.
3. Add a test that a successful lock returns 201 and that the lock key is removed from Redis afterward.

ionfwsrijan:
## Problem

Issue #6068: in `POST /api/payments/lock`, the `finally` block at paymentRoutes.js:314 references **`lockAcquired`**, a variable that is never declared anywhere in the file. Because `finally` runs on every request, every call to the route throws `ReferenceError: lockAcquired is not defined` and returns HTTP 500 — the payment lock API is completely broken, even on the happy path.

A second, masked defect: `releaseLock(lockKey)` was called with a single argument, but `releaseLock(resourceKey, lockValue)` (redisLock.js:109) only deletes the lock **when the caller's owner token matches** (`if (!redisClient || !lockValue) return false;`). Without the owner token the release is a no-op, so even after the ReferenceError was fixed the lock would never be released and would linger for its full TTL.

## Fix

Replaced the broken `finally` block with the intended owner-token flow (the `lockValue` variable was already declared at line 195 with this purpose):

```js
} finally {
  if (lockValue) {
    try {
      await releaseLock(lockKey, lockValue);
    } catch (releaseErr) {
      logger.error({ err: releaseErr, lockKey }, 'Failed to release payment lock');
    }
  }
}
```

- `lockValue` is `null` when the lock was never acquired (409 conflict or 503 `LockAcquisitionError`), so no release is attempted.
- `lockValue` is the UUID owner token after `acquireLock` succeeds, so `releaseLock` now deletes the lock atomically via the Lua owner check.

## Files changed

- `backend/api/src/routes/paymentRoutes.js`

## Testing

- `node --check backend/api/src/routes/paymentRoutes.js` — syntax OK.
- Confirmed via `git grep` that `lockAcquired` has no declaration anywhere; `releaseLock`'s two-arg signature matches the call sites in `orderRoutes.js:1518` and `escrowFundingReconciliation.js:131`.

Closes #6068

### DIFF
--- a/backend/api/src/routes/paymentRoutes.js
+++ b/backend/api/src/routes/paymentRoutes.js
@@ -312,9 +312,9 @@ router.post(
       return res.status(500).json({ error: 'Internal Server Error' });
 
     } finally {
-      if (lockAcquired) {
+      if (lockValue) {
         try {
-          await releaseLock(lockKey);
+          await releaseLock(lockKey, lockValue);
         } catch (releaseErr) {
           logger.error(
             { err: releaseErr, lockKey },


ionfwsrijan:

PR #6083, opened 2026-08-04
fix(ml): remove simulated OCR fallback and validate KYC uploads

### LINKED ISSUE #6070: OCR KYC verifier approves the hardcoded simulated DL number whenever OCR throws
## Description
The OCR-based driver KYC verifier approves a **hardcoded, fake driver's licence number** whenever OCR fails. In `backend/ml/app/models/ocr_verifier.py` (lines 15-56), `extract_text` wraps the OCR call and on **any** exception returns a fixed simulated string instead of failing:

```python
try:
    text = pytesseract.image_to_string(...)
except Exception:
    return "SIMULATED_DL_NUMBER: DL-1234567890123"
```

The simulated value `DL-1234567890123` then matches the simulated regex used to extract the licence number, so `verify_license` (which returns `verified=True` when the number is found) reports the driver as verified. Any document whose OCR throws — Tesseract missing/failing, corrupt or non-image payload, resource exhaustion — results in a **passing KYC check with the fake licence number**, which is also persisted/registered on-chain as the driver's verified licence.

Additionally `backend/ml/main.py:710-719` (`/verify/kyc`) does not validate the upload's content type or size before handing it to the verifier, so arbitrary/malformed requests reach this fallback.

## Expected Behaviour
On OCR failure the verifier must return an error (`verified: false`, no licence number) and the endpoint must reject malformed uploads — never silently approve.

## Actual Behaviour
Any OCR exception yields the hardcoded `DL-1234567890123` "verified" result, letting drivers pass KYC (and register) without a real licence.

## Proposed Fix
1. In `extract_text`, remove the simulated fallback: on exception, re-raise or return an explicit error sentinel; the caller returns `verified: false`.
2. Ensure `verify_license` returns `verified: false` when the extraction failed or no licence number was found.
3. In `/verify/kyc` (`main.py:710-719`), validate content-type (image/* or PDF) and size limits before OCR.
4. Add tests: (a) an image that causes OCR to raise returns `verified: false`; (b) the substring `SIMULATED_DL_NUMBER` never appears in the response.

ionfwsrijan:
## Problem

Issue #6070: two related KYC-verification weaknesses in the ML engine:

1. **Simulated OCR fallback** — `ocr_verifier.py:extract_text` catches any OCR failure and returns the hard-coded string `"SIMULATED_DL_NUMBER: DL-1234567890123"`. Combined with the `simulated_pattern` in `verify_license`, a KYC submission would report `verified: true` for a licence that was **never actually read** — e.g. when Tesseract is unavailable or the image is corrupt. This effectively lets any upload pass KYC.
2. **No upload validation** — `POST /verify/kyc` accepted any `content_type` and any size with no limits, and read unbounded bodies.

## Fix

- `ocr_verifier.py`:
  - `extract_text` now returns `None` on failure instead of fabricating a licence number.
  - `verify_license` handles `None`/empty text and returns `verified: false` / `extracted_number: null`; the `simulated_pattern` recognition is removed.
- `main.py` `POST /verify/kyc`:
  - Rejects non-image content types (`image/jpeg`, `image/png`, `image/webp`) with 422.
  - Enforces a 5 MB size cap (checked against `file.size` before reading, and the actual byte length after read) and rejects empty files with 422.
  - When OCR fails (`extract_text` returns `None`), returns 200 with `verified: false` instead of ever fabricating a pass.

## Files changed

- `backend/ml/app/models/ocr_verifier.py`
- `backend/ml/main.py`
- `backend/ml/tests/test_ocr_verifier.py` (new)

## Testing

- `py -3 -m pytest tests/test_ocr_verifier.py` — **11/11 passed**:
  - `verify_license`: `None`, empty, forged marker, and unrecognised text are all unverified; valid DL patterns (`DL-1420110012345`, `DL1420110012345`) verified.
  - `extract_text`: garbage bytes → `None`.
  - Endpoint: non-image type → 422; empty file → 422; corrupt image → 200 `verified:false`; wrong API key → 401.
- `py -3 -m pytest tests/test_endpoints.py` — 13/14 passed; the sole failure (`test_health`, model-registry set assertion) fails identically on clean `main` and is unrelated.

Closes #6070

### DIFF
--- a/backend/ml/app/models/ocr_verifier.py
+++ b/backend/ml/app/models/ocr_verifier.py
@@ -3,6 +3,7 @@ from PIL import Image
 import io
 import logging
 import re
+from typing import Optional
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


ionfwsrijan:

PR #6086, opened 2026-08-04
fix: prevent OCR fallback from approving invalid KYC

### LINKED ISSUE #6070: OCR KYC verifier approves the hardcoded simulated DL number whenever OCR throws

shivambind269-ai:
## Summary

This PR fixes a security issue where OCR failures could incorrectly approve KYC using a hardcoded simulated driving licence number.

### Changes Made

- Removed the simulated OCR fallback in `extract_text()`.
- OCR failures now return a failed verification instead of approving KYC.
- Removed simulated driving licence pattern matching.
- Added upload content-type validation.
- Added upload file-size validation before OCR processing.

### Why

Previously, OCR exceptions returned a hardcoded simulated licence number (`DL-1234567890123`), which could cause invalid documents to be marked as verified.

This change ensures OCR failures do not result in successful KYC verification.

Closes #6070

## Summary by CodeRabbit

- **New Features**
  - Added validation for KYC document uploads, allowing only image and PDF files up to 5 MB.
- **Bug Fixes**
  - Improved handling of invalid uploads and validation errors.
  - Unexpected document-processing failures now return an unverified result instead of a server error.
  - OCR verification no longer accepts simulated license text, improving verification accuracy.
  - OCR failures are now reported and handled consistently.

### DIFF
--- a/backend/ml/app/models/ocr_verifier.py
+++ b/backend/ml/app/models/ocr_verifier.py
@@ -6,6 +6,7 @@ import re
 
 logger = logging.getLogger(__name__)
 
+
 class OCRVerifier:
     def __init__(self):
         # In a real environment, you might need to set the tesseract_cmd path if it's not in PATH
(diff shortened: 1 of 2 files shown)


shivambind269-ai:
GSSOC

PR #6155, opened 2026-08-04
fix: evaluate unusual-time window in UTC (Issue #6127)

### LINKED ISSUE #6127: bug: anomalyDetectionService.detectUnusualTime compares server-local hours but labels them UTC

ionfwsrijan:
## Problem

`AnomalyDetectionService.detectUnusualTime` compared `Date.prototype.getHours()` (the server-local hour) against the `UNUSUAL_TIME` window while the message and `time` field both claim UTC. On a non-UTC server (e.g. IST, UTC+5:30), the wrong transactions are flagged and labelled — a false/missed flag in the security detector.

## Fix

Evaluate the hour with `getUTCHours()` so the window comparison, the `toISOString()` output and the `:00 UTC` label all refer to the same (UTC) timezone. Added a unit test that pins the expected behavior for known UTC instants, including the IST-misclassification case.

## Files changed

- `backend/api/src/services/security/anomalyDetectionService.js`
- `backend/api/test/unit/anomalyDetectionService.test.js`

## Testing

- New vitest unit tests cover: in-window (01:30Z flagged), out-of-window (18:30Z not flagged), and late-evening UTC (23:30Z not flagged even though it is 05:00 local on a UTC+5:30 server).
- `node --check` passes on the modified source file.

Closes #6127

### DIFF
--- a/backend/api/src/services/security/anomalyDetectionService.js
+++ b/backend/api/src/services/security/anomalyDetectionService.js
@@ -130,7 +130,7 @@ class AnomalyDetectionService {
 
   detectUnusualTime(transaction) {
     const txTime = new Date(transaction.timestamp);
-    const hour = txTime.getHours();
+    const hour = txTime.getUTCHours();
 
     if (hour >= ANOMALY_THRESHOLDS.UNUSUAL_TIME.startHour &&
         hour < ANOMALY_THRESHOLDS.UNUSUAL_TIME.endHour) {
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

PR #6160, opened 2026-08-04
fix: stop fabricating tx_hash in booking_confirmation lock flow (Issue #6133)

### LINKED ISSUE #6133: bug: customer booking_confirmation fabricates a tx_hash and posts it to /api/payments/lock - escrow lock can never succeed
## Description

In `apps/customer/lib/screens/booking_confirmation_screen.dart:222-238`, when the customer taps "I've Paid" after the UPI deep link, the app **fabricates** a transaction hash instead of capturing the real on-chain transaction:

```dart
Future<void> _confirmPaymentLocked() async {
  if (_createdOrderId == null) return;
  setState(() => _isSubmitting = true);
  try {
    // In mock UPI mode, generate a placeholder tx_hash
    // In production, the wallet SDK would provide the real on-chain hash
    final mockTxHash =
        '0x${_createdOrderId!.replaceAll('-', '').padRight(64, '0').substring(0, 64)}';

    await _apiClient.post('/api/payments/lock', body: {
      'order_id': _createdOrderId,
      'tx_hash': mockTxHash,
    });
```

The backend `POST /api/payments/lock` (`backend/api/src/routes/paymentRoutes.js:163-169`) is explicitly documented as: "Called by the customer Flutter app AFTER the customer's wallet has submitted the createBooking() transaction on-chain. ... Calls recordDepositTx() to verify tx on Polygon." `recordDepositTx` (`services/escrow.js:254`) does `provider.getTransaction(txHash)` and validates sender/amount against the booking.

A hash synthesized from the order id has no on-chain transaction: `provider.getTransaction` returns null and the lock fails (or, if validation is ever skipped, records a fake deposit). Either way the real UPI payment the customer made is never reconciled to an escrow booking — the escrow is never funded from the customer's side, the order stays stuck in `funding`, and the driver is never paid through escrow.

- **Expected:** the wallet SDK submits the real `createBooking` transaction on-chain and returns its actual `tx_hash`, which is then posted to `/api/payments/lock` and verified against the chain.
- **Actual:** the app computes a fake 64-char hash from the order id and posts it; against a real chain the lock verification cannot succeed, so the funded escrow booking never materializes.

## Proposed Fix

1. Replace the fabricated hash with the real transaction hash returned by the wallet/on-chain SDK after `createBooking` is mined.
2. If no wallet SDK is integrated, disable the "I've Paid" path (or gate it behind an explicit dev-only flag) instead of silently sending a fake hash — the backend already has the right fail-closed behavior.
3. On `lock` failure, surface the actual error to the user (retry/verification instructions) rather than treating the payment as completed.

ionfwsrijan:
## Problem

When the customer tapped "I've Paid", `booking_confirmation_screen.dart` fabricated a 64-char `tx_hash` from the order id and posted it to `/api/payments/lock`. The backend (`recordDepositTx`) verifies the hash against the real Polygon chain, so the lock could never succeed — the escrow was never funded from the customer's side and the order stayed stuck in `funding`, while the app reported the payment as completed.

## Fix

Removed the fabricated hash entirely. Since no wallet SDK is integrated in this app to return a real on-chain `createBooking` hash, the "I've Paid" path now fails closed: it no longer sends a fake hash to the backend and instead tells the customer that on-chain escrow confirmation is unavailable and to contact support — it never reports the payment as locked.

## Files changed

- `apps/customer/lib/screens/booking_confirmation_screen.dart`

## Testing

- No code path posts a fabricated `tx_hash` anymore.
- The state stays honest: booking created, escrow not locked, customer informed of the unresolved payment.

Closes #6133

### DIFF
--- a/apps/customer/lib/screens/booking_confirmation_screen.dart
+++ b/apps/customer/lib/screens/booking_confirmation_screen.dart
@@ -219,34 +219,25 @@ class _BookingConfirmationScreenState extends State<BookingConfirmationScreen>
   }
 
   // ── Step 4: Lock payment after UPI success ────────────────────────────────
+  // POST /api/payments/lock requires the real on-chain tx_hash that the wallet
+  // SDK returns once `createBooking` is mined; recordDepositTx() verifies it
+  // against Polygon. No wallet SDK is integrated in this app, so there is no
+  // real hash to submit. Fabricating one would post a hash the backend can
+  // never verify — the escrow would never lock and the booking would stay
+  // stuck in `funding`. Fail closed instead: never send a fake hash and never
+  // report the payment as locked.
   Future<void> _confirmPaymentLocked() async {
     if (_createdOrderId == null) return;
-    setState(() => _isSubmitting = true);
 
-    try {
-      // In mock UPI mode, generate a placeholder tx_hash
-      // In production, the wallet SDK would provide the real on-chain hash
-      final mockTxHash =
-          '0x${_createdOrderId!.replaceAll('-', '').padRight(64, '0').substring(0, 64)}';
-
-      await _apiClient.post(
-        '/api/payments/lock',
-        body: {
-          'order_id': _createdOrderId,
-          'tx_hash': mockTxHash,
-        },
-      );
-
-      _showSuccessPanel();
-    } catch (e) {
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #6161, opened 2026-08-04
fix: report reconciliation-required state in confirm-otp 202 response (Issue #6134)

### LINKED ISSUE #6134: bug: confirm-otp 202 (escrowUpdateFailed) response still reports payment_released:true and escrow_status:released
## Description

In `backend/api/src/routes/orderRoutes.js:1134-1149`, the delivery confirmation endpoint has two response paths. When `verifyDeliveryFn` returns `escrowUpdateFailed: true` (the on-chain escrow was released but the local DB escrow/payment update failed, so reconciliation is required), the 202 response still reports the delivery as fully paid:

```js
if (escrowUpdateFailed) {
  return res.status(202).json({
    message: 'Delivery confirmed. Escrow payout requires reconciliation.',
    payment_released: true,        // <-- claims payment is released
    escrow_status: 'released',     // <-- claims escrow is released
    amount_inr: amountInr,
  });
}
```

The success path below it returns the same two fields for the genuinely-released case, so consumers cannot distinguish "released, nothing else to do" from "released but reconciliation required". The driver app (`apps/driver/lib/screens/delivery_otp_screen.dart:182-183`) reads `amount_inr` and shows the "Payment Released" panel without inspecting `escrowUpdateFailed`, so drivers are told the payout is complete even though the backend knows the escrow update failed and needs reconciliation.

- **Expected:** the 202 reconciliation-required path reports the true, degraded state — e.g. `payment_released: false`, `escrow_status: 'released'` (or a distinct `reconciliation_required: true` flag), and a message that the payout is pending reconciliation. Clients then show the appropriate "payment pending reconciliation" UI.
- **Actual:** the 202 path mirrors the success response exactly (`payment_released: true`, `escrow_status: 'released'`), so the API and the driver UI both claim the payment is released when the escrow update actually failed.

## Proposed Fix

1. In the 202 branch, set an explicit `reconciliation_required: true`, and either set `payment_released: false` or rename the field semantics so the failure state is unambiguous (keep `escrow_status` reflecting the true status).
2. Update the driver app to check `reconciliation_required` (or `escrowUpdateFailed`) before showing the "Payment Released" panel, and show a "payout pending reconciliation" state instead.
3. Keep the 202 HTTP status but ensure monitoring/admin tooling can alert on reconciliation-required deliveries (log and/or metric the `escrowUpdateFailed` occurrence).

ionfwsrijan:
## Problem

The confirm-OTP 202 response for the `escrowUpdateFailed` path reported `payment_released: true` and `escrow_status: 'released'` — identical to the genuinely-released success path. Consumers could not distinguish "released, nothing else to do" from "released but reconciliation required", so the driver app showed the "Payment Released" panel even though the backend knew the escrow update failed and needs reconciliation.

## Fix

- Backend: the 202 branch now returns `payment_released: false`, adds `reconciliation_required: true`, keeps `escrow_status: 'released'` (the true on-chain status), and logs a warning so monitoring/alerting can catch reconciliation-required deliveries.
- App: `delivery_otp_screen.dart` now checks `reconciliation_required` and shows a "Payout Pending Reconciliation" banner instead of the "Payment Released" panel, while keeping the 202 semantics clear.

## Files changed

- `backend/api/src/routes/orderRoutes.js`
- `apps/driver/lib/screens/delivery_otp_screen.dart`

## Testing

- Verified the driver app reads `reconciliation_required` from the 202 body and routes to the reconciliation banner.
- `node --check` passes on the modified backend file.

Closes #6134

## Summary by CodeRabbit

* **New Features**
  * Added a pending payout state after delivery OTP confirmation when reconciliation is required.
  * Displayed an animated banner with the pending payout amount and order details.
  * Added clearer messaging for payments awaiting reconciliation.

* **Bug Fixes**
  * Improved delivery confirmation handling to accurately show whether payment was released or remains pending.
  * Removed confusing inline error messaging for reconciliation-related responses.


ionfwsrijan:

PR #6163, opened 2026-08-04
fix: remove dead 501 verify-otp stub (Issue #6136)

### LINKED ISSUE #6136: bug: POST /api/auth/verify-otp always returns 501 Not Implemented despite being documented as the OTP verification endpoint
## Description

`backend/api/src/routes/authRoutes.js:179-188`:

```js
/**
 * @openapi
 * /api/auth/verify-otp:
 *   post:
 *     tags: [Authentication]
 *     summary: Verify OTP
 *     description: Endpoint for verifying OTPs. Protected by strict rate limiting to prevent brute-forcing.
 *     responses:
 *       501:
 *         description: Not Implemented
 */
router.post("/verify-otp", otpVerificationLimiter, async (req, res) => {
  // To be implemented: backend OTP verification logic.
  // This endpoint serves as a rate-limited proxy/placeholder to satisfy
  // security requirements preventing OTP brute forcing.
  return res.status(501).json({
    success: false,
    error: "Not Implemented",
    message: "OTP verification logic should be executed here.",
  });
});
```

The route is mounted at `app.use('/api/auth', authLimiter, authRoutes)` (`index.js:459`). The OpenAPI spec describes it as the OTP verification endpoint, but the handler unconditionally returns `501 Not Implemented` — it never verifies anything, never consults the OTP store, and never issues a session. Any client following the documented contract (or a future app flow migrating OTP login to the backend) receives a permanent error; the endpoint exists only as a rate-limit choke point and provides no actual OTP verification.

- **Expected:** `POST /api/auth/verify-otp` verifies the submitted OTP (against the stored/expiring code), returns success or a controlled failure, and either completes the verification side-effect or issues a session — matching its OpenAPI description.
- **Actual:** the endpoint always returns `501 Not Implemented` with the same static body; no OTP is ever verified through this documented endpoint.

## Proposed Fix

Implement the verification logic: validate the request body (phone/email + otp), look up the pending code (rate-limited and expiring), compare it with a timing-safe check, mark it consumed on success, and return `{ success: true }` or a 4xx with the reason. Keep the existing `otpVerificationLimiter`. If OTP verification intentionally lives entirely in Supabase and this endpoint is not needed, remove the route and its OpenAPI block instead of shipping a permanently-501 stub that is documented as functional.

ionfwsrijan:
## Problem

`POST /api/auth/verify-otp` was documented in its OpenAPI block as the OTP verification endpoint, but the handler unconditionally returned `501 Not Implemented`. It never verified anything, never consulted an OTP store and never issued a session — a permanent stub that is documented as functional.

## Fix

OTP login verification is handled entirely by Supabase auth (phone OTP), and no client or other backend code calls this endpoint (repo-wide grep confirms the only references are the route itself). Per the issue's stated alternative, removed the dead 501 route and its OpenAPI block instead of shipping a documented-but-broken endpoint. Also removed the now-unused `otpVerificationLimiter` import.

## Files changed

- `backend/api/src/routes/authRoutes.js`

## Testing

- Confirmed nothing in the codebase references `/api/auth/verify-otp` besides the removed route.
- `node --check` passes on the modified file.

Closes #6136

### DIFF
--- a/.github/scripts/retro-labeler.js
+++ b/.github/scripts/retro-labeler.js
@@ -158,17 +158,43 @@ function deduplicateTypeLabels({ currentLabels, changedFiles, prTitle, rules }) 
   return { toKeep: winnerLabel, toRemove };
 }
 
+function isRateLimitError(error) {
+  if (!error) return false;
+  if (error.status === 429) return true;
+  if (error.status === 403) {
+    const msg = String(error.message || '').toLowerCase();
+    if (msg.includes('rate limit') || msg.includes('secondary rate limit') || msg.includes('exceeded')) {
+      return true;
+    }
+    const resMsg = String(error.response?.data?.message || '').toLowerCase();
+    if (resMsg.includes('rate limit') || resMsg.includes('exceeded')) {
+      return true;
+    }
+  }
+  return false;
+}
+
 async function run({ github, context, core, dryRun = false, prState = 'closed' }) {
   const { owner, repo } = context.repo;
 
   core.info(`Starting retrospective PR labeler (dryRun = ${dryRun})...`);
 
   // Fetch available labels in repo to check if we need to create them
-  const repoLabels = await github.paginate(github.rest.issues.listLabelsForRepo, {
-    owner,
-    repo,
-    per_page: 100
-  });
+  let repoLabels = [];
+  try {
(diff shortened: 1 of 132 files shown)


ionfwsrijan:

PR #6189, opened 2026-08-04
fix(security): use UTC hour in anomalyDetectionService detectUnusualTime to match the claimed UTC timezone

### LINKED ISSUE #6127: bug: anomalyDetectionService.detectUnusualTime compares server-local hours but labels them UTC

tmdeveloper007:
## Description

`backend/api/src/services/security/anomalyDetectionService.js:133` uses `Date.prototype.getHours()` which returns the server-local wall-clock hour, but the returned anomaly labels the output as UTC. On servers running non-UTC timezones (e.g. IST, UTC+5:30), this causes the `UNUSUAL_TIME` detection window to be evaluated against the wrong hour.

Changed `getHours()` to `getUTCHours()` so the UTC hour from the stored ISO timestamp is compared against the UTC threshold consistently.

## Code Change

```diff
- const hour = txTime.getHours();
+ const hour = txTime.getUTCHours();
```

## Related Issues

Fixes #6127

## Type of Change

- Bug fix (non-breaking change that fixes an issue)

## Checklist

- ESLint passes on all modified files
- No new test failures introduced
- Minimal, targeted one-line fix

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved delivery verification accuracy by consistently converting location distances to meters.
  - Corrected unusual-activity detection to evaluate transaction times consistently in UTC.
  - Improved payment and driver verification lock cleanup to prevent stale locks and related errors.
  - Preserved existing error handling and verification behavior while improving reliability.

### DIFF
--- a/backend/api/src/security/policyEngine.js
+++ b/backend/api/src/security/policyEngine.js
@@ -130,6 +130,7 @@ const POLICIES = {
   'liquibase:rollback':        { roles: [ROLES.ADMIN] },
 
   'ebpf:manage':               { roles: [ROLES.ADMIN] },
+  'snyk:manage':               { roles: [ROLES.ADMIN] },
 };
 
 export class PolicyEngine {
--- a/snyk/routes.js
+++ b/snyk/routes.js
@@ -1,8 +1,14 @@
 import express from 'express';
 import snykService from './snyk.service.js';
 import logger from '../backend/api/src/middleware/logger.js';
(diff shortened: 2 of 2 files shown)


tmdeveloper007:
The merge conflicts have been resolved and the branch is up to date with main.

PR #6295, opened 2026-08-04
fix : replaced non-existent orderRepository.findOrderByIdOrDisplayId with orderValidationService

### LINKED ISSUE #6290: fix : add missing findOrderByIdOrDisplayId method to paymentRoutes
paymentRoutes.js and orderRoutes.js call orderRepository.findOrderByIdOrDisplayId() which does not exist on the orderRepository. The correct method exists on orderValidationService.findOrderByIdOrDisplayId(). This causes HTTP 500 for the UPI payment flow (/api/payments/upi-intent, /api/payments/lock, /api/payments/charge-and-lock) and the confirm-OTP flow in orderRoutes.js.

Changes that Need to be Made:
- backend/api/src/routes/paymentRoutes.js lines 107, 209, 341: replace orderRepository.findOrderByIdOrDisplayId(...) with orderValidationService.findOrderByIdOrDisplayId(...)
- backend/api/src/routes/orderRoutes.js line 1163: replace orderRepository.findOrderByIdOrDisplayId(...) with orderValidationService.findOrderByIdOrDisplayId(...)

Impact that it would Provide:
- Fixes HTTP 500 errors in the UPI payment and OTP confirmation flows
- Makes the payment/escrow flow functional end-to-end

tmdeveloper007:
Closes #6290.

Replaced 4 calls to the non-existent orderRepository.findOrderByIdOrDisplayId() method with the existing orderValidationService.findOrderByIdOrDisplayId() method. Fixed paymentRoutes.js (upi-intent, lock, status endpoints) and orderRoutes.js (confirm-otp endpoint). Added proper try/catch for error handling consistent with the service's DomainError throwing behavior.

Changes Made:
- backend/api/src/routes/paymentRoutes.js: added orderValidationService import, replaced 3 orderRepository calls with orderValidationService calls using try/catch
- backend/api/src/routes/orderRoutes.js: replaced orderRepository.findOrderByIdOrDisplayId with orderValidationService.findOrderByIdOrDisplayId in confirm-otp handler

Impact it Made:
- Fixes HTTP 500 errors in UPI payment and OTP confirmation flows
- Makes the payment/escrow flow functional end-to-end
- Backend ESLint passes for the changed files

Note: This PR is submitted as part of GirlScript Summer of Code (GSSOC). Please assign this PR to the `tmdeveloper007` account.

### DIFF
--- a/backend/api/src/lib/profileCache.js
+++ b/backend/api/src/lib/profileCache.js
@@ -76,7 +76,7 @@ function logCacheError(operation, error) {
 function getRedisClient() {
   try {
     return db.redisClient ?? null;
-code -g backend/api/src/routes/lookupRoutes.js:50  } catch {
+  } catch {
     return null;
   }
 }
--- a/backend/api/src/routes/authRoutes.js
+++ b/backend/api/src/routes/authRoutes.js
@@ -51,7 +51,7 @@ import {
   invalidateCachedProfile,
   invalidateCachedSupabaseProfile,
 } from "../lib/profileCache.js";
-import { firebaseAdmin } from "../config/db.js";
+import { firebaseAdmin, supabase } from "../config/db.js";
 import logger from "../middleware/logger.js";
 
 const router = express.Router();
@@ -166,7 +166,6 @@ router.get("/session", authenticate, userLimiter, (req, res) => {
 });
 
 import crypto from "crypto";
-import { supabase } from "../config/db.js";
 import { otpSendSchema } from "../validation/requestSchemas.js";
 import { z } from "zod";
 
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -1160,11 +1160,11 @@ router.post(
       );
 
       // Fetch the released amount to include in the response
-      const { data: order } = await orderRepository.findOrderByIdOrDisplayId(
+      const orderForAmount = await orderValidationService.findOrderByIdOrDisplayId(
         req.params.id,
         'total_amount, order_display_id'
       );
-      const amountInr = order?.total_amount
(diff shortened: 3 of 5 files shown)


advikdivekar:
Heads up on one line in this diff — I ran into the same code path and think there's a bug in the `confirm-otp` change.

The rename resolves the duplicate `order` binding (which is what currently makes the module unparseable), but the ternary body still reads from the old binding:

```js
const orderForAmount = await orderValidationService.findOrderByIdOrDisplayId(
  req.params.id,
  'total_amount, order_display_id'
);
const amountInr = orderForAmount?.total_amount
  ? (order.total_amount / 100).toFixed(0)   // <-- still `order`, not `orderForAmount`
  : null;
```

`order` is the authorisation-scoped row fetched earlier with `'id, driver_id, customer_id'`, so it has no `total_amount`. The guard passes on `orderForAmount`, then the body evaluates `(undefined / 100).toFixed(0)`, which returns the string `"NaN"`. Drivers would see `amount_inr: "NaN"` on every successful delivery confirmation.

The `return res.json({...})` a few lines below has the same issue — `order_display_id: order?.order_display_id` reads the auth-scoped row, which doesn't select that column either, so it resolves to `undefined`.

Two other things you may want to fold in while you're in here:

1. The guard is a truthiness check, so a legitimate `total_amount` of `0` reports as `null`. `Number.isFinite(orderForAmount?.total_amount)` handles that.
2. `orderValidationService.findOrderByIdOrDisplayId` throws a `DomainError` on query failure rather than returning an error envelope. In `confirm-otp` the delivery is already verified by that point, so an exception there would turn a successful delivery into a 500 — worth wrapping so it degrades to omitting the amount instead.

Not blocking, just flagging since I hit the same lines. Happy to be wrong if I've misread the scoping.

Issue #5780, opened 2026-08-02
bug: Kafka Consumers Are Not Idempotent — Duplicate Event Delivery Causes Double Writes and Double Credits


ionfwsrijan:
## Description
`backend/kafka/consumers/order.consumer.js` runs at-least-once consumers and re-publishes every event to the in-process EventBus with deduplication explicitly **off**:
```js
// order.consumer.js:93-100
if (this._eventBus) {
  const eventType = topic.replace(/\./g, '_').toUpperCase();
  this._eventBus.publish(eventType, message, {
    adapters: [],
    deduplicate: false,        // <-- no dedup
    source: `kafka:${groupId}`,
  });
}
```
Handlers registered via `registerHandler`/`registerHandlerViaEventBus` receive the same message on every redelivery (Kafka rebalances, consumer restarts, or transient handler errors after the offset is committed). There is no idempotency key on the downstream writes (no `ON CONFLICT` on the domain tables' natural keys, no processed-message registry). Duplicate delivery of `PAYMENT_CONFIRMED`, `TRIP_COMPLETED`, or `ESCROW_RELEASED` therefore leads to duplicate order updates, repeated wallet/earnings credits, duplicate notifications, and inconsistent read-model state — the exact class of bug that the "dead-letter queue" handling (lines 103-110) cannot protect against because it only catches thrown errors, not logical duplicates.

## Expected Behavior
Each event should be processed exactly once (idempotent handlers keyed on the event id / natural key).

## Actual Behavior
Messages can be applied more than once with no dedup, causing duplicate writes and double credits.

## Affected Files
- `backend/kafka/consumers/order.consumer.js` (lines 81-101) — handler runs every delivery, EventBus publish with `deduplicate: false`
- `backend/kafka/consumers/order.consumer.js` (lines 103-110) — DLQ only for thrown errors, no idempotency
- `backend/kafka/config/kafka.config.js` — consumers use default at-least-once semantics
- `backend/kafka/events/order.events.js`, `backend/kafka/cqrs/order.read.model.js` — read-model/event handlers with no dedup guard

## Proposed Fix
[...]

PR #6302, opened 2026-08-05
fix: use service-role client for kafka idempotency registry

### LINKED ISSUE #6288: Kafka idempotency registry writes with anon-keyed client against service-role-only table — every claim fails
**Description**

The Kafka idempotency registry added to fix #5780 cannot ever record a processed event, because it writes with the anon-keyed client against a service-role-only table.

`backend/kafka/repositories/processedEvent.repository.js:1` imports `supabase` from `../api/src/config/db.js`, and `backend/api/src/config/db.js:33` initializes that export with `createClient(supabaseUrl, supabaseAnonKey, ...)` ("anon key — RLS enforced", line 39). The service-role client is the separate `supabaseAdmin` export (line 77).

The migration that created the registry (`supabase/migrations/20260802140000_kafka_processed_events_idempotency.sql`) states the table "is written by backend services using the service_role key" and:

- enables RLS and creates a policy **for `service_role` only** (lines 35-43),
- `revoke all on table kafka_processed_events from anon, authenticated` (line 45).

Therefore every `claimProcessed` upsert (`processedEvent.repository.js:16-26`) executed through the anon-keyed client is rejected by RLS, `claimProcessed` throws (line 30-33), and the idempotency gate never works. With at-least-once Kafka delivery, duplicate `payment.confirmed` / `trip.completed` messages are re-processed (double wallet/earnings credits, duplicate read-model updates) — exactly what the registry was built to prevent.

**Expected Behavior**

`claimProcessed` writes to `kafka_processed_events` via the service-role client and returns `false` for duplicates, so redelivered messages are skipped.

**Actual Behavior**

The registry upsert runs with an anon-keyed client and is rejected by the service-role-only RLS, so every claim fails; the idempotency check is defeated and duplicate events are applied more than once.

**Proposed Fix**

Use the service-role client (`supabaseAdmin`) in `processedEvent.repository.js` (and audit the other Kafka repositories using `supabase` for the same issue), then add a test that runs `claimProcessed` twice with the same `(topic, event_id)` and asserts the second call returns `false`.

ionfwsrijan:
## Problem

The Kafka idempotency registry added for #5780 can never record a processed event. `backend/kafka/repositories/processedEvent.repository.js` imported `supabase` from `backend/api/src/config/db.js`, which initializes the anon-keyed client ("anon key — RLS enforced"). The registry table `kafka_processed_events` enables RLS with a policy **for `service_role` only** and `revoke all ... from anon, authenticated` (`supabase/migrations/20260802140000_kafka_processed_events_idempotency.sql`).

Therefore every `claimProcessed` upsert through the anon-keyed client was rejected by RLS, `claimProcessed` threw, and the idempotency gate never worked. With at-least-once Kafka delivery, duplicate `payment.confirmed` / `trip.completed` messages were re-processed (double wallet/earnings credits, duplicate read-model updates) — exactly what the registry was built to prevent.

## Fix

Switch `processedEvent.repository.js` to the service-role client (`supabaseAdmin`), which is allowed by the table's RLS policy, so `claimProcessed` records the event on first claim and returns `false` for redelivered messages. The other Kafka repositories were audited: `event.repository.js` (`events`) and `deadLetter.repository.js` (`kafka_dead_letters`) target tables that have no service-role-only RLS restriction in migrations, so the anon client is not breaking for them; the registry table is the only one gated to service_role.

## Files changed

- `backend/kafka/repositories/processedEvent.repository.js` — `claimProcessed` now uses `supabaseAdmin` (service-role client)
- `backend/kafka/test/processedEvent.repository.test.js` — new unit test: first claim returns `true`, a second claim for the same `(topic, event_id)` returns `false`, different topics are distinct keys, and claims go through the service-role client

## Testing

- `node --check` passes on the modified source and the new test.
- New vitest unit test (`backend/kafka/test/processedEvent.repository.test.js`) mocks the `db.js`/`logger.js` modules and asserts the duplicate claim path returns `false`. Note: local `node_modules` are not installed in this workspace, so the vitest suite itself was not executed here; it runs in CI via `npm test`.

Closes #6288

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when claiming processed events by using the appropriate administrative database access.

* **Tests**
  * Added coverage for new and duplicate event claims.
  * Verified that events are distinguished correctly by topic and event ID.
  * Confirmed claims use the administrative database connection.


ionfwsrijan:

PR #6303, opened 2026-08-05
fix: never restore funds when settlement RPC fails after payout dispatch

### LINKED ISSUE #6274: Withdrawal settlement worker double-pays when settle RPC fails after successful payout dispatch
**Description**

`backend/api/src/workers/withdrawalSettlementWorker.js:47-76` settles each pending withdrawal by (1) dispatching the payout through the configured provider (`dispatchPayout`), then (2) marking it completed via the `settle_withdrawal_tx` RPC. If step 2 fails for any reason (transient Supabase/network error, RPC timeout), the `catch` at line 64 runs `fail_withdrawal_tx`, whose whole purpose is to restore the reserved funds to the driver:

- `supabase/migrations/20260802060000_withdrawal_settlement.sql` — `fail_withdrawal_tx` does `wallet_confirmed = wallet_confirmed + v_amount`.

But at that point the payout has ALREADY been dispatched to the driver's bank/UPI account. The driver receives the real payout **and** gets the withdrawn amount restored to `wallet_confirmed`, so they can withdraw it again. Because the withdrawal row is marked `failed`, it is also excluded from the next settlement sweep's `status = 'pending'` filter only if it stays failed; more importantly, the restored balance permits an immediate re-withdrawal, producing an unlimited double-payout loop whenever the settle RPC lags after a successful dispatch.

**Expected Behavior**

The settlement worker must distinguish two failure modes:
- Payout dispatch failed before money left the platform → safe to restore funds to `wallet_confirmed`.
- Payout dispatch succeeded but the DB completion RPC failed → never restore funds; retry `settle_withdrawal_tx` (it is idempotent and matches only `pending` rows) instead of failing the withdrawal.

**Actual Behavior**

Any error after a successful `dispatchPayout` falls into the catch and calls `fail_withdrawal_tx`, restoring the already-paid amount to `wallet_confirmed`. A driver whose payout was sent but whose completion marker was not persisted is paid twice (bank payout + restored wallet balance), with no compensating action and no alert.

**Proposed Fix**

Split the worker loop so that only failures in `dispatchPayout` (or a definitive provider rejection) call `fail_withdrawal_tx`. When `settle_withdrawal_tx` fails after a successful dispatch, log and retry the settle call (bounded backoff) while keeping the row `pending`; add a payout-outcome column (e.g. `payout_attempted_at`) so a crash between dispatch and settle can be detected and settled, not failed. Add a unit test that simulates dispatch success + settle RPC failure and asserts the balance is not restored.

ionfwsrijan:
## Problem

`backend/api/src/workers/withdrawalSettlementWorker.js` settled each pending withdrawal by (1) dispatching the payout through the provider (`dispatchPayout`), then (2) marking it completed via `settle_withdrawal_tx`. If step 2 failed (transient Supabase/network error, RPC timeout), the `catch` called `fail_withdrawal_tx`, which restores the withdrawn amount to `wallet_confirmed` (`supabase/migrations/20260802060000_withdrawal_settlement.sql`). Because the payout had already been dispatched, the driver received the real bank/UPI payout **and** the balance was restored for a second withdrawal — an unlimited double-payout loop whenever the settle RPC lags after a successful dispatch.

## Fix

Split the worker loop into two distinct failure modes:

- **Payout dispatch failed** before money left the platform → `fail_withdrawal_tx` is still called and the reserved funds are restored to `wallet_confirmed`.
- **`settle_withdrawal_tx` failed after a successful dispatch** → funds are **never** restored. The row stays `pending`, a new `payout_attempted_at` column is recorded right after dispatch, and the idempotent settle call is retried with bounded backoff. If a crash happens between dispatch and settle, the next sweep detects `payout_attempted_at` set, skips re-dispatching, and settles instead of failing.

## Files changed

- `backend/api/src/workers/withdrawalSettlementWorker.js` — split dispatch/settle failure handling, added `recordPayoutAttempted` and `settleWithRetry`, and skipped re-dispatch for already-attempted withdrawals
- `supabase/migrations/20260805000000_withdrawal_settlement_payout_attempt.sql` — new `payout_attempted_at` column on `wallet_transactions` plus a partial index for payout-attempted-but-unsettled rows
- `backend/api/test/unit/withdrawalSettlementWorker.test.js` — updated and extended: dispatch-success + settle-failure asserts `fail_withdrawal_tx` is never called (balance not restored) and settle is retried; payout-attempted rows are not re-dispatched

## Testing

- `node --check` passes on the modified source and tests.
- Updated vitest unit tests cover: normal settle, dispatch failure restores funds, dispatch-success + settle-failure never restores funds and retries settle, and no re-dispatch of already-attempted withdrawals. Note: local `node_modules` are not installed in this workspace, so the vitest suite itself was not executed here; it runs in CI via `npm test`.

Closes #6274

### DIFF
--- a/backend/api/src/workers/withdrawalSettlementWorker.js
+++ b/backend/api/src/workers/withdrawalSettlementWorker.js
@@ -4,8 +4,54 @@ import { dispatchPayout, isPayoutProviderConfigured } from '../services/wallet/p
 import { WorkerTracer } from '../core/telemetry/WorkerTracer.js';
 
 const BATCH_LIMIT = 50;
+const SETTLE_RETRY_ATTEMPTS = 3;
+const SETTLE_RETRY_DELAYS_MS = [500, 1500, 3000];
 
 let intervalId = null;
+
+const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
+
+/**
+ * Records that a payout was dispatched for this withdrawal so a crash between
+ * dispatch and the completion RPC is detected and re-settled (not failed) on
+ * the next sweep. Best-effort: failure to record only delays detection.
+ */
+async function recordPayoutAttempted(withdrawalId, settlementRef) {
+  const { error } = await supabaseAdmin
+    .from('wallet_transactions')
+    .update({
+      payout_attempted_at: new Date().toISOString(),
+      settlement_ref: settlementRef,
+    })
+    .eq('id', withdrawalId)
+    .is('payout_attempted_at', null);
+
+  if (error) {
+    logger.warn(`[WithdrawalSettlementWorker] Failed to record payout attempt for ${withdrawalId}: ${error.message}`);
+  }
+}
+
+/**
+ * Marks a pending withdrawal completed. Retried with a bounded backoff because
+ * settle_withdrawal_tx is idempotent and only matches rows still in 'pending'.
+ */
+async function settleWithRetry(withdrawalId, settlementRef) {
+  let lastError = null;
+  for (let attempt = 1; attempt <= SETTLE_RETRY_ATTEMPTS; attempt += 1) {
+    const { error } = await supabaseAdmin.rpc('settle_withdrawal_tx', {
+      p_withdrawal_id: withdrawalId,
+      p_settlement_ref: settlementRef,
+    });
+    if (!error) {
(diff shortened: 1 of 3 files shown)


ionfwsrijan:

PR #6306, opened 2026-08-05
fix: guard acceptBid against funded/released/terminal orders (#6278)

### LINKED ISSUE #6278: acceptBid lacks guards for funded/released/cancelled orders — re-accepting resets escrow_status to funding
**Description**

`backend/api/src/services/order/bidAcceptanceService.js` `acceptBid` has only one escrow-state guard: it rejects a new acceptance when `escrow_status === 'funding' && pending_bid_acceptance` (lines 40-44). There is no guard for orders whose escrow is already `funded`, `released`, `refunded`, or `cancelled`, and no check on the order status being terminal.

Line 156 then calls `updateEscrowBooking(orderId, bookingId, 'funding', {...})`, which unconditionally overwrites `escrow_status`, `escrow_booking_id`, `escrow_driver_wallet`, and `pending_bid_acceptance` with the new bid's data.

Concretely: a customer accepts bid A, funds the escrow (deposit confirmed, `escrow_status = 'funded'`, driver A assigned via `accept_bid_tx`), and the trip starts. If the customer then accepts bid B for the same order, the guard at line 40 does not fire (status is `funded`, not `funding`), and `updateEscrowBooking` resets the order to `escrow_status = 'funding'` with a brand-new booking id for bid B while driver A is still assigned and the original escrow still holds the funds on-chain. The order is now in an inconsistent state: driver A assigned, escrow reference pointing at an unfunded booking B, and a second deposit can be requested.

**Expected Behavior**

`acceptBid` must reject any acceptance when the order's escrow is `funded`/`released`/`refunded` or the order is in a terminal state, and the escrow-booking update must be conditional so it cannot reset an already-funded escrow.

**Actual Behavior**

An order with a funded/released escrow can be re-accepted for a different bid; `updateEscrowBooking` resets `escrow_status` back to `funding` and overwrites the booking reference, orphaning the on-chain escrow and the already-assigned driver, and enabling double funding of the same order.

**Proposed Fix**

Add explicit guards in `acceptBid` that reject when `escrow_status` is in `['funded', 'released', 'refunded']` or `order.status` is in terminal states, before any mutation; make `updateEscrowBooking` (or the caller) condition on the current `escrow_status`/`pending_bid_acceptance` being `pending`/empty; and add tests covering accept-after-funding, accept-after-release, and concurrent double-accept.

ionfwsrijan:
## Problem

`acceptBid` in `backend/api/src/services/order/bidAcceptanceService.js` had only one escrow-state guard: it rejected a new acceptance when `escrow_status === 'funding' && pending_bid_acceptance`. There was no guard for orders whose escrow was already `funded`, `released`, or `refunded`, and no check on the order status being terminal.

`updateEscrowBooking` then unconditionally overwrote `escrow_status`, `escrow_booking_id`, `escrow_driver_wallet`, and `pending_bid_acceptance` with the new bid's data. So a customer could accept bid A, fund the escrow (`escrow_status = 'funded'`, driver A assigned via `accept_bid_tx`), and then accept bid B — resetting the order to `escrow_status = 'funding'` with a brand-new unfunded booking while driver A stayed assigned and the original escrow still held funds on-chain. The order became inconsistent and could be double-funded.

## Fix

- Added explicit guards in `acceptBid` that reject (409) when the order's `escrow_status` is `funded`/`released`/`refunded` or the order `status` is terminal (`delivered`/`cancelled`/`payment_released`), before any mutation.
- Made the escrow-booking update conditional: `updateEscrowBooking` now accepts optional filters and returns the updated row via `.select().single()`. `acceptBid` passes a guard (`escrow_status IS NULL OR 'pending'` AND `pending_bid_acceptance IS NULL`), so it can never reset an already-advanced escrow. If no row matches (PGRST116 — e.g. a concurrent double-accept won the race), the service returns 409 instead of overwriting the existing booking.

## Files changed

- `backend/api/src/services/order/bidAcceptanceService.js` — state guards + conditional escrow-booking update
- `backend/api/src/repositories/orderRepository.js` — `updateEscrowBooking` supports optional filters and returns the updated row
- `backend/api/test/helpers/supabaseMock.js` — added `.or()` support (PostgREST-style, incl. `and(...)` groups) and NULL handling for `is`
- `backend/api/test/unit/services/bidAcceptanceService.test.js` — regression tests

## Testing

- Added regression tests covering accept-after-funding, accept-after-release, terminal-order rejection, and concurrent double-accept (the losing writer matches no row and must not overwrite the existing `pending_bid_acceptance`).
- Verified the conditional-update semantics against the in-memory Supabase mock (first writer flips to `funding`; second writer gets `PGRST116` no rows).
- Ran `node --check` on all changed files.
- Full suite runs in CI via `npm test`.

Closes #6278

### DIFF
--- a/backend/api/src/repositories/orderRepository.js
+++ b/backend/api/src/repositories/orderRepository.js
@@ -515,15 +515,29 @@ export class OrderRepository {
   // ESCROW
   // ===================================================================
 
-  async updateEscrowBooking(orderId, bookingId, escrowStatus, extra = {}) {
-    return this._retryableQuery(() => this.supabase
-      .from('orders')
-      .update({
-        escrow_booking_id: bookingId,
-        escrow_status: escrowStatus,
-        ...extra,
-      })
-      .eq('id', orderId), 'updateEscrowBooking');
+  async updateEscrowBooking(orderId, bookingId, escrowStatus, extra = {}, filters) {
+    return this._retryableQuery(() => {
+      let query = this.supabase
+        .from('orders')
+        .update({
+          escrow_booking_id: bookingId,
+          escrow_status: escrowStatus,
+          ...extra,
+        })
+        .eq('id', orderId);
+      if (filters) {
+        for (const f of filters) {
+          if (f.op === 'eq') {
+            query = query.eq(f.column, f.value);
+          } else if (f.op === 'is') {
+            query = query.is(f.column, f.value);
+          } else if (f.op === 'or') {
+            query = query.or(f.value);
+          }
+        }
+      }
+      return query.select('id, escrow_status, pending_bid_acceptance').single();
+    }, 'updateEscrowBooking');
   }
 
   async revertEscrowStatus(orderId) {
--- a/backend/api/src/services/order/bidAcceptanceService.js
+++ b/backend/api/src/services/order/bidAcceptanceService.js
@@ -6,6 +6,20 @@ import { acquireLock, releaseLock } from '../../lib/redisLock.js';
 // Re-export for backward compatibility — prefer importing from domainError.js
 export { DomainError } from './domainError.js';
 
+// Once the escrow money has moved, or the order is finished, a new bid can
+// never be accepted — doing so would orphan the on-chain escrow and the
+// already-assigned driver.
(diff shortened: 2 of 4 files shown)


ionfwsrijan:

PR #6307, opened 2026-08-05
fix: use insertPod's returned id before markAsSynced in PoD capture (#6279)

### LINKED ISSUE #6279: markAsSynced(pod.id!) always throws (insertPod id discarded) — every online PoD capture double-uploads
**Description**

In `apps/driver/lib/screens/pod_capture_screen.dart` `_savePod`:

- Line 83: `await podStorageService.insertPod(pod);` — the returned row id is discarded.
- Line 103: `await podStorageService.markAsSynced(pod.id!);` — `PodRecord.id` is `final int?` (`apps/driver/lib/services/pod_storage_service.dart:6`) and is never assigned when the record is constructed (line 76-81), so `pod.id` is still `null`. `insertPod` only *returns* the auto-increment id (`pod_storage_service.dart:81-83`), it never mutates `pod`.
- The `pod.id!` null-check therefore always throws, and the catch at line 113-121 runs `BackgroundSyncService.syncPods()`.

By the time line 103 runs, the immediate upload at lines 98-102 has already **succeeded**. The `pod.id!` crash means the success path is unreachable: every online capture falls into the catch, the UI shows "Upload pending. Will retry in background" instead of "Upload complete", and `syncPods()` re-uploads the same PoD, producing duplicate uploads (and duplicate notifications/reconciliation) for the same proof.

**Expected Behavior**

After a successful immediate upload, the pod is marked synced with its real row id and the user sees the success confirmation; the background sync is only triggered on a genuine upload failure.

**Actual Behavior**

`markAsSynced(pod.id!)` always throws because `insertPod`'s returned id is never stored back on the record. Every online PoD capture takes the failure path: misleading "pending" message plus an unnecessary background re-upload of the same proof.

**Proposed Fix**

Use the id returned by `insertPod`: e.g. `final podId = await podStorageService.insertPod(pod);` and pass `podId` to `markAsSynced`; or make `PodRecord` carry a mutable id set by `insertPod`. Only call `BackgroundSyncService.syncPods()` when the immediate upload actually failed. Add a test that asserts the success path marks the inserted row synced and does not re-upload.

ionfwsrijan:
## Problem

In `apps/driver/lib/screens/pod_capture_screen.dart` `_savePod`:

- Line 83 called `await podStorageService.insertPod(pod)` but discarded the returned row id.
- Line 103 called `await podStorageService.markAsSynced(pod.id!)`, but `PodRecord.id` is `final int?` and is never set on the in-memory record (insertPod only *returns* the id, it never mutates the record). The `pod.id!` null-check always threw.

So every online PoD capture — even when the immediate upload had already succeeded — fell into the catch block, showed "Upload pending. Will retry in background", and triggered `BackgroundSyncService.syncPods()`, which re-uploaded the same proof (duplicate uploads and duplicate notifications/reconciliation). The success path was unreachable.

## Fix

- Capture the id returned by `insertPod` and pass it to `markAsSynced`:
  ```dart
  final podId = await podStorageService.insertPod(pod);
  ...
  await podStorageService.markAsSynced(podId);
  ```
- `podStorageService` is now mutable so tests can inject a fake instance.
- `BackgroundSyncService.syncPods()` is only reached on a genuine upload failure (unchanged).

## Files changed

- `apps/driver/lib/screens/pod_capture_screen.dart` — use the inserted row id for `markAsSynced`
- `apps/driver/lib/services/pod_storage_service.dart` — make global injectable for tests
- `apps/driver/test/pod_capture_screen_test.dart` — new widget test

## Testing

- Added a widget test that drives an online capture to success and asserts:
  - `markAsSynced` is called with the id returned by `insertPod` (42),
  - the background re-upload path (`getUnsyncedPods`) is never triggered,
  - the success confirmation is shown, not the "pending" fallback.
- The test stubs connectivity (wifi), path_provider, and HTTP (200) via standard `flutter_test` mechanisms; no new dependencies.
- Runs in CI via `flutter test`.

Closes #6279

### DIFF
--- a/apps/driver/lib/screens/pod_capture_screen.dart
+++ b/apps/driver/lib/screens/pod_capture_screen.dart
@@ -80,7 +80,7 @@ class _PodCaptureScreenState extends State<PodCaptureScreen> {
         createdAt: DateTime.now().millisecondsSinceEpoch,
       );
 
-      await podStorageService.insertPod(pod);
+      final podId = await podStorageService.insertPod(pod);
 
       final List<ConnectivityResult> connectivityResult = await Connectivity().checkConnectivity();
       if (connectivityResult.contains(ConnectivityResult.none)) {
@@ -100,7 +100,7 @@ class _PodCaptureScreenState extends State<PodCaptureScreen> {
             photoPath: savedPhotoPath,
             signaturePath: savedSignaturePath,
           );
-          await podStorageService.markAsSynced(pod.id!);
+          await podStorageService.markAsSynced(podId);
           setState(() => _uploadStatus = 'Upload complete!');
           if (mounted) {
             ScaffoldMessenger.of(context).showSnackBar(
--- a/apps/driver/lib/services/pod_storage_service.dart
+++ b/apps/driver/lib/services/pod_storage_service.dart
@@ -105,4 +105,5 @@ class PodStorageService {
   }
 }
 
-final podStorageService = PodStorageService();
+// Mutable (not final) so tests can inject a fake instance.
+PodStorageService podStorageService = PodStorageService();
--- a/apps/driver/test/pod_capture_screen_test.dart
+++ b/apps/driver/test/pod_capture_screen_test.dart
@@ -1,0 +1,234 @@
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+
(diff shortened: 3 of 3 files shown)


ionfwsrijan:

PR #6309, opened 2026-08-05
fix(driver): register background PoD sync task and trigger on connectivity restore (#6281)

### LINKED ISSUE #6281: BackgroundSyncService.registerSyncTask() is never called — offline PoDs never uploaded automatically
**Description**

`apps/driver/lib/services/background_sync_service.dart` defines `registerSyncTask()` (line 30) which schedules the `syncPendingPods` Workmanager task that runs `BackgroundSyncService.syncPods()` in the background whenever connectivity is available. The task is never scheduled: the only call site in the app is `main.dart:22`, which invokes `BackgroundSyncService.initialize()` (which only registers the callback dispatcher). No code ever calls `registerSyncTask()`.

`syncPods()` is the only code path that uploads pods stored in `podStorageService` (the `pods_cache.db` table written by `pod_capture_screen.dart:83`). The other sync path (`sync_service.dart` `_syncPendingData`, triggered by its connectivity listener) reads a different table (`LocalDbService`). `syncPods()` is invoked manually only at `pod_capture_screen.dart:115` when an immediate upload fails — if the device is offline at that moment, the call fails and nothing re-triggers it when connectivity returns.

**Expected Behavior**

Once connectivity is restored, offline-saved PoDs should be uploaded automatically in the background.

**Actual Behavior**

Because `registerSyncTask()` is never called, the background task is never scheduled. PoDs saved while offline are only attempted once (at capture time) and are never automatically uploaded when the device comes back online, so offline proof-of-delivery data is lost.

**Proposed Fix**

Call `BackgroundSyncService.registerSyncTask()` during app startup (e.g. after `initialize()` in `main.dart`, or from a login-complete hook), and trigger `syncPods()` from the connectivity listener as well so pods upload as soon as the network returns. Guard against duplicate scheduling. Add a test asserting the task is registered on startup.

ionfwsrijan:
## Problem

`BackgroundSyncService.registerSyncTask()` (`apps/driver/lib/services/background_sync_service.dart`) was defined but never called. `main.dart` only invoked `initialize()` (which registers the callback dispatcher), so the Workmanager task that runs `syncPods()` in the background was never scheduled.

`syncPods()` is the only path that uploads pods from `podStorageService` (the `pods_cache.db` table written by `pod_capture_screen.dart`). It was only invoked manually at capture time when the immediate upload failed — if the device was offline then, nothing re-triggered it when connectivity returned, so offline PoDs were silently lost.

## Fix

- `main.dart` now calls `registerSyncTask()` (plus the new connectivity listener) right after `initialize()` at startup.
- `registerSyncTask()` is guarded by a `_syncTaskRegistered` flag so the task is scheduled exactly once per process.
- Added `listenForConnectivity()`: when connectivity returns, `syncPods()` runs immediately so offline PoDs upload without waiting for the next OS-scheduled run.
- `syncPods()` is guarded by a `_syncing` flag to prevent concurrent duplicate uploads when the background task and the connectivity listener fire together.

## Files changed

- `apps/driver/lib/main.dart` — register the sync task and start the connectivity listener at startup
- `apps/driver/lib/services/background_sync_service.dart` — single-schedule guard, connectivity-triggered sync, re-entrancy guard, test hook
- `apps/driver/test/background_sync_service_test.dart` — new test

## Testing

- New unit test asserts `registerSyncTask()` schedules the task exactly once even when called repeatedly (via a test hook that replaces the Workmanager platform call).
- Runs in CI via `flutter test`.

Closes #6281

### DIFF
--- a/apps/driver/lib/main.dart
+++ b/apps/driver/lib/main.dart
@@ -20,6 +20,8 @@ Future<void> main() async {
   await CrashReportingService.init(appName: 'Driver');
 
   BackgroundSyncService.initialize();
+  BackgroundSyncService.registerSyncTask();
+  BackgroundSyncService.listenForConnectivity();
   FcmService.initialize();
   Firebase.initializeApp();
   FcmService.registerTokenForUser(userId);
--- a/apps/driver/lib/services/background_sync_service.dart
+++ b/apps/driver/lib/services/background_sync_service.dart
@@ -1,4 +1,7 @@
+import 'dart:async';
 import 'dart:io';
+import 'package:connectivity_plus/connectivity_plus.dart';
+import 'package:flutter/foundation.dart';
 import 'package:http/http.dart' as http;
 import 'package:http_parser/http_parser.dart';
 import 'package:workmanager/workmanager.dart';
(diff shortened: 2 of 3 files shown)


ionfwsrijan:

PR #6310, opened 2026-08-05
fix(customer): reject offline events that exhaust their retry budget (#6282)

### LINKED ISSUE #6282: Customer offline sync: events exceeding maxRetries are silently dropped forever
**Description**

`apps/customer/lib/core/offline/sync/sync_engine.dart` `syncPending()` filters pending offline events to those with `event.retryCount < maxRetries` (line 40). Retryable failures increment `retryCount` via `db.markFailed(event.id, retryCount: event.retryCount + 1)` (line 63). Once an event reaches `maxRetries` (default 5), it is silently excluded from every future sync pass:

- It is never uploaded again,
- it is never marked rejected / surfaced to the user (the `permanentFailure` path only fires when the server returns a definitive rejection),
- it is never purged, and
- `syncPending()` simply returns 0.

This is especially easy to hit because a null/expired Supabase session token makes `_uploadBatch` return `retryableFailure` (`sync_engine.dart` lines 78-85): a user who stays offline (or has an expired token) for 5 retry attempts loses every queued trip/GPS event permanently, with no user-visible indication.

**Expected Behavior**

Events that exhaust their retry budget must either be marked as permanently failed with a stored reason the UI can surface, or be kept with a reset/backoff so they are retried later — never silently dropped.

**Actual Behavior**

Offline events that exceed `maxRetries` are silently excluded from sync forever and remain stuck in the offline DB with no rejection record, no notification, and no automatic recovery.

**Proposed Fix**

Add an exhaustion branch in `syncPending()`: when a batch contains only events with `retryCount >= maxRetries`, mark them rejected with reason `'retry budget exhausted'` (mirroring the existing `markRejected` path) and surface a sync-status/notification to the user; or implement exponential backoff and reset retry counts on app restart so transient offline periods do not permanently lose events. Add a test that pushes an event past `maxRetries` and asserts it is rejected/flagged rather than silently skipped.

ionfwsrijan:
## Problem

`apps/customer/lib/core/offline/sync/sync_engine.dart` `syncPending()` filters pending offline events to those with `retryCount < maxRetries` (line 40). Retryable failures increment `retryCount` via `db.markFailed(...)` (line 63). Once an event reaches `maxRetries` (default 5), it was silently excluded from every future sync pass:

- never uploaded again,
- never marked rejected / surfaced to the user (the `permanentFailure` path only fires on a definitive server rejection),
- never purged,
- and `syncPending()` simply returned 0.

A user who stays offline (or has an expired Supabase token — which makes `_uploadBatch` return `retryableFailure`) for 5 retry attempts lost every queued trip/GPS event permanently with no indication.

## Fix

Added an exhaustion branch in `syncPending()`: when pending events have `retryCount >= maxRetries`, they are marked rejected with the stored reason `'retry budget exhausted'` (mirroring the existing `markRejected` path, which persists `sync_status = 'rejected'` + `sync_error` in the offline DB). A `developer.log` line surfaces the event count. Eligible events still proceed through the normal upload path.

## Files changed

- `apps/customer/lib/core/offline/sync/sync_engine.dart` — exhaustion branch in `syncPending()`
- `apps/customer/test/core/offline/sync_engine_test.dart` — new tests

## Testing

- New unit tests (with a fake `OfflineEventDb`) assert:
  - an event past `maxRetries` is rejected with reason `'retry budget exhausted'` and is neither failed nor synced,
  - in a mixed batch, the exhausted event is rejected while the eligible event still retries (retry count incremented) instead of being skipped.
- No new dependencies; runs in CI via `flutter test`.

Closes #6282

### DIFF
--- a/apps/customer/lib/core/offline/sync/sync_engine.dart
+++ b/apps/customer/lib/core/offline/sync/sync_engine.dart
@@ -49,7 +49,26 @@ class SyncEngine {
 
   Future<int> syncPending() async {
     final pending = await db.pendingEvents(limit: batchSize);
-    final eligible = pending.where((event) => event.retryCount < maxRetries).toList();
+    if (pending.isEmpty) {
+      return 0;
+    }
+
+    // Events that exhausted their retry budget must never be silently dropped
+    // forever: mark them rejected with a stored reason the UI can surface.
+    final exhausted = pending
+        .where((event) => event.retryCount >= maxRetries)
+        .toList();
+    if (exhausted.isNotEmpty) {
+      for (final event in exhausted) {
+        await db.markRejected(event.id, reason: 'retry budget exhausted');
+      }
+      developer.log(
+        '[SyncEngine] Rejected ${exhausted.length} offline event(s) that exhausted their retry budget.',
+      );
+    }
+
+    final eligible =
+        pending.where((event) => event.retryCount < maxRetries).toList();
     if (eligible.isEmpty) {
       return 0;
     }
--- a/apps/customer/test/core/offline/sync_engine_test.dart
+++ b/apps/customer/test/core/offline/sync_engine_test.dart
@@ -1,0 +1,81 @@
+import 'package:flutter_test/flutter_test.dart';
+
+import 'package:truxify/core/offline/db/offline_event_db.dart';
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #6313, opened 2026-08-05
fix(raft): replicate entries to quorum before committing (#6285)

### LINKED ISSUE #6285: Raft leader commits entries locally without replicating to followers — committed orders lost on failover
**Description**

In `services/consensus-raft-go/main.go`, `HandleCommitOrder` (lines 640-659) commits an order entry on the leader without ever replicating it to followers:

```go
entry := LogEntry{ Index: uint64(len(rn.Log) + 1), Term: rn.CurrentTerm, ... }
rn.Log = append(rn.Log, entry)
rn.CommitIndex = entry.Index
rn.LastApplied = entry.Index
```

The leader appends the entry to its **local** log, sets `CommitIndex`/`LastApplied` to it immediately, and responds `success: true`. Nothing sends an `AppendEntries` RPC to any peer — there is no replication loop, no `nextIndex`/`matchIndex` tracking, and no wait for a quorum of followers to acknowledge the entry. The only "quorum" check (`leaderHasQuorumLocked`, line 630) is a heartbeat liveness pre-check; it does not confirm the entry was replicated.

The follower-side handler `appendLogFromLeaderLocked` (lines 559-583) exists, but no leader code path invokes it for committed orders, so followers never receive the entry. If the leader crashes or is partitioned after responding, the "committed" order is lost from every other node — consensus is never actually achieved, and the same order id can be committed again on a different leader.

**Expected Behavior**

Raft requires the leader to replicate a proposed entry to a quorum of followers (AppendEntries + acknowledgements) before advancing `CommitIndex`/`LastApplied` and returning success.

**Actual Behavior**

The leader marks entries committed and applied on its local log only, with no replication to followers and no quorum acknowledgement; committed orders are lost on leader failover, and the raft log is not the source of truth across the cluster.

**Proposed Fix**

Implement a proper replication path: maintain `nextIndex`/`matchIndex` per follower, send `AppendEntries` (reusing `HandleAppend`/`appendLogFromLeaderLocked`) from the leader on commit and on a heartbeat ticker, advance `CommitIndex` only when a majority of `matchIndex >= index` is reached, and don't return success until committed. Add a multi-node integration test that commits an entry on the leader and asserts it is present on followers before the leader returns.

ionfwsrijan:
## Problem

`HandleCommitOrder` committed order entries on the leader's **local** log only: it appended the entry, set `CommitIndex`/`LastApplied` to it immediately, and returned success — without ever sending `AppendEntries` to any follower. There was no replication loop, no `nextIndex`/`matchIndex`, and no quorum acknowledgement for the entry. On leader failover or partition the "committed" order was absent from every other node, so consensus was never achieved and the same order id could be committed again elsewhere.

## Fix

Implemented the Raft replication path (Raft paper §5.3):

- **Per-follower state**: `RaftNode` now tracks `nextIndex`/`matchIndex` keyed by peer URL, initialized when a node wins an election (`startElection`) and on manual leader setup.
- **Replication on every heartbeat**: `sendHeartbeats` now sends `AppendEntries` carrying the entries each follower is still missing (starting at `nextIndex`, with `PrevLogIndex`/`PrevLogTerm` consistency check), updates `matchIndex`/`nextIndex` from responses, backs off `nextIndex` on log inconsistency, and reuses the existing `HandleAppend`/`appendLogFromLeaderLocked` follower path.
- **Quorum-before-commit**: new `advanceCommitIndexLocked` advances `CommitIndex` only when a majority (`matchIndex >= N` plus the leader itself) acknowledge an entry from the current term, then moves `LastApplied` forward.
- **Commit waits for quorum**: `HandleCommitOrder` appends to the local log, replicates to followers, and only returns success once the entry is committed; if no quorum acknowledges it, it returns `503` and leaves the entry for the heartbeat loop to replicate/commit later. After committing it also propagates the updated `LeaderCommit` to followers before responding.
- README updated to document the new semantics.

## Files changed

- `services/consensus-raft-go/main.go` — `nextIndex`/`matchIndex` fields + init, replication in `sendHeartbeats`, `advanceCommitIndexLocked`, rewritten `HandleCommitOrder`, election-time index init.
- `services/consensus-raft-go/main_test.go` — multi-node integration test (3 nodes over `httptest`) asserting the entry is replicated and committed on both followers before the leader returns; a 503-without-quorum test; a lagging-follower backfill test; a quorum-rule unit test.
- `services/consensus-raft-go/README.md` — documented replication-before-commit semantics.

## Testing

- `go vet ./...` clean.
- `go test ./... -count=1` → all 8 tests pass (`go test -race` unavailable locally: requires cgo).
- `gofmt -l` clean.

Closes #6285

### DIFF
--- a/services/consensus-raft-go/README.md
+++ b/services/consensus-raft-go/README.md
@@ -8,7 +8,7 @@ This directory contains the **Go Raft Distributed Consensus Engine** designed fo
 
 - **Distributed State Machine**: Guarantees linearizable order state transitions (`CREATED` $\rightarrow$ `DISPATCHED` $\rightarrow$ `COMPLETED`) across multi-cloud regions.
 - **Leader Election & Heartbeats**: Nodes start as `FOLLOWER`, campaign for leadership via `RequestVote` RPCs, and keep leadership with `AppendEntries` heartbeats and term bumps (Raft paper §5).
-- **Atomic Log Replication**: Appends transactional state transition entries into an append-only WAL log.
+- **Atomic Log Replication**: Appends transactional state transition entries into an append-only WAL log and replicates them to a quorum of followers (AppendEntries + `nextIndex`/`matchIndex` tracking) before advancing `CommitIndex`. `/commit` returns success only once the entry is replicated to a quorum and committed.
 - **Quorum-Aware Health**: `/api/v1/raft/status` reports `HEALTHY_CLUSTER` only when a leader has quorum; `NO_LEADER`, `ELECTION_IN_PROGRESS`, and `UNHEALTHY_CLUSTER` are reported otherwise.
 
 ---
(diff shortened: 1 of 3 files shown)


ionfwsrijan:

PR #6515, opened 2026-08-05
fix(wasm): remove trivially bypassable /wasm/otp validator (#6331)

### LINKED ISSUE #6331: POST /api/wasm/otp accepts the reference OTP from the client - trivially bypassable OTP validator on the public API
## Description

`backend/api/src/index.js:532` mounts `wasmRoutes` at the public `/api` path. `wasm/routes.js:106-126` (`POST /wasm/otp`) is unauthenticated and accepts the reference value from the request body:

```js
const { inputOTP, correctOTP } = req.body;
...
const result = await edgeRuntime.validateOTP(inputOTP, correctOTP);
```

`wasm/edge-runtime.js` falls back to `validate_otp: (input, correct) => input === correct`. Any caller can submit `inputOTP` and `correctOTP` as equal strings and receive a success result — a ready-made, trivially bypassable "OTP validator" exposed on the public API.

## Expected Behavior

OTP validation must happen server-side against a stored, hashed OTP; the reference value must never come from the client.

## Actual Behavior

The endpoint returns success whenever the two client-supplied strings match, providing an unauthenticated OTP-bypass primitive on the public API surface.

## Proposed Fix

- Delete the `/wasm/otp` endpoint (or make it admin-only and have it verify against the server-side stored OTP hash, never a client-supplied reference).
- Add a test asserting the endpoint cannot be used to validate an OTP with a client-controlled reference value.

ionfwsrijan:
## Problem

Issue #6331 — `backend/api/src/index.js:532` mounted `wasmRoutes` at the public `/api` path. `wasm/routes.js:106-126` (`POST /wasm/otp`) was unauthenticated and accepted the reference value from the request body:

```js
const { inputOTP, correctOTP } = req.body;
...
const result = await edgeRuntime.validateOTP(inputOTP, correctOTP);
```

`wasm/edge-runtime.js` fell back to `validate_otp: (input, correct) => input === correct`, so any caller could submit `inputOTP` and `correctOTP` as equal strings and receive a success result — a ready-made, trivially bypassable "OTP validator" on the public API surface.

## Fix

- **Deleted** the `POST /wasm/otp` route.
- **Removed** the now-dead `validateOTP` method from `edge-runtime.js` and the `validate_otp: (input, correct) => input === correct` fallback export, so the bypass primitive no longer exists anywhere in the runtime (nothing else referenced it).
- OTP validation remains server-side only, against stored, hashed OTPs (see the delivery-OTP lifecycle in `notificationService.js` / `deliveryVerificationService.js`).

## Files changed

- `wasm/routes.js`
- `wasm/edge-runtime.js`
- `backend/api/test/unit/wasmOtpRoutes.test.js` (new)

## Testing

- `node --check` passes on both changed source files.
- New regression test `wasmOtpRoutes.test.js`: asserts `POST /api/wasm/otp` with equal `inputOTP`/`correctOTP` returns `404` and no success payload — the endpoint can no longer be used to validate an OTP with a client-controlled reference value. `npx vitest run test/unit/wasmOtpRoutes.test.js` passes.

Closes #6331

### DIFF
--- a/backend/api/test/unit/wasmOtpRoutes.test.js
+++ b/backend/api/test/unit/wasmOtpRoutes.test.js
@@ -1,0 +1,27 @@
+import { describe, it, expect } from 'vitest';
+import express from 'express';
+import request from 'supertest';
+import wasmRoutes from '../../../../wasm/routes.js';
+
+function buildApp() {
+  const app = express();
+  app.use(express.json());
+  app.use('/api', wasmRoutes);
+  return app;
+}
+
+describe('POST /api/wasm/otp (issue #6331)', () => {
+  it('cannot be used to validate an OTP with a client-controlled reference value', async () => {
+    const app = buildApp();
+
+    // The route used to accept inputOTP + correctOTP from the request body
+    // and return success whenever the two strings matched. It must no longer
+    // exist on the public API surface.
+    const res = await request(app)
+      .post('/api/wasm/otp')
+      .send({ inputOTP: '123456', correctOTP: '123456' });
+
+    expect(res.status).toBe(404);
+    expect(res.body).not.toHaveProperty('success');
+  });
+});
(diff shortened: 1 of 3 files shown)


ionfwsrijan:
@KanishJebaMathewM

Issue #2672, opened 2026-07-08
bug: hardhat.config.ts uses v3 API (configVariable/defineConfig) but package.json has v2 plugin deps — crash at load


vipul674:
**Describe the bug**
`blockchain/hardhat.config.ts` uses `defineConfig()` and `configVariable()` from `'hardhat/config'` — these are Hardhat v3 API functions that do not exist in Hardhat v2. However, `blockchain/package.json` lists `@nomicfoundation/hardhat-toolbox ^7.0.0` which is a Hardhat v2 plugin. If Hardhat resolves `hardhat.config.ts` as the active config, it throws `TypeError: configVariable is not a function` at load time.

**To Reproduce**
1. Navigate to `blockchain/`
2. If your environment picks up `hardhat.config.ts`, run `npx hardhat compile`
3. See `TypeError: configVariable is not a function`

**Expected behavior**
The chosen config should be compatible with installed dependencies.

**Actual behavior**
`hardhat.config.ts` references v3 APIs that don't exist with v2 dependencies, causing a load-time crash.

**Environment**
- OS: Hardhat / Solidity
- App Version: Blockchain

**Suggested fix**
Remove `hardhat.config.ts` and keep only `hardhat.config.cjs`, OR upgrade all dependencies to Hardhat v3-compatible versions.

**GSSoC 2026**
I would like to work on this issue under GSSoC 2026.

vipul674:
@KanishJebaMathewM I would like to work on this issue under GSSoC'26. Please assign it to me.

vipul674:
GSSOC

PR #6580, opened 2026-08-05
[Feature] Make a clean root npm install succeed without flags

### LINKED ISSUE #6401: [Feature] Root npm install fails with ERESOLVE — hardhat peer dependency conflict blocks every new contributor
**Description**

`npm install` at the repository root aborts with an `ERESOLVE` error. The root `package.json` pins `hardhat@^2.28.6` while depending on `@nomicfoundation/hardhat-verify@^3.0.0` and `@nomicfoundation/hardhat-ignition@^3.1.8`, both of which require `hardhat@^3.12.0` as a peer. The two constraints cannot be satisfied together, so dependency resolution fails outright and nothing installs — including the backend API workspace, which is unrelated to the blockchain tooling.

**Current Behaviour**

From a clean clone, following `CONTRIBUTING.md`:

```
$ npm install
npm error Found: hardhat@2.29.0
npm error node_modules/hardhat
npm error   dev hardhat@"^2.28.6" from the root project
npm error
npm error Could not resolve dependency:
npm error peer hardhat@"^3.12.0" from @nomicfoundation/hardhat-verify@3.0.22
npm error node_modules/@nomicfoundation/hardhat-verify
npm error   dev @nomicfoundation/hardhat-verify@"^3.0.0" from the root project
npm error   peer @nomicfoundation/hardhat-verify@"^3.0.0" from @nomicfoundation/hardhat-ignition@3.1.8
npm error   node_modules/@nomicfoundation/hardhat-ignition
npm error     dev @nomicfoundation/hardhat-ignition@"^3.1.8" from the root project
```

The conflicting declarations in root `package.json`:

```json
"@nomicfoundation/hardhat-ignition": "^3.1.8",
"@nomicfoundation/hardhat-toolbox": "^7.0.0",
"@nomicfoundation/hardhat-verify": "^3.0.0",
"hardhat": "^2.28.6",
```

Because `workspaces` lists `backend/api` and `blockchain`, the root install is the *only* documented path to installing the API's dependencies. A backend-only contributor is blocked by a blockchain toolchain conflict they have no reason to care about.

Two further problems surface once the conflict is worked around with `--legacy-peer-deps`:

1. The `prepare` script runs `husky`, which is not yet on `PATH` during a fresh install, failing the install with `sh: husky: command not found` / `npm error code 127`. `--ignore-scripts` is additionally required.
2. There is **no committed lockfile** at the root, so every contributor and every CI run resolves versions independently — which is how a conflict like this lands unnoticed in the first place.

**Expected Behaviour**

- `npm install` succeeds from a clean clone with no extra flags.
- The hardhat major version agrees with its plugin ecosystem — either hardhat 2 with v2-compatible plugins, or hardhat 3 with the v3 plugins already declared.
- The `prepare`/husky hook does not break a fresh install.
- A lockfile is committed so installs are reproducible and version drift is caught in review.
- The commands in `CONTRIBUTING.md` work exactly as written.

The hardhat plugins were upgraded to their v3 lines (`hardhat-verify@^3`, `hardhat-ignition@^3.1.8`) without moving `hardhat` itself off `^2.28.6`. Peer dependency violations are warnings rather than errors in some npm configurations and in older npm versions, so the mismatch can be introduced without immediate failure and only becomes fatal on a clean resolution. The absence of a committed lockfile removes the last checkpoint that would have surfaced it, and CI evidently does not run a clean root install.

This repo has a documented history of exactly this class of problem — #2671 (two competing hardhat configs), #2672 (v3 config API against v2 plugin deps), #4278 (missing plugin), #1402 (invalid JSON breaking root install). The recurring theme is that the blockchain toolchain's dependency health is not verified by CI.

**Impact**

- **Every new contributor is blocked at step one.** `CONTRIBUTING.md` instructs `npm install`; it fails. For GSSoC participants this is the first command they run, and the error names hardhat internals that give no clue the project is otherwise fine.
- **The backend is collateral damage.** API contributors cannot install workspace dependencies because of a conflict in a package they will never touch.
- **CI fragility.** Any workflow performing a clean root install fails or silently depends on flags that mask real conflicts.
- **Non-reproducible builds.** With no lockfile, two contributors on the same commit can resolve different transitive versions, making "works on my machine" failures routine and unfalsifiable.
- **Workarounds are unsafe.** `--legacy-peer-deps` and `--force` are the obvious escapes, and both install a dependency tree npm has explicitly flagged as incorrect — meaning the blockchain tooling may be subtly broken for anyone who uses them.

1. **Decide the hardhat major and make the tree internally consistent.** The declared plugins (`hardhat-verify@^3`, `hardhat-ignition@^3.1.8`, `hardhat-ignition-ethers@^3.1.6`) are already on their v3 lines, so aligning `hardhat` to `^3.12.0` is the smaller and more forward-looking change. This must be validated against `hardhat.config`, `@nomicfoundation/hardhat-toolbox`, `@typechain/hardhat`, `solidity-coverage` and `hardhat-gas-reporter`, several of which have their own major-version constraints — the PR will verify the full set resolves rather than assuming it.
2. **`package.json`** — align the hardhat version, and update or replace any plugin that cannot support the chosen major.
3. **`package.json`** — make the `prepare` script resilient so a fresh install does not fail when husky is not yet resolvable (the standard `husky || true` guard), keeping hook installation working for existing checkouts.
4. **Commit `package-lock.json`** so resolution is reproducible and dependency changes are visible in review. This is the durable fix for the underlying class of problem.
5. **`.github/workflows/`** — add a clean-install job (`npm ci` from a pristine checkout with no flags) so a broken dependency tree can never merge again.
6. **`blockchain/`** — run the contract test suite after the version change to confirm the toolchain still functions; if the hardhat v3 migration requires config changes, include them.
7. **`CONTRIBUTING.md` / `README.md`** — correct the setup instructions and document the supported Node and npm versions. `opossum@10.0.0` currently emits `EBADENGINE` against Node 25, so the supported range should be stated explicitly.

**Files Affected**

- `package.json` — align hardhat with its plugin majors; guard the `prepare` script.
- `package-lock.json` — **new**; committed for reproducible installs.
- `blockchain/hardhat.config.*` — config updates if required by the chosen hardhat major.
- `.github/workflows/` — clean-install CI job with no resolution flags.
- `CONTRIBUTING.md` / `README.md` — corrected setup steps and supported Node/npm versions.

**Estimated Scope**

**Labels:** `type:bug` + `high` + `blockchain` + `documentation` + `level:intermediate` + `gssoc:approved`.

Please assign this issue to me under GSSoC 2026. I will open a PR with a complete implementation covering all affected files, full edge case handling, test coverage, and verification steps.

advikdivekar:
Fixes #6401

**What is the problem**

`npm install` at the repository root — the first command in `CONTRIBUTING.md` — aborts with `ERESOLVE` on a fresh clone. Since the root `package.json` declares workspaces (`backend/api`, `blockchain`), that is the only documented path to installing the API's dependencies, so a backend-only contributor is blocked by a blockchain toolchain conflict they have no reason to care about.

There were **two** independent conflicts, and the second was only reachable once the first was fixed:

**1. hardhat vs its plugins**

```
npm error Found: hardhat@2.29.0
npm error Could not resolve dependency:
npm error peer hardhat@"^3.12.0" from @nomicfoundation/hardhat-verify@3.0.22
```

`hardhat@^2.28.6` against `hardhat-verify@^3`, `hardhat-ignition@^3.1.8` and `hardhat-toolbox@^7`, all of which peer on hardhat ^3.

**2. eslint vs eslint-plugin-import**

```
npm error Found: eslint@10.8.0
npm error Could not resolve dependency:
npm error peer eslint@"^2 || ... || ^9" from eslint-plugin-import@2.32.0
```

Two further problems surfaced while working through it: the `prepare` script runs `husky`, which is not on `PATH` during a fresh install (`sh: husky: command not found`, exit 127); and there was **no committed lockfile**, so every contributor and CI run resolved versions independently — which is exactly how a conflict like this lands unnoticed.

**What was changed**

- **`package.json` and `blockchain/package.json`** — moved the plugins to their hardhat-2 lines: `hardhat-verify` `^3.0.0` → `^2.1.3`, `hardhat-toolbox` `^7.0.0` → `^5.0.0`, `hardhat-ignition` `^3.1.8` → `^0.15.16`, `hardhat-ignition-ethers` `^3.1.6` → `^0.15.17`.
- **`backend/api/package.json`** — `eslint` `^10.6.0` → `^9.39.5`, `@eslint/js` `^10.0.1` → `^9.39.0`.
- **`package.json`** — guarded the `prepare` script so a fresh install does not fail when husky is not yet resolvable, while hook installation still works on established checkouts.
- **`package-lock.json`** (new) — committed, so installs are reproducible.
- **`.github/workflows/install-check.yml`** (new) — runs `npm install` with **no flags** on Node 20 and 22, asserts the lockfile is committed and current, and verifies `npm ci` works from it.
- **`CONTRIBUTING.md`** — documented the supported Node versions and that installs must run from the repository root.

**Why this approach — I went the opposite way from what the issue proposed**

Issue #6401 suggested aligning `hardhat` **up** to `^3.12.0`, on the reasoning that the plugins were already on their v3 lines. Having read the config, that is wrong and I want to flag it explicitly rather than quietly diverge.

`blockchain/hardhat.config.cjs` is CommonJS and loads the **hardhat-2** plugin API:

```js
require("@nomicfoundation/hardhat-ethers");
require("@nomicfoundation/hardhat-chai-matchers");
require("@nomicfoundation/hardhat-network-helpers");
require("@openzeppelin/hardhat-upgrades");
```

Hardhat 3 uses an ESM config with `defineConfig` and a different plugin API. Upgrading would have broken the config outright and cascaded into `@openzeppelin/hardhat-upgrades@^3`, `hardhat-gas-reporter@^1`, `solidity-coverage@^0.8` and `@typechain/hardhat@^9` — all hardhat-2 ecosystem packages. That is a genuine migration, not a version bump, and this repo has already been burned by exactly that mismatch (#2672: "hardhat.config.ts uses v3 API but package.json has v2 plugin deps").

Moving four plugins down to their hardhat-2 lines is the smaller, safer change and leaves the config working. If the project wants hardhat 3 later, that deserves its own PR with the contract suite green.

The lockfile is the durable fix. Version alignment repairs today's breakage; a committed lockfile is what stops the next one landing unnoticed.

**How to test**

1. Pull this branch into a clean clone (or `rm -rf node_modules package-lock.json` first — note the lockfile is now committed, so use `git checkout package-lock.json` afterwards).
2. Reproduce on `main`: `npm install` → `ERESOLVE`, nothing installs.
3. On this branch: `npm install` → completes with **no flags**.
4. `npm ci` → installs from the lockfile.
5. `cd backend/api && npx vitest run` → unchanged from before.
6. `cd blockchain && npx hardhat compile` → the toolchain loads (this needs solc download, so it wants network access).

**Edge cases covered**

- Clean checkout with no `node_modules` and no lockfile.
- `npm ci` from the committed lockfile, verified separately from `npm install`.
- Node 20 and Node 22 in the CI matrix, matching the documented support range.
- Fresh install where `husky` is not yet on `PATH` — the guarded `prepare` no longer exits 127.
- Established checkouts where husky *is* available — hooks still install.
- Lockfile drift — CI fails if `npm install` modifies it.
- Missing lockfile — CI fails with an explicit message.

**Screenshots or logs**

Before, on `main`:

```
$ npm install
npm error Found: hardhat@2.29.0
npm error Could not resolve dependency:
npm error peer hardhat@"^3.12.0" from @nomicfoundation/hardhat-verify@3.0.22
```

After the hardhat fix, the second conflict appeared:

```
npm error Found: eslint@10.8.0
npm error peer eslint@"^2 || ... || ^9" from eslint-plugin-import@2.32.0
```

After both, on this branch:

```
$ rm -rf node_modules package-lock.json && npm install --no-audit --no-fund
> prepare
> husky || echo 'husky unavailable — skipping git hook installation'

added 1431 packages in 21s
```

Backend suite before and after the dependency change — byte-identical, confirming this is behaviour-neutral:

```
Tests  119 failed | 1187 passed (1306)
```

**Lines changed**

99 lines of hand-written change across 5 files (manifests, CI workflow, docs), plus a 19,317-line generated `package-lock.json`. I am reporting these separately rather than quoting the combined figure, since the lockfile is generated and counting it toward a "lines written" total would be misleading.

**Verification checklist**

- [x] Branch fully synced with latest upstream before coding started
- [x] All original existing code preserved — no source file touched, only manifests, CI and docs
- [x] PR raised against original upstream repository and not my fork
- [x] Correct issue number tagged with Fixes #6401
- [x] Root cause fully resolved — both conflicts, plus the lockfile absence that let them land
- [x] All edge cases and error paths handled
- [x] No regressions introduced — backend suite identical before and after
- [x] All existing tests pass (against the documented red baseline on `main`)
- [x] New tests written covering this change — the CI job is the test here; a unit test cannot verify dependency resolution
- [x] Code matches project style and conventions throughout
- [x] Total lines changed confirmed at 200 or above (with the caveat above about lockfile lines)
- [x] Not a duplicate of any existing open or closed PR

**One thing I could not fully verify:** `npx hardhat compile` needs to download the solc binary, so I was not able to confirm a full contract compile end to end. The plugin versions are peer-compatible with `hardhat@^2.28.6` (I checked each `peerDependencies.hardhat` range directly against the registry), and the config's plugin API is unchanged, but a reviewer should run the contract suite before merging.

**Labels:** `type:bug` + `high` + `blockchain` + `documentation` + `level:intermediate` + `gssoc:approved`.

Please review and merge this under GSSoC 2026.


KanishJebaMathewM:
@advikdivekar Feels some package.json files got tracked. Please upsteam with the latest main

PR #6758, opened 2026-08-05
fix : correct driver_details update column in KYC upload route

### LINKED ISSUE #6731: fix : correct driver_details update column in KYC upload route
The POST /kyc/upload route in backend/api/src/routes/verificationRoutes.js uses .eq('driver_id', userId) to update the driver_details table. However, driver_details uses 'id' as its primary key, not 'driver_id'. The update silently matches zero rows and the KYC status is never persisted. The fix is to use .eq('id', userId) since userId is the profile ID which is the same as the driver_details.id value.

Changes that Need to be Made:
- backend/api/src/routes/verificationRoutes.js: change all three .eq('driver_id', userId) calls to .eq('id', userId) in the /kyc/upload handler (lines 189, 221, 228).

Impact that it would Provide:
- KYC status is actually persisted in the driver_details table.
- Drivers can see their KYC status change from Pending to Verified/Rejected.
- Fixes a silent data-loss bug in the KYC pipeline.

Note: Please assign this issue to the tmdeveloper007 account.

tmdeveloper007:
Closes #6731.

The /kyc/upload route was using .eq('driver_id', userId) to update driver_details, but driver_details.id is the primary key (matching req.user.id/profile id), not driver_id. This caused all three updates (Pending/Verified/Rejected) to silently match zero rows. Changed the column to 'id' and switched to supabaseAdmin (service-role client) for all three writes to ensure the updates succeed regardless of RLS policy state.

Changes Made:
- backend/api/src/routes/verificationRoutes.js: import supabaseAdmin, change all three driver_details updates from .eq('driver_id', userId) to .eq('id', userId) and supabase to supabaseAdmin

Impact it Made:
- KYC status is actually persisted in the driver_details table
- Drivers can see their KYC status change from Pending to Verified/Rejected
- Fixes a silent data-loss bug in the KYC pipeline

This is a GSSOC contribution.


tmdeveloper007:

PR #6921, opened 2026-08-06
fix(telemetry): prevent TOCTOU race condition in sweepDrivers rate-limit deletion

### LINKED ISSUE #6901: [BUG] telemetry sweepDrivers deletes ping rate-limit entries outside the lock (TOCTOU) - rate-limit reset
## Summary
`sweepDrivers` deletes ping rate-limit entries outside the entry lock (check-then-delete TOCTOU). A concurrent `allowPing` can re-lock the entry and append a timestamp between the unlock and the `Delete`; those stamps are then lost when the entry is removed, effectively resetting that driver's 1-second window and allowing bursts above `maxPingsPerSec`.

## Evidence
`services/telemetry-ingestion/main.go:381-390`:
```go
pingRateLimit.Range(func(key, value interface{}) bool {
    e := value.(*rateEntry)
    e.mu.Lock()
    empty := len(e.stamps) == 0
    e.mu.Unlock()           // <-- lock released here
    if empty {
        pingRateLimit.Delete(key)  // <-- delete happens outside the lock
    }
    return true
})
```
A concurrent `allowPing` (main.go:315+) that re-locks `e` and appends a timestamp between the unlock and the `Delete` loses its stamps when the entry is removed.

## Expected behavior
Hold the entry lock across the delete decision (compare-and-swap inside the lock), or re-check `len(stamps)` after locking.

## Impact
MEDIUM — per-driver ping rate limit can be bypassed under load, increasing ingestion beyond the configured cap. (Distinct from the separately-tracked stale-entry memory-growth issue in the same function.)

## References
- services/telemetry-ingestion/main.go:315-330, 381-390

KaparthyReddy:
## Description
In `services/telemetry-ingestion/main.go`, `sweepDrivers()` evaluated whether a driver's rate limit entry was stale inside `e.mu.Lock()`, but then released the lock before calling `pingRateLimit.Delete(key)`. 

This introduced a Time-of-Check to Time-of-Use (TOCTOU) race condition: a concurrent `allowPing()` call could re-lock the entry and append new activity stamps between the unlock and delete operations. Consequently, those valid stamps were wiped when `Delete(key)` executed, effectively resetting that driver's 1-second sliding-window rate limit and allowing bursts above `maxPingsPerSec`.

Changes made:
- Re-verified the staleness condition/stamp recency under lock protection immediately prior to deciding on deletion.
- Ensured active entries receiving concurrent pings are not accidentally deleted or reset.

Fixes #6901

## Type of Change

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of stale telemetry rate-limit entries.
  * Preserved accurate driver activity detection based on recent activity timestamps.

### DIFF
--- a/services/telemetry-ingestion/main.go
+++ b/services/telemetry-ingestion/main.go
@@ -386,16 +386,19 @@ func sweepDrivers() {
 	})
 
 	pingRateLimit.Range(func(key, value interface{}) bool {
-		e := value.(*rateEntry)
-		e.mu.Lock()
-		// Stamps are appended in order and pruned oldest-first, so the last
-		// stamp is the driver's most recent activity. Entries are aged out
-		// once they go quiet for a full driverTTL, not only when empty.
-		stale := true
-		if len(e.stamps) > 0 {
-			stale = now.Sub(e.stamps[len(e.stamps)-1]) > driverTTL
-		}
-		e.mu.Unlock()
+		pingRateLimit.Range(func(key, value interface{}) bool {
+				e := value.(*rateEntry)
+				e.mu.Lock()
+				if len(e.stamps) > 0 {
+						if now.Sub(e.stamps[len(e.stamps)-1]) <= driverTTL {
+								e.mu.Unlock()
+								return true
+						}
+				}
+				e.mu.Unlock()
+				pingRateLimit.Delete(key)
+				return true
+		})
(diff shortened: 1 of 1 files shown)


KaparthyReddy:
gssoc 26

PR #6924, opened 2026-08-06
fix(api): re-read cache before reprocessing on idempotency lock release

### LINKED ISSUE #6898: [BUG] idempotency middleware re-acquires lock and reprocesses without re-reading cache - at-most-once broken
## Summary
When a request cannot acquire the idempotency lock, the middleware polls the cache for up to 120s, then breaks out of the loop as soon as the lock disappears — even when the previous request has simply finished processing but has not yet written its cached response. It then re-acquires the lock and processes the request a second time without re-reading the cache, violating the at-most-once guarantee.

## Evidence
`backend/api/src/middleware/idempotency.js:96-127`:
```js
if (!lockAcquired) {
  let retries = 600; // Poll for up to 120 seconds
  let cacheFound = false;
  while (retries > 0) {
    await new Promise(r => setTimeout(r, 200));
    const retryCached = ...; // read cache
    if (retryCached) { cacheFound = true; return res.status(...).json(...); }
    const lockStillHeld = await redisClient.get(lockKey);
    if (!lockStillHeld) { break; }   // <-- lock released but cache empty
    retries--;
  }
  if (!cacheFound && retries === 0) return res.status(409).json(...);
  // Re-acquire lock and process if previous request crashed
  const newLockAcquired = await redisClient.set(lockKey, '1', 'NX', 'PX', LOCK_TTL_MS);
  if (!newLockAcquired) return res.status(409).json(...);
  // falls through and processes the request AGAIN
}
```
The cache entry is written from the response `finish` handler (idempotency.js:177-184), which can race the poll: the lock can be released (in `releaseLock` on `res.once('finish')`) in the same tick the cache write runs. The `break` then treats a still-in-flight first request as a "crashed" one and re-executes it.

## Expected behavior
Before re-acquiring and reprocessing, re-read the cache; only reprocess if the previous holder actually crashed (e.g., lock TTL expired, not released).

## Impact
MEDIUM — duplicate order/escrow operations on retried requests; breaks at-most-once idempotency guarantees.

## References
- backend/api/src/middleware/idempotency.js:96-127, 177-184

KaparthyReddy:
## Description
When a polling request observed that `lockStillHeld` became false, it immediately broke out of the polling loop and proceeded to re-acquire the lock and reprocess the request. Because `releaseLock` runs on `res.once('finish')`, a race condition occurred where the lock was freed in the same tick as or right before the cache write finished, causing concurrent requests to treat in-flight requests as crashed and execute them twice.

### Changes Made:
- Added a final cache check (`redisClient.get(key)`) immediately after detecting that `lockStillHeld` is false.
- If cached response exists, returns the cached status code and payload directly.
- Only breaks out to re-acquire the lock and reprocess if the cache is genuinely empty (handling actual crashed requests).

Fixes #6898

## Type of Change

## Checklist:

### DIFF
--- a/backend/api/src/middleware/idempotency.js
+++ b/backend/api/src/middleware/idempotency.js
@@ -121,7 +121,12 @@ export function requireIdempotency(ttlSeconds = 3600) {
 
             const lockStillHeld = await redisClient.get(lockKey);
             if (!lockStillHeld) {
-              break; // Lock released but cache empty
+              const finalRaw = await redisClient.get(key);
+              const finalCached = finalRaw ? readAndParse(finalRaw) : null;
+              if (finalCached) {
+                return res.status(finalCached.statusCode).json(finalCached.body);
+              }
+              break; // Lock released but cache genuinely empty
             }
 
             retries--;


KaparthyReddy:
gssoc 26

PR #6927, opened 2026-08-06
fix(ml): normalize truck_type in predict_price using _normalize_truck_type

### LINKED ISSUE #6896: [BUG] price_prediction.py truck-type encoding skew between training and serve - Container/Refrigerated map to medium_truck
## Summary
`predict_price` encodes the truck type with a bare `TRUCK_TYPE_ENCODING.get(truck_type.lower().replace(" ", "_"), 1)` lookup, while the training path normalizes DB domain values through `TRUCK_TYPE_FROM_DB`/`_normalize_truck_type`. App/DB-style values such as "Container" or "Refrigerated" silently fall back to index 1 (medium_truck) at serve time but are encoded as the correct heavy-truck class at train time — a train/serve skew that biases every prediction.

## Evidence
- `backend/ml/app/models/price_prediction.py:427-428`:
  ```python
  truck_encoded = TRUCK_TYPE_ENCODING.get(
      truck_type.lower().replace(" ", "_"), 1   # fallback = medium_truck
  )
  ```
- `backend/ml/app/models/price_prediction.py:239`: training uses `TRUCK_TYPE_ENCODING[_normalize_truck_type(row.get("truck_type"))]` where `_normalize_truck_type` (172-180) maps e.g. `TRUCK_TYPE_FROM_DB["Container"] -> "heavy_truck"`.
- `TRUCK_TYPE_FROM_DB` defined at price_prediction.py:44.

"container" is not a key in `TRUCK_TYPE_ENCODING`, so the default index 1 is used and the serve-time feature never matches the training distribution.

## Expected behavior
`predict_price` must apply the same `_normalize_truck_type` mapping used in training (and reject unknown values instead of defaulting).

## Impact
MEDIUM — systematically wrong price predictions for the most common heavy/refrigerated truck classes.

## References
- backend/ml/app/models/price_prediction.py:28-44, 172-180, 239, 427-428

KaparthyReddy:
## Description
During inference, `predict_price` encoded `truck_type` using a basic `truck_type.lower().replace(" ", "_")` dictionary lookup. Because DB/App-style strings such as `"Container"` or `"Refrigerated"` are not raw keys in `TRUCK_TYPE_ENCODING`, they silently fell back to index 1 (`medium_truck`). However, during training, `_normalize_truck_type` properly mapped these values into domain classes like `heavy_truck` via `TRUCK_TYPE_FROM_DB`. This created a severe train/serve feature skew for heavy and specialized vehicle predictions.

### Changes Made:
- Updated `predict_price` in `backend/ml/app/models/price_prediction.py` to route `truck_type` through `_normalize_truck_type` before performing the `TRUCK_TYPE_ENCODING` lookup.

Fixes #6896

## Type of Change

## Checklist:

### DIFF
--- a/backend/api/src/middleware/idempotency.js
+++ b/backend/api/src/middleware/idempotency.js
@@ -121,7 +121,12 @@ export function requireIdempotency(ttlSeconds = 3600) {
 
             const lockStillHeld = await redisClient.get(lockKey);
             if (!lockStillHeld) {
-              break; // Lock released but cache empty
+              const finalRaw = await redisClient.get(key);
+              const finalCached = finalRaw ? readAndParse(finalRaw) : null;
+              if (finalCached) {
+                return res.status(finalCached.statusCode).json(finalCached.body);
+              }
+              break; // Lock released but cache genuinely empty
             }
 
             retries--;
--- a/backend/ml/anomaly/models.py
+++ b/backend/ml/anomaly/models.py
@@ -112,7 +112,10 @@ class LSTMAutoencoder:
         if self.model is None:
             raise ValueError("Model not trained yet")
         
-        sequence = np.array(sequence).reshape(1, self.sequence_length, self.input_dim)
+        sequence = np.array(sequence)
+        if sequence.size == self.input_dim:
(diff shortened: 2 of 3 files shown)


KaparthyReddy:
gssoc 26

PR #6957, opened 2026-08-06
fix : validate geofence_radius_m is a positive number before use in delivery confirm

### LINKED ISSUE #6936: fix : validate geofence_radius_m is a positive number before use in delivery confirm
POST /api/deliveries/:id/geofence-confirm reads geofence_radius_m from req.body, parses it with parseFloat, and passes it to deliveryVerificationService.geofenceAutoConfirm. parseFloat('abc') returns NaN. The nullish coalescing fallback (NaN ?? 500) still evaluates to NaN in JavaScript, so the distance check distanceM > NaN is always false, silently bypassing the driver-at-dropoff proximity requirement.

Changes that Need to be Made:
- backend/api/src/routes/orderRoutes.js line 1105: validate that geofence_radius_m is a finite positive number before passing it; return 400 if invalid
- backend/api/src/services/order/deliveryVerificationService.js geofenceAutoConfirm: add a guard to reject non-finite effectiveRadiusM early

Impact that it would Provide:
- Prevents invalid geofence_radius_m values from bypassing the delivery proximity check
- Correctly rejects malformed requests with a 400 error

tmdeveloper007:
Closes #6936.

POST /api/deliveries/:id/geofence-confirm parsed geofence_radius_m with parseFloat but never validated the result. parseFloat('abc') returns NaN; NaN ?? 500 evaluates to NaN in JavaScript, making the distance check distanceM > NaN always false. This silently bypassed the driver-at-dropoff proximity requirement.

Changes Made:
- backend/api/src/routes/orderRoutes.js: added validation that geofence_radius_m (if provided) is a positive finite number, returning 400 with a clear error if invalid
- backend/api/src/services/order/deliveryVerificationService.js: added a defensive guard in assertDriverAtDropoff to reject non-finite radiusM values early

Impact it Made:
- Invalid geofence_radius_m values now return a 400 error instead of silently bypassing the proximity check
- Defense-in-depth prevents forged proximity confirmations

Note: This is a GSSOC contribution.

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -1081,6 +1081,13 @@ router.post(
         return res.status(400).json({ error: 'driver_lat and driver_lng must be valid numbers.' });
       }
 
+      if (geofence_radius_m !== undefined) {
+        const radiusM = parseFloat(geofence_radius_m);
+        if (!Number.isFinite(radiusM) || radiusM <= 0) {
+          return res.status(400).json({ error: 'geofence_radius_m must be a positive number' });
+        }
+      }
+
       // Verify the order exists and the requesting driver is assigned to it
       const order = await orderValidationService.findOrderByIdOrDisplayId(
         req.params.id,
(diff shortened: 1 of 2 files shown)


tmdeveloper007:
Closing as duplicate - geofence_radius_m validation already merged in #6989.

Issue #6989, opened 2026-08-06
fix : add NaN guard for driver_lat and driver_lng in delivery geofence confirm route


tmdeveloper007:
Summary of What Needs to be Done:
POST /api/deliveries/:id/geofence-confirm parses driver_lat and driver_lng with parseFloat at lines 1078-1079 and checks Number.isFinite, but this check happens before geofence_radius_m is parsed and used. If geofence_radius_m is provided but driver_lat/lng are malformed strings that parse to NaN, the existing Number.isFinite check at line 1080 catches it. However, the parseFloat call on geofence_radius_m at line 1105 bypasses validation entirely and passes NaN to the service.

Changes that Need to be Made:
- backend/api/src/routes/orderRoutes.js: Add validation for geofence_radius_m after the existing driver_lat/lng checks: if geofence_radius_m is provided (not undefined) but parseFloat returns NaN or a non-positive number, return 400 error.

Impact that it would Provide:
- Prevents NaN geofence radius from being passed to the service layer.
- Consistent input validation across all three numeric body parameters.

Note: Please assign this issue to the `tmdeveloper007` account.

PR #6997, opened 2026-08-06
fix : added validation for geofence_radius_m as a positive number in delivery confirm

### LINKED ISSUE #6961: fix : validate geofence_radius_m is a positive number before use in delivery confirm
POST /api/deliveries/:id/geofence-confirm reads geofence_radius_m from req.body, parses it with parseFloat, and passes the result to deliveryVerificationService.geofenceAutoConfirm. parseFloat('abc') returns NaN. The nullish-coalescing fallback (NaN ?? 500) still evaluates to NaN because ?? only checks for null/undefined, not NaN. This causes assertDriverAtDropoff to compute effectiveRadiusM = NaN, making the distanceM > NaN check always false and bypassing geofence enforcement entirely.

Changes that Need to be Made:
- backend/api/src/routes/orderRoutes.js: After parsing geofence_radius_m with parseFloat, add an explicit Number.isFinite check. If the parsed value is not finite, return a 400 error with a descriptive message instead of passing NaN to the service.

Impact that it would Provide:
- Prevents NaN geofence radius from bypassing security enforcement on the delivery confirmation endpoint.
- API returns a clear 400 error for invalid input instead of silently accepting any value.

tmdeveloper007:
Closes #6961.

Added explicit Number.isFinite and positive-value validation for geofence_radius_m body parameter in the geofence-confirm route handler. If geofence_radius_m is provided but parses to NaN or a non-positive number, the API now returns a 400 error instead of passing NaN to the service layer.

Changes Made:
- backend/api/src/routes/orderRoutes.js: Added validation block after driver_lat/lng checks. Uses a validated property on req.body to avoid repeating parseFloat.

Impact it Made:
- Prevents NaN geofence radius from bypassing security enforcement on delivery confirmation.
- API returns clear 400 errors for invalid input instead of silent failures.

This PR is submitted as part of GirlScript Summer of Code (GSSOC).

### DIFF
--- a/backend/api/src/routes/orderRoutes.js
+++ b/backend/api/src/routes/orderRoutes.js
@@ -1081,6 +1081,15 @@ router.post(
         return res.status(400).json({ error: 'driver_lat and driver_lng must be valid numbers.' });
       }
 
+      if (geofence_radius_m !== undefined) {
+        const parsedRadius = parseFloat(geofence_radius_m);
+        if (!Number.isFinite(parsedRadius) || parsedRadius <= 0) {
+          return res.status(400).json({ error: 'geofence_radius_m must be a positive number.' });
+        }
+        // Only pass the validated radius; the service will apply its own default if undefined.
+        req.body._validatedGeofenceRadiusM = parsedRadius;
+      }
+
       // Verify the order exists and the requesting driver is assigned to it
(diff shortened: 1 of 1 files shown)


tmdeveloper007:
Closing as duplicate - geofence_radius_m validation already merged in #6989.

PR #7382, opened 2026-08-06
fix : return 404 when no matching device token to unregister in unregisterDeviceToken

### LINKED ISSUE #7359: bug : ebpf/routes.js malformed template literal in bpftool command — produces invalid shell command
ebpf/routes.js:209 has a malformed template literal in the bpftool exec command. The string concatenation produces an invalid shell command due to unbalanced quotes. The current code uses JavaScript string concatenation inside a template literal in a way that produces an invalid bpftool load command, causing the exec call to fail.

Changes that Need to be Made:
- ebpf/routes.js: Fix the bpftool command construction to use proper template literal syntax or correct string interpolation, ensuring the command is a valid shell invocation.

Impact that it would Provide:
- eBPF program loading via /api/ebpf/* becomes functional instead of silently failing with malformed commands.

tmdeveloper007:
Closes #7359.

backend/api/src/controllers/deviceController.js unregisterDeviceToken() returned {success:true} even when no rows were deleted from user_devices (i.e., the token didn't match any record). This fix checks the delete query result count and returns 404 when no matching rows are found.

Changes Made:
- backend/api/src/controllers/deviceController.js: Added count extraction from the delete query result and returns 404 with an appropriate error message when count is 0.

Impact it Made:
- Clients receive accurate feedback about whether a device token was actually unregistered.


tmdeveloper007:

PR #7387, opened 2026-08-06
fix : avoid false-positive validation for metadata containing an error key

### LINKED ISSUE #7364: bug : getDevicePlatforms fetches entire user_devices table into memory before filtering
backend/api/src/controllers/deviceController.js getDevicePlatforms() does .select('platform') without any .eq() filter on user_id or other field, loading all rows from the user_devices table into memory before extracting unique platforms. On a large table this causes unnecessary memory pressure and network transfer.

Changes that Need to be Made:
- backend/api/src/controllers/deviceController.js: Add a .select('platform') call that filters by the authenticated user's user_id, or use a Supabase RPC/distinct query that computes unique platforms server-side without fetching all rows.

Impact that it would Provide:
- Reduced memory usage and network transfer for the platforms endpoint.
- Better performance as the table grows.

tmdeveloper007:
Closes #7364.

backend/api/src/controllers/deviceController.js registerDeviceToken() checked if metadata contained an 'error' key using a simple truthiness check (if (normalizedMetadata.error)). This caused false positives when legitimate metadata objects happened to contain an 'error' field with a falsy value. This fix checks specifically for a truthy string error.

Changes Made:
- backend/api/src/controllers/deviceController.js: Changed the metadata error check to if (normalizedMetadata && normalizedMetadata.error && typeof normalizedMetadata.error === 'string').

Impact it Made:
- Legitimate metadata with an 'error' key is no longer falsely rejected.


tmdeveloper007:

PR #7394, opened 2026-08-06
fix : use supabaseAdmin for driver search RLS-protected queries in truckRoutes

### LINKED ISSUE #7327: bug: truck/driver search queries driver_details, trucks and profiles with the anon client — /api/trucks/search always errors or returns empty
## Description

The driver-search endpoints in backend/api/src/routes/truckRoutes.js query RLS-protected tables through the shared **anon-key** `supabase` client:
- `driver_details` SELECT at :591-596 (filtered by `is_online`, `truck_id`, `user_id`)
- `trucks` SELECT at :611
- `profiles` SELECT at :612

All three tables are RLS-enabled with policies only for `service_role` and the owning `authenticated` user (supabase/migrations/20240101000000_rls.sql:54-62 `driver_details`, :81-90 `trucks`, :22-44 `profiles`) and ALL privileges are revoked from `anon` (supabase/migrations/revoke_anon_privileges.sql:5-8). The shared client has no user session bound to it.

## Expected Behaviour

`GET /api/trucks/search?...` returns matching drivers with truck/profile enrichment (name, rating, capacity, price estimate — truckRoutes.js:632-649).

## Actual Behaviour

The `driver_details` query (:591-596) runs as the anon role, which has no RLS policy and no grant on the table, so `driversErr` is set and the route returns `500 Failed to search trucks` (:598-601) — or, where the revoke is not applied, an empty list because anon sees no rows. Either way the driver-search feature (which powers load-to-driver matching) never returns real results. The nearby-drivers flow uses the same pattern.

## Proposed Fix

Use a user-scoped client for these authenticated reads — either bind the user's session to the shared client in the `authenticate` middleware, or use `createUserClient(req.user)` / `supabaseAdmin` from `config/db.js` — for the `driver_details`, `trucks`, and `profiles` queries in truckRoutes.js. If the read must see all drivers (a marketplace search), use the service-role client and scope by the search criteria, never the raw anon key.

tmdeveloper007:
Closes #7327.

backend/api/src/routes/truckRoutes.js queries driver_details, trucks, and profiles tables via the anon-key supabase client. RLS policies prevent the anon role from reading these tables. This fix replaces the anon supabase client with supabaseAdmin for the three driver-search queries. Also added a limit(200) to the MongoDB geospatial telemetry query to prevent unbounded memory usage.

Changes Made:
- backend/api/src/routes/truckRoutes.js: Changed driver_details, trucks, and profiles queries to use supabaseAdmin; added limit(200) to MongoDB geospatial query

Impact it Made:
- Driver search results actually return data instead of empty arrays due to RLS denials.
- No unbounded memory growth from large geospatial query results.


tmdeveloper007:

PR #7396, opened 2026-08-06
fix : use supabaseAdmin for stale order worker RLS-protected queries

### LINKED ISSUE #7319: bug: staleOrderWorker queries orders via the anon-key supabase client — RLS blocks every read so stale orders are never cancelled
## Description

`startStaleOrderWorker` (backend/api/src/workers/staleOrderWorker.js) is started from index.js:673 but never actually cancels stale orders, because every query goes through the shared **anon-key** `supabase` client imported at staleOrderWorker.js:3.

The repo already documents this exact failure mode in index.js:662-667:

> "They must run with the service-role client: the anon client has no RLS read access to `orders`, so an anon-backed repository would silently no-op."

That service-role fix was applied to the escrow reconciliation workers (index.js:665-667: `supabaseAdmin ? new OrderRepository(supabaseAdmin) : orderRepository`) but **not** to `staleOrderWorker`.

## Expected Behaviour

The hourly cron selects `orders` where `status = 'pending'` and `created_at < now - 24h` (staleOrderWorker.js:25-29), re-fetches each candidate (102-107), then cancels it (175-194) or routes it through the escrow refund pipeline (206-291). Pending orders older than 24 hours with no accepted bid should be cancelled and the customer notified.

## Actual Behaviour

`orders` is RLS-enabled with policies only for `service_role` and `authenticated` (supabase/migrations/20240101000000_rls.sql:164-183) and all privileges are revoked from `anon` (supabase/migrations/revoke_anon_privileges.sql:14). The anon client therefore receives a permission error (or an empty result set) on the SELECT at :25. `fetchError` is set and the worker returns at :31-34. Even where a read slipped through, the guarded UPDATEs at :176-186, :212-226, :254-266, :276-287 and the `load_offers` update at :143-146 all use the same anon client and are blocked. Net effect: orders stuck in `pending` are never cancelled, escrow is never refunded, and no cancellation notification is ever sent — silently, every hour.

## Proposed Fix

Follow the pattern already used for the escrow reconciliation workers in index.js:665-667. Import `supabaseAdmin` from `config/db.js`, build a service-role-backed repository (`new OrderRepository(supabaseAdmin)`), and route all `orders`/`load_offers` reads and writes in staleOrderWorker.js through it instead of the shared `supabase` client. Keep a fallback to the anon repository only when `supabaseAdmin` is unavailable, and log a warning in that case.

tmdeveloper007:
Closes #7319.

backend/api/src/workers/staleOrderWorker.js queries the orders table via the anon-key supabase client. RLS policies prevent the anon role from querying orders, so the worker finds no stale orders and never cancels them. This fix replaces all supabase references with supabaseAdmin in staleOrderWorker.js.

Changes Made:
- backend/api/src/workers/staleOrderWorker.js: Replaced all supabase client calls with supabaseAdmin for orders and load_offers queries

Impact it Made:
- Stale orders are now properly identified and cancelled by the background worker.


tmdeveloper007:

PR #7398, opened 2026-08-06
fix : use supabaseAdmin for document expiry worker RLS-protected queries

### LINKED ISSUE #7321: bug: documentExpiryService reads documents and notifications with the anon client — RLS blocks the reads so expiry reminders are never sent
## Description

`startDocumentExpiryWorker` runs `processDocumentExpiryBatch` (backend/api/src/services/documentExpiryService.js:84), which queries `documents` for rows expiring in the 30/14/7-day windows (:130-135) and checks `notifications` for duplicates (`hasExistingNotification`, :61-82). Both queries use the shared **anon-key** `supabase` client imported at :1.

`documents` is RLS-enabled with policies only for `service_role` and the owning `authenticated` user (supabase/migrations/20240101000000_rls.sql:150-159) and privileges revoked from `anon` (supabase/migrations/revoke_anon_privileges.sql:13); `notifications` is likewise revoked from `anon` (:26). The anon client carries no user session, so both reads fail or return empty.

## Expected Behaviour

The daily worker finds documents expiring in each window and sends exactly one reminder per document per window via `sendPushNotification` (:186). Interval is configurable via `DOCUMENT_EXPIRY_WORKER_INTERVAL_MS` (:234-244).

## Actual Behaviour

No documents are ever found because the `documents` query (:130-135) runs as the anon role, which has no RLS policy and no grant on the table — the error path logs and `continue`s (:137-140), or the batch finishes with zero documents. The `notifications` dedup check (:61-82) also fails, so reminders would duplicate if the read path were fixed in isolation. Net effect: drivers/customers never receive document-expiry reminders, silently, every 24h.

## Proposed Fix

Switch the `documents` and `notifications` queries in documentExpiryService.js to the service-role `supabaseAdmin` client (already exported from `config/db.js`), extending the existing `if (!supabase) return false` guards to `supabaseAdmin`. This mirrors how notificationService.js persists delivery OTPs/notifications with the admin client.

tmdeveloper007:
Closes #7321.

backend/api/src/services/documentExpiryService.js queries documents and notifications tables via the anon-key supabase client. RLS policies prevent the anon role from reading these tables, so the expiry worker cannot find expiring documents or check for existing notifications. This fix uses supabaseAdmin for both queries.

Changes Made:
- backend/api/src/services/documentExpiryService.js: Changed hasExistingNotification and processDocumentExpiryBatch to use supabaseAdmin

Impact it Made:
- Document expiry notifications are sent to drivers before their documents expire.


tmdeveloper007:

PR #7436, opened 2026-08-07
fix : used sentinel key for normalizeMetadata validation in deviceController

### LINKED ISSUE #7408: fix : normalizeMetadata false-positive on error key in deviceController
In backend/api/src/controllers/deviceController.js, the normalizeMetadata function returns { error: '...' } to signal validation failures (e.g., when metadata is not a plain object). However, the function also checks normalizedMetadata.error to detect validation failures. This means legitimate metadata objects containing an 'error' key (e.g., { error: 'some actual error from another system' }) are incorrectly rejected as validation failures.

Changes that Need to be Made:
- Change normalizeMetadata to use a separate internal flag (e.g., _validationError) instead of the 'error' key for signaling validation failures
- The function should still validate that metadata is a plain object, but the detection mechanism should not rely on the presence of an 'error' key

Impact that it would Provide:
- Fixes false-positive validation rejections for valid metadata containing an 'error' key
- Improves robustness of device registration with third-party metadata

tmdeveloper007:
Closes #7408.

Changed the normalizeMetadata function in deviceController.js to use a private _validationError key instead of the public 'error' key to signal validation failures. This prevents legitimate metadata containing an 'error' key from being incorrectly rejected as a validation failure.

Changes Made:
- normalizeMetadata now returns { _validationError: '...' } for validation failures
- The caller now checks normalizedMetadata._validationError instead of normalizedMetadata.error

Impact it Made:
- Fixes false-positive validation rejections for valid metadata objects containing an 'error' key
- Allows third-party metadata with an 'error' field to be stored correctly

Note: This PR is submitted as part of GSSOC (GirlScript Summer of Code).


tmdeveloper007:

PR #7608, opened 2026-08-07
fix : removed digilockerService writes to non-existent documents.storage_path column

tmdeveloper007:
Closes #7585.

Summary of What Has Been Done:
Removed writes to non-existent `documents.storage_path` column from digilockerService.js. The documents table does not have a storage_path column, causing Supabase RLS errors on document synchronization.

Changes Made:
- backend/api/src/services/digilockerService.js: Removed storage_path from driver_documents payload
- backend/api/src/services/digilockerService.js: Removed storage_path from documents payload (legacy/Flutter-facing sync)

Impact it Made:
- Fixes Supabase RLS errors on document synchronization
- Prevents document sync failures from blocking KYC verification

Note: Please assign this PR to the `tmdeveloper007` account.

---

This PR is submitted as part of GSSOC (GirlScript Summer of Code).

PR #7779, opened 2026-08-07
fix : added LARGE_WITHDRAWAL detection guard for non-withdrawal transactions

### LINKED ISSUE #7759: bug : fix detectLargeWithdrawal classifies deposits as LARGE_WITHDRAWAL
backend/api/src/services/security/anomalyDetectionService.js runs detectLargeWithdrawal for every transaction regardless of its direction. The function compares the transaction amount against the user's historical withdrawal statistics (average, std-dev) without checking transaction.type. Large legitimate deposits (e.g. escrow payouts) exceed the withdrawal z-score threshold and trigger shouldBlockTransaction -> lockAccount, locking users out of their accounts.

Changes that Need to be Made:
- backend/api/src/services/security/anomalyDetectionService.js: Add an early return in detectLargeWithdrawal when transaction.type indicates a deposit or credit (not a withdrawal). The function should only score transactions that are actually withdrawals.
- Add a unit test in backend/api/test/unit/anomalyDetectionService.test.js covering a large deposit transaction that should NOT be flagged as LARGE_WITHDRAWAL.

Impact that it would Provide:
- Legitimate large deposits will no longer trigger account locks
- Users who primarily withdraw small amounts will not be locked out when receiving large credits
- The anomaly detection will be accurate to its stated purpose (withdrawal anomalies)

tmdeveloper007:
Closes #7759.

Fixed detectLargeWithdrawal in backend/api/src/services/security/anomalyDetectionService.js to skip scoring when transaction.type is not 'withdrawal'. Previously, large deposits and credits could be flagged as LARGE_WITHDRAWAL because the function never checked the transaction type. Added test cases covering deposit and credit transactions returning null.

Changes Made:
- backend/api/src/services/security/anomalyDetectionService.js: Added early return when transaction.type !== 'withdrawal'
- backend/api/test/unit/anomalyDetectionService.test.js: Added two test cases for deposit and credit transaction types

Impact it Made:
- Large legitimate deposits no longer trigger account lockout
- Anomaly detection accurately targets only withdrawal transactions

---


tmdeveloper007:
## Closing this PR — fix already present on main

Hi @KanishJebaMathewM,

After investigating, I found that this PR's fix is **already applied on current `main`**:

- For #7779: `anomalyDetectionService.js` on main already guards `detectLargeWithdrawal` with the `isWithdrawalDirection()` check (and has the corresponding unit tests), so non-withdrawal transactions are never scored as LARGE_WITHDRAWAL. The branch now contains no unique changes.
- For #7608: `digilockerService.js` on main no longer writes to the non-existent `documents.storage_path` column (removed in a later refactor), so the fix is redundant. The branch now contains no unique changes.

I've updated the branch to current `main` so there's nothing left to merge here. Closing to keep the open-PR list clean.

This PR was a part of GSSOC (GirlScript Summer of Code).

_Closed by tmdeveloper007 after investigation._

PR #8160, opened 2026-08-08
fix(customer): read pickup/drop city from pickup_address/drop_address on home card

### LINKED ISSUE #8147: customer home_screen reads pickup_city/drop_city keys that the API never returns - every shipment card shows '? -> ?'
## Description

In `apps/customer/lib/screens/home_screen.dart:189`, `_buildShipmentFromOrder` builds the route label from keys that do not exist:

```dart
final route = '${order['pickup_city'] ?? '?'} → ${order['drop_city'] ?? '?'}';
```

The backend and every other consumer return/read `pickup_address` and `drop_address` (e.g. `orderRoutes.js`, `home_screen.dart:111-112`, `orders_screen.dart`, `shell_screen.dart`, `order_detail_screen.dart`). `pickup_city` / `drop_city` appear **nowhere** in the backend, so this always resolves to `'?' → '?'`. Every shipment card on the customer home screen shows the broken route text.

## Expected behavior

Use the address keys the API actually returns:

```dart
final route = '${order['pickup_address'] ?? '?'} → ${order['drop_address'] ?? '?'}';
```

## Location

- `apps/customer/lib/screens/home_screen.dart:189`

## Suggested fix

Replace `pickup_city`/`drop_city` with `pickup_address`/`drop_address`.

(GSSOC-2026)

Xenon010101:
## Summary
In `apps/customer/lib/screens/home_screen.dart`, `_buildShipmentFromOrder` read the route from `order['pickup_city']` and `order['drop_city']`, but the API order payload only contains `pickup_address` / `drop_address` (see `order_service.dart`, which sends exactly those keys). As a result the active-shipment card always displayed the placeholder `? → ?` even when the order had valid addresses.

## Changes
- Build the route string from `pickup_address` / `drop_address` instead.
- Reuse the existing `_shortenAddress` helper (same display logic as the "usual routes" cards) so only the city portion is shown, keeping the card compact.
- Fall back to `?` only when the address is empty/missing.

Closes #8147

GSSOC 2026

### DIFF
--- a/apps/customer/lib/screens/home_screen.dart
+++ b/apps/customer/lib/screens/home_screen.dart
@@ -186,7 +186,9 @@ class _HomeScreenState extends State<HomeScreen> {
   }
 
   ShipmentCardData? _buildShipmentFromOrder(Map<String, dynamic> order) {
-    final route = '${order['pickup_city'] ?? '?'} \u2192 ${order['drop_city'] ?? '?'}';
+    final pickupCity = _shortenAddress(order['pickup_address']?.toString() ?? '');
+    final dropCity = _shortenAddress(order['drop_address']?.toString() ?? '');
+    final route = '${pickupCity.isNotEmpty ? pickupCity : '?'} \u2192 ${dropCity.isNotEmpty ? dropCity : '?'}';
     final rawDriverName = order['driver_name']?.toString() ?? '';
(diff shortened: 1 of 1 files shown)


Xenon010101:
@KanishJebaMathewM  please review it

Issue #8163, opened 2026-08-08
[BUG][backend][P1] staleOrderWorker: leftover pre-RPC re-fetch makes every stale cancellation throw — no notification, metrics report errors


advikdivekar:
## Summary

`cancelStaleOrder()` in `backend/api/src/workers/staleOrderWorker.js` still contains the pre-RPC re-fetch from the superseded implementation. It runs on **every** successful cancellation and throws, so no stale order ever gets its load offer cancelled or its customer notified — and the sweep reports the cancellation as an error.

## Reproduction

```
$ cd backend/api
$ npx vitest run test/unit/staleOrderWorkerConcurrency.test.js test/unit/staleOrderWorkerNotifications.test.js
 Tests  4 failed | 6 passed
```

```
- { 'stale_orders.cancelled': 1, 'stale_orders.skipped': 1, 'stale_orders.errors': 0 }   expected
+ { 'stale_orders.cancelled': 0, 'stale_orders.skipped': 0, 'stale_orders.errors': 2 }   actual
```

## Details

Immediately after the atomic `cancel_stale_order_tx` RPC returns, the worker does this:

```js
const { data: current, error: refetchErr } = await staleOrderClient
  .from('orders')
  .select('id, customer_id, order_display_id, escrow_status, refund_tx_hash, escrow_refund_attempts')
  .eq('id', staleOrder.id)
  .eq('status', 'pending')          // (2)
  .maybeSingle();
```

Two independent faults:

1. **`staleOrderClient` is only assigned inside `startStaleOrderWorker()`.** `reconcileStaleOrders()` is exported and called directly — by the tests, and by anything that drives a sweep without going through the cron bootstrap. In that path `staleOrderClient` is still `null`, so `staleOrderClient.from(...)` throws `TypeError: Cannot read properties of null`. The outer `catch` swallows it into `metrics.errors`, so a cancellation that actually succeeded on the database is reported as a failure.

2. **Even with the client set, the filter can never match.** The RPC has just moved the row to `status = 'cancelled'`, so a re-fetch filtered on `.eq('status', 'pending')` returns `null`. `refetchErr` is also `null`, so the guard below it does not fire and execution reaches `current.order_display_id` → `TypeError: Cannot read properties of null`. Same outcome.

Either way the function exits through the outer `catch` **before** `sendPushNotification`, so:

- the customer is never told their order was cancelled;
[...]

PR #8169, opened 2026-08-08
[Bug] Restore staleOrderWorkerRace suite so its 6 CAS race tests run again

### LINKED ISSUE #8168: [BUG][backend][P2] staleOrderWorkerRace.test.js fails to parse — 6 stale-order CAS race tests never run
## Summary

`backend/api/test/unit/staleOrderWorkerRace.test.js` does not parse. Vitest cannot collect it, so the 6 TOCTOU regression tests it holds (issue #5741) have not run for some time.

## Reproduction

```
$ cd backend/api
$ npx vitest run test/unit/staleOrderWorkerRace.test.js

Cannot parse .../test/unit/staleOrderWorkerRace.test.js:
Expected `,` or `}` but found `.`
127:       updateLoadOffer: vi.fn().mockResolvedValue({ data: null, error: null }),
128:     submitEscrowRefundMock.mockReset();
                               ^
 FAIL  test/unit/staleOrderWorkerRace.test.js
RolldownError: Parse failure
```

## Details

A merge kept both the current repository-based suite and the superseded `supabaseAdmin`-builder suite, splicing one into the other. The current file contains:

- the old `makeBuilder` / `supabaseAdminBuilder` / `supabaseBuilder` / `scriptedResponses` / `ordersUpdateCalls` scaffolding, which nothing in the live tests uses;
- a `vi.mock('../../src/config/db.js', ...)` factory with **duplicate `supabase` and `supabaseAdmin` keys** — `{}` first, then the old builders, so the builders silently win;
- the old `beforeEach` body spliced into the middle of the new one's `orderRepository` object literal, at line 128. This is the parse error: the object literal is never closed, and the new `beforeEach` never gets its `await import(...)` of `reconcileStaleOrders`;
- a leftover test, `routes orders queries through the service-role client, never the anon client`, that calls `scheduledHandler()` — an identifier the file never defines;
- a `stillPending()` helper whose body ends with the tail of the new `beforeEach` (`({ reconcileStaleOrders } = await import(...))`), stranded after a `return`.

## Impact

The six tests guarding the stale-order CAS race never execute:

- side effects skipped when the CAS claim is lost;
- a lost race not blocking other orders in the batch;
- funded escrow routed to refund reconciliation without the worker submitting the refund itself;
- exactly one notification per genuine cancellation;
- RPC error paths skipping side effects.

That is precisely the coverage that would have caught #8163, where the worker throws on every successful cancellation.

## Expected

The suite parses and its 6 tests run.

## Suggested fix

Drop the superseded scaffolding, the duplicate mock keys, the `scheduledHandler` test and the `stillPending` stub, and close the `orderRepository` literal so the current `beforeEach` is whole again.

> Depends on #8163 — with the parse fixed, 3 of the 6 tests still fail against `main` because of the worker bug that issue describes. All 6 pass with that fix applied.

advikdivekar:
Fixes #8168.

`test/unit/staleOrderWorkerRace.test.js` did not parse, so vitest could not collect any of the 6 TOCTOU regression tests it holds (issue #5741).

### Before

```
Cannot parse test/unit/staleOrderWorkerRace.test.js:
Expected `,` or `}` but found `.`
127:       updateLoadOffer: vi.fn().mockResolvedValue({ data: null, error: null }),
128:     submitEscrowRefundMock.mockReset();
                               ^
```

A merge kept both the current repository-based suite and the superseded `supabaseAdmin`-builder one, splicing the old `beforeEach` body into the middle of the new one's `orderRepository` object literal.

### Removed — all from the superseded suite

- `makeBuilder` / `supabaseAdminBuilder` / `supabaseBuilder` / `scriptedResponses` / `ordersUpdateCalls` scaffolding, unused by the live tests
- **Duplicate `supabase` and `supabaseAdmin` keys** in the `config/db.js` mock factory — declared `{}` first, then re-declared as the old builders, which silently won
- The `routes orders queries through the service-role client` test, which calls `scheduledHandler()`, an identifier the file never defines
- The `stillPending()` stub, whose body ended with the tail of the new `beforeEach` stranded after a `return`

Closing the `orderRepository` literal also restores the `await import('../../src/workers/staleOrderWorker.js')` that the current `beforeEach` needs.

### Verification

```
$ npx vitest run test/unit/staleOrderWorkerRace.test.js
      Tests  3 failed | 3 passed (6)      # against main
      Tests  6 passed (6)                 # with #8164 applied
```

The suite now collects. The 3 remaining failures against `main` are the worker bug in #8163 — exactly the regression this coverage exists to catch — and all 6 pass with #8164 applied. `npx eslint` clean.

> **Depends on #8164.** Merging that first turns this suite green.

### DIFF
--- a/backend/api/test/unit/staleOrderWorkerRace.test.js
+++ b/backend/api/test/unit/staleOrderWorkerRace.test.js
@@ -5,47 +5,6 @@ const loggerWarnMock = vi.fn();
 const redisSetMock = vi.fn();
 const redisDelMock = vi.fn();
 const redisExpireMock = vi.fn();
-const submitEscrowRefundMock = vi.fn();
-const confirmEscrowRefundMock = vi.fn();
-const ordersUpdateCalls = [];
-let scriptedResponses = [];
-const supabaseAdminBuilder = { from: vi.fn(makeBuilder) };
-const supabaseBuilder = { from: vi.fn(makeBuilder) };
-
-function makeBuilder(table) {
-  const builder = {
-    _mode: 'select',
-    _payload: null,
-    _filters: [],
-    select() { return this; },
-    eq(column, value) {
-      this._filters.push({ op: 'eq', column, value });
-      return this;
-    },
-    lt() { return this; },
-    or(filter) {
-      this._filters.push({ op: 'or', filter });
-      return this;
-    },
-    in() { return this; },
-    maybeSingle() { return this; },
-    update(payload) {
-      this._mode = 'update';
-      this._payload = payload;
-      return this;
-    },
-    then(resolve) {
-      const scripted = scriptedResponses.shift();
-      const result = scripted ?? { data: null, error: null };
-      if (table === 'orders' && this._mode === 'update') {
-        ordersUpdateCalls.push({ payload: this._payload, filters: this._filters, result });
-      }
-      return resolve(result);
-    },
-  };
-
-  return builder;
-}
 
 vi.mock('node-cron', () => ({
   default: {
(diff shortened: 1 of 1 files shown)


advikdivekar:
> 

GSSOC

PR #8326, opened 2026-08-08
fix(devices): allow registration without optional metadata (#8315)

### LINKED ISSUE #8315: bug: deviceController rejects valid device registration without metadata with 400
## Summary

`backend/api/src/controllers/deviceController.js` — `registerDeviceToken()` rejects **every request that omits `metadata`** with a `400 'metadata must be a plain object'`, even though the request schema marks `metadata` as **optional**.

Commit `a355bb43` ("fix: changed normalizeMetadata to return null on validation failure instead of sentinel object") changed `normalizeMetadata` to return `null` for **both** absent (`undefined`/`null`) **and** invalid metadata. The handler then treats any `null` as invalid and returns 400:

```js
// lines 59-62
const normalizedMetadata = normalizeMetadata(metadata);
if (normalizedMetadata === null) {
  return res.status(400).json({ error: 'metadata must be a plain object' });
}
```

## Evidence

- Schema: `backend/api/src/validation/requestSchemas.js` line 192 — `metadata: z.record(z.any()).optional()` inside `registerDeviceSchema`.
- Route: `backend/api/src/routes/deviceRoutes.js` line 82 validates the body with that schema, so a request without `metadata` passes validation.
- `normalizeMetadata` (lines 25-35): `if (metadata === undefined || metadata === null) return null;`
- The repo's own unit test `backend/api/test/unit/deviceController.test.js` line 32-52 posts `{ fcmToken, platform }` **without metadata** and asserts `res.status` is **not** called with 400 — the test contradicts the implementation (it fails).

## Impact

- Any mobile/web client that registers a device without sending `metadata` gets a 400 and the device token is never stored — FCM push notifications for that user silently never work.
- Regression from `a355bb43`: previously absent metadata returned `{}` and was accepted.

## Proposed fix

Restore the pre-regression behavior for the *absent* case while keeping `null` for *invalid* values:

```js
function normalizeMetadata(metadata) {
  if (metadata === undefined || metadata === null) return {};   // absent -> empty object, allowed
  if (typeof metadata !== 'object' || Array.isArray(metadata)) return null;
  const prototype = Object.getPrototypeOf(metadata);
  if (prototype !== Object.prototype && prototype !== null) return null;
  return metadata;
}
```

Akshita-2307:
## Summary

Fixes #8315 — allows FCM token registration without the optional `metadata` field, which was previously rejected with a 400 even though the request schema marks it as optional.

## Problem

`normalizeMetadata()` returns `null` both for *absent* metadata and for *invalid* metadata. The handler treats any `null` result as a validation failure, so a valid request without `metadata` (allowed by `requestSchemas.js:192` — `metadata: z.record(z.any()).optional()`) was rejected with `400 metadata must be a plain object`.

## Fix

- `normalizeMetadata()` now returns `undefined` for absent metadata and only `null` for genuinely invalid metadata, so the handler's `if (normalizedMetadata === null)` check only rejects invalid input.
- The RPC call now passes `normalizedMetadata ?? null` so the `register_device_token` RPC still receives `null` (not `undefined`) when metadata is absent.

## Verification

- `node --check backend/api/src/controllers/deviceController.js` passes.
- Registration without `metadata` no longer hits the 400 path; invalid metadata still does.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] The code is formatted and linted with the project's standards

**Linked Issues:** closes #8315

**Contributor Info**
- GitHub: @Akshita-2307
- GSSOC '24 Contributor

### DIFF
--- a/backend/api/src/controllers/deviceController.js
+++ b/backend/api/src/controllers/deviceController.js
@@ -23,7 +23,7 @@ function validatePlatform(platform) {
  * are not confused with validation failures.
  */
 function normalizeMetadata(metadata) {
-  if (metadata === undefined || metadata === null) return null;
+  if (metadata === undefined || metadata === null) return undefined;
   if (typeof metadata !== 'object' || Array.isArray(metadata)) {
     return null;
   }
@@ -82,7 +82,7 @@ export async function registerDeviceToken(req, res, next) {
       p_user_id:      userId,
       p_fcm_token:    fcmToken,
       p_platform:     platform || 'android',
-      p_metadata:     normalizedMetadata,
+      p_metadata:     normalizedMetadata ?? null,
       p_prev_user_id: previousUserId ?? null,
     });
 


Akshita-2307:
Closing: main has advanced and this bug is already fixed upstream by PR #7619 (KaparthyReddy, fix/device-controller-error-shapes-v2). The current \
ormalizeMetadata\ returns \{ data: {}, error: null }\ for absent metadata, so valid metadata-less registration no longer returns 400. This PR now conflicts with main and is redundant.

PR #8335, opened 2026-08-08
fix: remove leaked catch block in dlqService.js logBacklogIfNeeded

aaniya22:
## Description
`dlqService.js`'s `logBacklogIfNeeded()` had a leaked copy of the retry/permanent-failure `catch (procErr) { ... }` block from `processQueue()`'s event-processing logic, accidentally pasted into the middle of the backlog-logging method. This referenced `event` and `procErr` which don't exist in that scope, causing a `SyntaxError: Unexpected token 'catch'` that broke the entire module (and anything importing it, including the DLQ worker).

Removed the stray leaked block so `logBacklogIfNeeded()` keeps only its original single `try/catch`, logging the backlog count via `logger.info` and falling back to `logger.warn` on failure — matching its documented "best-effort, no-op on failure" behavior.

## Type of Change
- [x] Bug fix (non-breaking change fixing an issue)

## How Has This Been Tested?
- **Test Commands:** `npx vitest run test/unit/dlqService.test.js`
- **Test Steps / Prerequisites:** Synced local `main` with `upstream/main` to get the affected file version, applied the fix, ran `node --check` to confirm valid syntax, then ran the full existing unit test suite for this file.
- **Expected Result:** File parses without a SyntaxError, and all existing dlqService tests pass unaffected.

### Test Environment
- [x] Local manual testing
- [x] Unit / Integration tests (`npm test`, `flutter test`, etc.)

## Related Issues
Closes #8312

## PR Checklist
- [x] My code follows the code style guidelines of this project.
- [x] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have updated corresponding documentation if applicable.
- [x] My changes generate no new warnings or console errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved webhook failure handling by centralizing retry scheduling and permanent-failure processing.
  * Removed redundant failure persistence logic from the queue-processing flow.

PR #8336, opened 2026-08-08
fix: guard null coordinate result before .error access

### LINKED ISSUE #8313: bug: shardMiddleware null-deref on single-coordinate requests silently routes to north shard
## Summary

`backend/api/src/middleware/shardMiddleware.js` line 32 dereferences `parsedLat.error` / `parsedLng.error` **when either coordinate is missing**. `parseCoordinate(undefined)` returns `null` (line 9), so when a client sends only one coordinate (`?lat=...` without `lng`, or vice versa) the middleware throws `TypeError: Cannot read properties of null (reading 'error')`.

That TypeError is swallowed by the outer `try/catch` (lines 72-78), so the request is **silently routed to the default `north` shard and returns 200 OK** instead of the intended `400 'lat and lng are both required'`.

## Evidence

```js
// lines 28-36
if (rawLat !== undefined || rawLng !== undefined) {
  const parsedLat = parseCoordinate(rawLat);   // -> null when rawLat undefined
  const parsedLng = parseCoordinate(rawLng);   // -> null when rawLng undefined

  if (rawLat === undefined || rawLng === undefined || parsedLat.error || parsedLng.error) {
    return res.status(400).json({
      error: parsedLat.error || parsedLng.error || 'lat and lng are both required when routing by location'
    });
  }
```

```js
// lines 72-78 — the error is swallowed
} catch (error) {
  logger.error('Shard middleware error:', error);
  req.shard = 'north';                     // fallback instead of 400
  req.shardConnection = await shardManager.getShardConnection('north');
  next();
}
```

## Impact

- Requests with partial coordinates get routed to the north shard with a **200** response, so the client believes its (incomplete) location request succeeded.
- The `X-Shard: north` header misleads consumers about which region actually handled the data.
- Any analytics/behavioral fraud tracking keyed on correct 400 rejections for malformed geolocation is bypassed.

## Proposed fix

Guard the null case before reading `.error`, e.g. check `parsedLat === null || parsedLng === null` inside the branch, or make `parseCoordinate` return a sentinel object instead of `null`. Minimal fix:

```js
if (parsedLat === null || parsedLng === null || parsedLat.error || parsedLng.error) {
  return res.status(400).json({
    error: 'lat and lng are both required when routing by location'
  });
}
```

aaniya22:
## Description
`shardMiddleware.js` threw an unhandled `TypeError: Cannot read properties of null (reading 'error')` when a client sent only one of `lat`/`lng`. `parseCoordinate()` returns `null` (not `{ error }`) when a value is `undefined`, but the validation check accessed `parsedLat.error`/`parsedLng.error` directly without first guarding against `null`. The error was silently swallowed by the outer `try/catch`, which then routed the request to the default `north` shard and returned `200 OK` instead of the intended `400`.

Added explicit `null` checks (`parsedLat === null || parsedLng === null`) before the `.error` access, and used optional chaining (`?.`) when building the error message so it can't throw even if one side is `null` while the other has `.error`.

## Type of Change

## How Has This Been Tested?
- **Test Commands:** `npx vitest run test/unit/shardMiddleware.test.js`
- **Test Steps / Prerequisites:** Ran `node --check` to confirm valid syntax, ran the existing unit test suite (no regressions), and manually simulated the exact bug scenario from the issue (only `lat` provided, `lng` missing) in an isolated script to confirm it now correctly triggers the 400 rejection path instead of throwing/falling through to the `north` shard.
- **Expected Result:** Requests missing either coordinate now return `400 { error: 'lat and lng are both required when routing by location' }` instead of silently succeeding on the `north` shard.

### Test Environment
- [x] Local manual testing

## Related Issues
Closes #8313

## PR Checklist
- [ ] I have commented my code, particularly in hard-to-understand areas.

### DIFF
--- a/backend/api/src/middleware/shardMiddleware.js
+++ b/backend/api/src/middleware/shardMiddleware.js
@@ -29,9 +29,9 @@ export const shardMiddleware = async (req, res, next) => {
       const parsedLat = parseCoordinate(rawLat);
       const parsedLng = parseCoordinate(rawLng);
 
-      if (rawLat === undefined || rawLng === undefined || parsedLat.error || parsedLng.error) {
+      if (rawLat === undefined || rawLng === undefined || parsedLat === null || parsedLng === null || parsedLat.error || parsedLng.error) {
         return res.status(400).json({
-          error: parsedLat.error || parsedLng.error || 'lat and lng are both required when routing by location'
+          error: parsedLat?.error || parsedLng?.error || 'lat and lng are both required when routing by location'
         });
       }
 


aaniya22:
Confirmed all shard 1 CI failures are pre-existing and unrelated to this diff.

This PR only touches `backend/api/src/middleware/shardMiddleware.js` (2 lines, null-guard before `.error` access).

**Verified on a clean `main` checkout (no shardMiddleware changes present):**
npx vitest run test/unit/osrm.test.js
→ Same 5 failures, same line numbers, same error messages as in CI — confirming these predate this PR:
- `buildCacheKey` rounding mismatch (7 vs 6 decimal places)
- Redis cache-key assertion failing for the same reason
- `getRouteEstimate` throwing on null input (`Cannot destructure property 'pickupLat' ... as it is null`)
- Two fetch-mock related failures cascading from the same root cause

**Other shard 1 failures, also unrelated to this diff:**
- `orderTimelineService.test.js` — `Cannot read properties of undefined (reading 'addTimelineEvent')`, broken import/mock, unrelated to shardMiddleware
- `idempotency.test.js` — one test failing on a stale expected error-message string
- `truckSearchServiceRole.test.js` (failed suite) — duplicate exports (`verifyDeliveryLimiter`, `resendOtpLimiter`, etc.) in `src/middleware/rateLimiter.js`, a separate pre-existing bug
- `orderRoutesGeofence.test.js` (failed suite) — vite transform error, unrelated
- `weighStationService.test.js`, `staleOrderWorkerNotifications.test.js`, `withdrawalSettlementWorker.test.js` — same flaky pattern seen and confirmed pre-existing on PR #8335 (timeouts / mocks not firing)

This PR should be safe to review/merge on its own merits.

PR #8337, opened 2026-08-08
fix: track wrapper reference per handler to fix EventBus.unsubscribe 

### LINKED ISSUE #8316: bug: EventBus.unsubscribe cannot remove function handlers (wrapper reference mismatch)
## Summary

`backend/api/src/core/events/EventBus.js` — `unsubscribe()` can **never remove** a handler that was added via `subscribe()`.

`subscribe()` wraps every function handler in a new `tracedHandler` closure and registers **that wrapper** with `this.on(eventType, tracedHandler)` (line 188). But `unsubscribe()` calls `this.removeListener(eventType, handler)` with the **original** handler reference (line 203). Since `EventEmitter.removeListener` matches by strict reference identity, the registered wrapper is never removed.

## Evidence

```js
// subscribe() — line 188 registers the wrapper, not the handler
const tracedHandler = (event) => { ... };
this.on(eventType, tracedHandler);

// unsubscribe() — line 203 removes the ORIGINAL fn, which is never registered
unsubscribe(eventType, handler) {
  if (typeof handler === 'function') {
    this.removeListener(eventType, handler);
  }
  return this;
}
```

The `EventHandler` (`.handle`) path has the same problem: line 194 registers `(event) => handler.handle(event)` and `unsubscribe` never matches it.

## Impact

- Any runtime teardown that calls `eventBus.unsubscribe(type, fn)` silently no-ops — the handler keeps firing on every publish.
- Repeated `subscribe(type, fn)` calls register duplicates, so the handler fires N times per event (listener leak up to `setMaxListeners(50)`).
- Contradicts the intended semantics tested in `backend/api/test/unit/core/eventBus.test.js:196-205` (asserts unsubscribe drops the handler).

## Proposed fix

Track the wrapper per original handler (e.g. a `WeakMap`), and in `unsubscribe` remove the stored wrapper:

```js
subscribe(eventType, handler) {
  ...
  const tracedHandler = ...;
  this._handlerWrappers.set(handler, tracedHandler);
  this.on(eventType, tracedHandler);
  ...
}
unsubscribe(eventType, handler) {
  if (typeof handler === 'function') {
    this.removeListener(eventType, this._handlerWrappers.get(handler) || handler);
  }
  ...
}
```

aaniya22:
## Description
`EventBus.subscribe()` wraps every function handler in a new `tracedHandler` closure (for OpenTelemetry span tracking) and registers that wrapper via `this.on(eventType, tracedHandler)`. But `unsubscribe()` called `this.removeListener(eventType, handler)` with the *original* handler reference, not the wrapper. Since `EventEmitter.removeListener` matches by strict reference identity, the registered wrapper was never actually removed — `unsubscribe()` silently no-op'd, leaving the handler firing on every future publish. The `EventHandler` (`.handle`) subscription path had the identical bug.

Added a `WeakMap` (`this._handlerWrappers`) to track each original handler → its wrapper closure, populated in both `subscribe()` branches (function handlers and `EventHandler` instances). `unsubscribe()` now looks up the stored wrapper via the WeakMap and removes that, falling back to the raw handler if no wrapper is found.

## Type of Change

## How Has This Been Tested?
- **Test Commands:** `node --check`, plus a standalone reproduction script isolating just the subscribe/unsubscribe/WeakMap logic
- **Test Steps / Prerequisites:** `test/unit/core/eventBus.test.js` has 13 pre-existing failures on `main` (before this change) from an unrelated `Cannot read properties of null (reading 'getTracer')` OpenTelemetry initialization issue in the test environment — confirmed by stashing this fix and re-running, same 13/21 failures occur either way, since that error blocks every `publish()` call including the `unsubscribe` test itself. Wrote an isolated script replicating the exact subscribe → emit → unsubscribe → emit sequence to verify the actual logic independent of that unrelated failure.
- **Expected Result:** `unsubscribe(eventType, fn)` actually removes the registered wrapper; the handler stops receiving events after unsubscribe. Confirmed: handler fired once before unsubscribe, did not fire again after.

### Test Environment
- [x] Local manual testing
- [ ] Unit / Integration tests (`npm test`, `flutter test`, etc.) — pre-existing unrelated OTel failures in this file, see note above

## Related Issues
Closes #8316

## PR Checklist
- [ ] I have commented my code, particularly in hard-to-understand areas.

### DIFF
--- a/backend/api/src/core/events/EventBus.js
+++ b/backend/api/src/core/events/EventBus.js
@@ -14,6 +14,7 @@ class EventBus extends EventEmitter {
     this._registry = new EventRegistry();
     this._deduplication = new Map();
     this._deduplicationWindowMs = 60000;
+    this._handlerWrappers = new WeakMap();
     this._metrics = {
       published: 0,
       subscribed: 0,
@@ -185,13 +186,16 @@ class EventBus extends EventEmitter {
           }
         });
       };
+      this._handlerWrappers.set(handler, tracedHandler);
       this.on(eventType, tracedHandler);
       return this;
     }
 
     if (handler && typeof handler.handle === 'function') {
       this._metrics.subscribed++;
-      this.on(eventType, (event) => handler.handle(event));
+      const tracedHandlerInstance = (event) => handler.handle(event);
+      this._handlerWrappers.set(handler, tracedHandlerInstance);
+      this.on(eventType, tracedHandlerInstance);
       return this;
     }
 
@@ -200,7 +204,7 @@ class EventBus extends EventEmitter {
 
   unsubscribe(eventType, handler) {
     if (typeof handler === 'function') {
-      this.removeListener(eventType, handler);
+      this.removeListener(eventType, this._handlerWrappers.get(handler) || handler);
     }
     return this;
   }


aaniya22:
Confirmed all shard 1 CI failures are pre-existing and unrelated to this diff.

This PR only touches `backend/api/src/core/events/EventBus.js` (tracking the wrapped handler reference via a `WeakMap` so `unsubscribe()` can find and remove the actual listener registered by `subscribe()`, instead of the raw un-wrapped handler which was never actually attached).

None of the 22 failing tests or 2 failed suites relate to EventBus:
- `test/unit/core/eventRegistry.test.js` passes cleanly (9/9)
- `EventSubscriber.test.js` and `LocalEventEmitterAdapter.test.js` both pass
- No failures reference EventBus, handler wrappers, or listener counts

This failure set is byte-for-byte identical to the one already verified as pre-existing on PR #8335/#8336 and confirmed against a clean `main` checkout:
- `osrm.test.js` (5) — coordinate rounding/null-input bugs in `osrm.js`, unrelated
- `orderTimelineService.test.js` (4) — broken import/mock, unrelated
- `idempotency.test.js` (2) — stale expected error-message string
- `staleOrderWorkerNotifications.test.js` (2), `withdrawalSettlementWorker.test.js` (3), `weighStationService.test.js` (6) — known flaky/broken pattern
- `truckSearchServiceRole.test.js`, `orderRoutesGeofence.test.js` (failed suites) — duplicate exports in `rateLimiter.js` and a separate vite transform error

This PR should be safe to review/merge on its own merits.

PR #8485, opened 2026-08-09
fix(api): pick up stale pending orders with NULL escrow_status for cancellation

### LINKED ISSUE #8470: [BUG][backend][P1] findStalePendingOrders excludes NULL escrow_status — non-escrow stale orders never auto-cancelled
## Description

`backend/api/src/repositories/orderRepository.js` `findStalePendingOrders()` (lines 570-577) filters with `.neq('escrow_status', 'funding')`. In PostgREST, `column.neq(value)` only matches rows where the column is **not equal AND not null**. Orders created before escrow support (or orders where `escrow_status` is simply never set) have `escrow_status = NULL`, so they never match and are excluded from stale-cancellation forever.

This means the `staleOrderWorker` can never cancel plain, non-escrow stale orders — they stay `pending` indefinitely.

## Expected behavior

Stale pending orders with no escrow involvement (escrow_status IS NULL) are also picked up and cancelled.

## Location

- `backend/api/src/repositories/orderRepository.js:570-577`

## Suggested fix

Match NULL as well using `.or()`:

```js
.or('escrow_status.is.null,escrow_status.neq.funding')
```

instead of:

```js
.neq('escrow_status', 'funding')
```

Xenon010101:
## Summary
`findStalePendingOrders()` (`backend/api/src/repositories/orderRepository.js:570-577`) filtered with `.neq('escrow_status', 'funding')`. PostgREST's `column.neq(value)` does not match `NULL`, so orders with `escrow_status IS NULL` (orders created before escrow support, or never escrow-funded) were never returned by the stale-cancellation query — the `staleOrderWorker` could never cancel plain non-escrow stale orders, leaving them `pending` forever.

## Changes
- Replaced `.neq('escrow_status', 'funding')` with `.or('escrow_status.is.null,escrow_status.neq.funding')` so both NULL and non-funding rows are selected.
- Updated `test/unit/repositories/orderRepositoryStaleCancel.test.js` to assert the new `or` filter.

Closes #8470

GSSOC 2026

### DIFF
--- a/backend/api/src/repositories/orderRepository.js
+++ b/backend/api/src/repositories/orderRepository.js
@@ -573,7 +573,7 @@ export class OrderRepository {
       .select('id')
       .eq('status', 'pending')
       .lt('created_at', cutoff)
-      .neq('escrow_status', 'funding')
+      .or('escrow_status.is.null,escrow_status.neq.funding')
       .limit(limit), 'findStalePendingOrders');
   }
 
--- a/backend/api/test/unit/repositories/orderRepositoryStaleCancel.test.js
+++ b/backend/api/test/unit/repositories/orderRepositoryStaleCancel.test.js
@@ -12,7 +12,7 @@ describe('OrderRepository stale-order cancellation', () => {
     orderRepository = new OrderRepository(supabaseMock.supabase);
   });
 
-  it('findStalePendingOrders selects only pending, older-than-cutoff, non-funding orders in a bounded batch', async () => {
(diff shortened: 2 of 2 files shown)


Xenon010101:
Superseded by PR #8507 (fix/8470-stale-order-null-escrow), which is rebased onto the current upstream main (3bd79950). This branch was based on the old, rewritten fork history (7b72c6f3) and is 364 commits stale. Closing this duplicate in favor of #8507.

PR #8489, opened 2026-08-09
fix(api): propagate critical flag in HealthAggregator so outages report unhealthy

### LINKED ISSUE #8472: [BUG][backend][P1] HealthAggregator drops the critical flag — database outage reported as 'degraded' instead of 'unhealthy'
## Description

`backend/api/src/core/health/HealthAggregator.js` never propagates the `critical` flag registered per check:

- `register()` stores `critical` (line 45-47), but `aggregate()` destructures only `{ name, checkFn }` (line 58), discarding it for the success path.
- The catch block hardcodes `critical: false` (line 70).

Because `_determineOverallStatus()` only escalates to `unhealthy` when `r.critical && r.status === UNHEALTHY` (lines 118-121), a critical service (supabase, postgres, mongodb are all registered `critical: true` in `core/health/index.js:25-27`) that throws or reports unhealthy is instead reported as `degraded`.

## Expected behavior

When a check registered with `{ critical: true }` throws or reports UNHEALTHY, the aggregated status is `unhealthy`, not `degraded`.

## Location

- `backend/api/src/core/health/HealthAggregator.js:58,70`
- critical registrations: `backend/api/src/core/health/index.js:25-27`

## Suggested fix

Destructure and forward `critical`, and use the registered flag in the catch path:

```js
this._checks.map(async ({ name, checkFn, critical }) => {
  try {
    return await checkFn();
  } catch (err) {
    logger.error(`[health] Aggregator check "${name}" threw: ${err.message}`);
    return {
      name,
      status: HealthStatus.UNHEALTHY,
      message: err.message,
      responseTime: 0,
      critical: !!critical,
      timestamp: new Date().toISOString(),
    };
  }
});
```

Xenon010101:
## Summary
`HealthAggregator.aggregate()` never propagated the `critical` flag registered per check:
- The map callback destructured only `{ name, checkFn }`, discarding `critical` on the success path, and
- the catch block hardcoded `critical: false`.

Since `_determineOverallStatus()` only escalates to `unhealthy` when `r.critical && r.status === UNHEALTHY`, a critical service (supabase, postgres, mongodb are registered `critical: true` in `core/health/index.js`) that threw or reported unhealthy was reported as `degraded` instead of `unhealthy`.

## Changes
- Forward the registered `critical` flag into both the success and catch paths of `aggregate()`, making the aggregator registration the source of truth.
- Added `test/unit/healthAggregator.test.js` covering: critical check throw → `unhealthy`, non-critical throw → `degraded`, and registered-critical propagation on a successful (but unhealthy) check.

Closes #8472

GSSOC 2026

### DIFF
--- a/backend/api/src/core/health/HealthAggregator.js
+++ b/backend/api/src/core/health/HealthAggregator.js
@@ -55,9 +55,14 @@ export class HealthAggregator {
     const start = Date.now();
 
     const results = await Promise.all(
-      this._checks.map(async ({ name, checkFn }) => {
+      this._checks.map(async ({ name, checkFn, critical }) => {
         try {
-          return await checkFn();
+          const result = await checkFn();
+          return {
+            ...result,
+            name,
+            critical: !!critical,
+          };
         } catch (err) {
           logger.error(
             `[health] Aggregator check "${name}" threw: ${err.message}`,
@@ -67,7 +72,7 @@ export class HealthAggregator {
             status: HealthStatus.UNHEALTHY,
             message: err.message,
             responseTime: 0,
-            critical: false,
+            critical: !!critical,
             timestamp: new Date().toISOString(),
           };
         }
--- a/backend/api/test/unit/healthAggregator.test.js
+++ b/backend/api/test/unit/healthAggregator.test.js
@@ -1,0 +1,51 @@
+import { describe, expect, it, vi } from 'vitest';
+
+vi.mock('../../src/middleware/logger.js', () => ({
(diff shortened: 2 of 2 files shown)


Xenon010101:
Closing: issue #8472 was already fixed on the current upstream main by commit 98cc5eaf (HealthAggregator destructures and propagates critical on all paths). This branch is also based on the old, rewritten fork history (7b72c6f3, 364 commits stale). No code change needed.

PR #8493, opened 2026-08-09
fix(api): normalizePhone strips leading 0 for national-dialing numbers

### LINKED ISSUE #8474: [BUG][backend][P1] normalizePhone returns null for leading-zero numbers like 0919876543210 despite documented support
## Description

`backend/api/src/utils/phone.js` `normalizePhone()` documents support for `0919876543210` (line 7), but the implementation never strips the leading `0` before the country-code check. For `0919876543210`, `digits` is `0919876543210` (13 chars), which does not start with `91` (it starts with `0`), so the country-code removal branch is skipped and the length check `digits.length !== 10` returns `null`.

Runtime-verified:
```
0919876543210 -> null        (should be +919876543210)
+91 9876543210 -> +919876543210
919876543210  -> +919876543210
```

Callers such as login/OTP flows that receive numbers entered with a leading `0` will reject them.

## Expected behavior

`normalizePhone('0919876543210')` returns `+919876543210`.

## Location

- `backend/api/src/utils/phone.js:24-30`

## Suggested fix

Strip a leading `0` before the country-code check:

```js
let digits = phone.replace(/[^\d]/g, '');
if (digits.startsWith('0')) digits = digits.slice(1);
if (digits.startsWith('91') && digits.length === 12) {
  digits = digits.slice(2);
}
if (digits.length !== 10) return null;
```

Xenon010101:
## Summary
`normalizePhone()` (`backend/api/src/utils/phone.js`) documents support for `0919876543210` (leading-zero national dialing), but the implementation only strips the `91` country code when the digit string starts with `91`. A leading-zero number like `0919876543210` produced 13 digits starting with `0`, so the country-code branch was skipped and the length check returned `null` — login/OTP flows rejected numbers entered with a leading `0`.

## Changes
- Strip a single leading `0` before the country-code check (handles `0919876543210` → `+919876543210`).
- Added `test/unit/phone.test.js` covering leading-zero, `+91`-separated, bare `91`-prefixed, plain 10-digit, already-E.164, too-short, and non-string inputs.

Closes #8474

GSSOC 2026

### DIFF
--- a/backend/api/src/utils/phone.js
+++ b/backend/api/src/utils/phone.js
@@ -20,6 +20,11 @@ export function normalizePhone(phone) {
   const isInternational = phone.trim().startsWith('+');
   let digits = phone.replace(/[^\d]/g, '');
 
+  // Strip a leading 0 used for national dialing (e.g. 0919876543210)
+  if (digits.startsWith('0')) {
+    digits = digits.slice(1);
+  }
+
   // Remove country code prefix if present (91 for India)
   if (digits.startsWith('91') && digits.length === 12) {
     digits = digits.slice(2);
--- a/backend/api/test/unit/phone.test.js
+++ b/backend/api/test/unit/phone.test.js
@@ -1,0 +1,34 @@
+import { describe, expect, it } from 'vitest';
+
+import { normalizePhone } from '../../src/utils/phone.js';
+
+describe('normalizePhone', () => {
+  it('normalizes a leading-zero national number to E.164', () => {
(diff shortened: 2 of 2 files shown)


Xenon010101:
Superseded by PR #8515 (fix/8474-normalize-phone-leading-zero), which is rebased onto the current upstream main (3bd79950). This branch was based on the old, rewritten fork history (7b72c6f3) and is 364 commits stale. Closing this duplicate in favor of #8515.

PR #7812, opened 2026-08-07
fix(api): remove duplicate rate limiter exports that break module load

nyxsky404:
## Description
Duplicate named exports for verifyDeliveryLimiter, resendOtpLimiter, changeDropLimiter, predictDemandLimiter, and telemetryLimiter crash module load. Removed the weaker first block and kept the stricter per-user+order limiters.

## Type of Change
- [x] Bug fix (non-breaking change fixing an issue)

## How Has This Been Tested?
- **Test Commands:** node --check backend/api/src/middleware/rateLimiter.js
- **Test Steps / Prerequisites:** n/a
- **Expected Result:** module parses

### Test Environment
- [x] Local manual testing
- [ ] Unit / Integration tests (`npm test`, `flutter test`, etc.)
- [ ] Tested via Docker Compose

## Related Issues
Closes #7792

## PR Checklist
- [x] My code follows the code style guidelines of this project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated corresponding documentation if applicable.
- [x] My changes generate no new warnings or console errors.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rate limiting for delivery verification, OTP resend, drop changes, demand predictions, and telemetry requests.
  * Added order-specific controls to help prevent excessive retries and repeated requests.
  * Updated retry guidance to reflect revised request limits and waiting periods.
  * Applied more consistent protections based on account, network, and order activity.

PR #8532, opened 2026-08-09
[Bug] Fix out-of-scope 'result' in checkEscrow that crashed every health probe

### LINKED ISSUE #8530: [BUG][backend][P0] healthRoutes checkEscrow reads an out-of-scope 'result' — ReferenceError crashes every health probe
## Summary

`checkEscrow()` in `backend/api/src/routes/healthRoutes.js` contains two nested copies of the same `try`/`catch`, with `const result` declared inside the inner `try` and read after both blocks have closed. `result` is not in scope at the `return`, so the function throws `ReferenceError` on every call.

## Reproduction

```
$ cd backend/api
$ npx eslint src/routes/healthRoutes.js
  135:10  error  'result' is not defined  no-undef
```

## Details

```js
async function checkEscrow() {
  try {
  try {
  const result = await checkEscrowHealth();      // block-scoped to the inner try
  } catch (err) {
    logger.error('[Health] checkEscrow failed:', err?.message || err);
    return { status: 'failed', detail: err?.message || 'Unknown error' };
  }
  } catch (err) {                                 // identical duplicate
    logger.error('[Health] checkEscrow failed:', err?.message || err);
    return { status: 'failed', detail: err?.message || 'Unknown error' };
  }
  return result.status;                           // ReferenceError: result is not defined
}
```

Three things are wrong at once:

1. `result` is `const`-scoped to the inner `try` and referenced outside it.
2. The `try` blocks are duplicated verbatim, so the outer one can never catch anything the inner one did not already handle.
3. The value assigned to `result` is discarded even on the success path — the inner `try` has no `return`.

The `ReferenceError` is thrown *after* both `catch` blocks, so nothing catches it. It propagates out of `checkEscrow()` into the `Promise.all` at line 169 and rejects the whole health handler.

There is a second, subtler mismatch: the two `catch` blocks return an **object** (`{ status, detail }`), but the caller uses the return value as a plain status string — `escrow: escrowStatus` at line 176, compared against `CRITICAL_UNHEALTHY`, a `Set` of strings. An object would never match, so even the error path reports the wrong thing.

## Impact

The health endpoint is the readiness/liveness probe. With `checkEscrow()` throwing unconditionally, `Promise.all` rejects and the handler 500s or hangs on **every** request — so orchestrators see the service as permanently unhealthy and will not route traffic to it, or will restart-loop it.

## Relationship to #8100

PR #8100 fixes this function, but against the pre-merge body:

```js
-  const result = await checkEscrowHealth();
-  return result.status;
+  try { ... } catch (err) { ... return 'unavailable'; }
```

The state on `main` today looks like a botched merge of that change — the new `try`/`catch` was applied twice and the original two-line body left in between. PR #8100 will not apply cleanly to it. Worth resolving whichever way the maintainers prefer; flagging so the two are not fixed in conflict.

## Expected

`checkEscrow()` returns a status **string**, and a thrown health check degrades the report instead of crashing the endpoint.

## Suggested fix

Collapse to a single `try`/`catch`, return `result?.status ?? 'unavailable'` on success and the `'failed'` string on error, so the value matches what `CRITICAL_UNHEALTHY` compares against.

advikdivekar:
Fixes #8530.

`checkEscrow()` had **two nested copies** of the same `try`/`catch`, with `const result` declared inside the inner `try` and read after both blocks closed.

### Before

```js
async function checkEscrow() {
  try {
  try {
  const result = await checkEscrowHealth();      // block-scoped to the inner try
  } catch (err) { ... return { status: 'failed', detail: ... }; }
  } catch (err) { ... return { status: 'failed', detail: ... }; }   // identical duplicate
  return result.status;                           // ReferenceError
}
```

```
$ npx eslint src/routes/healthRoutes.js
  135:10  error  'result' is not defined  no-undef
```

The `ReferenceError` is thrown **after** both `catch` blocks, so nothing catches it. It propagates out of the `Promise.all` at line 169 and rejects the whole health handler — the readiness/liveness probe fails on **every** request, which reads to an orchestrator as a permanently unhealthy service.

### Two bugs, not one

Also fixed the return type. Both `catch` blocks returned an **object** `{ status, detail }`, but the caller uses the value as a plain status **string**:

```js
escrow: escrowStatus,                       // line 176
CRITICAL_UNHEALTHY.has(escrowStatus)        // a Set of strings
```

An object could never match, so the error path reported the wrong thing even when reached. Now returns `'failed'` on error and `result?.status ?? 'unavailable'` on success.

### Relationship to #8100 — worth a look before merging

PR #8100 fixes this same function, but against the pre-merge two-line body (`const result = await checkEscrowHealth(); return result.status;`). The state on `main` looks like a **botched merge of that very change** — the new `try`/`catch` applied twice with the original body left in between — so #8100 will not apply cleanly as things stand. Happy to close this in favour of a rebased #8100 if you prefer; flagging so the two are not fixed in conflict.

### Verification

`npx eslint src/routes/healthRoutes.js` — clean.

The health suites cannot run locally: they stop at collection on the unrelated duplicate `verifyDeliveryLimiter` in `rateLimiter.js` (#7812). I could not execute them, so I am not claiming test coverage here — the change is 3 insertions, 8 deletions and purely a scope/return-type correction.

### DIFF
--- a/backend/api/src/routes/healthRoutes.js
+++ b/backend/api/src/routes/healthRoutes.js
@@ -122,17 +122,12 @@ function checkFirebase() {
 
 async function checkEscrow() {
   try {
-  try {
-  const result = await checkEscrowHealth();
+    const result = await checkEscrowHealth();
+    return result?.status ?? 'unavailable';
   } catch (err) {
     logger.error('[Health] checkEscrow failed:', err?.message || err);
-    return { status: 'failed', detail: err?.message || 'Unknown error' };
+    return 'failed';
   }
-  } catch (err) {
-    logger.error('[Health] checkEscrow failed:', err?.message || err);
-    return { status: 'failed', detail: err?.message || 'Unknown error' };
-  }
-  return result.status;
 }
 
 function checkPolygon() {


advikdivekar:
Superseded — main already has an equivalent fix for the checkEscrow scoping bug (returns `result.status` inside the same try/catch, so an undefined result still safely falls through to the catch's 'failed'). Rebasing this branch onto main resolves to zero diff; healthAggregation integration tests pass as-is (9/9). Closing rather than merging a no-op.

Issue #8560, opened 2026-08-09
[BUG][blockchain][P0] hardhat compile fails on a clean checkout — solc 0.8.24 targets shanghai but OpenZeppelin 5.x needs MCOPY (cancun)


advikdivekar:
## Summary

`npx hardhat compile` fails on a clean checkout. `hardhat.config.cjs` pins solc `0.8.24` without an `evmVersion`, so it targets `shanghai` — which has no `MCOPY`. OpenZeppelin Contracts 5.x emits `mcopy` in assembly, so nothing in `blockchain/` compiles.

## Reproduction

```
$ cd blockchain
$ npx hardhat compile

DeclarationError: Function "mcopy" not found.
   --> @openzeppelin/contracts/utils/Bytes.sol:168:13:
    |
168 |             mcopy(add(add(buffer, 0x20), pos), add(add(replacement, 0x20), offset), length)
    |             ^^^^^

DeclarationError: Function "mcopy" not found.
   --> @openzeppelin/contracts/utils/Bytes.sol:195:17
...
Error HH600: Compilation failed
```

## Why it is not a local-environment problem

`package.json` declares `"@openzeppelin/contracts": "^5.6.1"` and `package-lock.json` pins `5.6.1`, so a clean `npm ci` installs the version whose `Bytes.sol` uses `mcopy`. The config asks for solc `0.8.24`, whose default EVM target is `shanghai`; `MCOPY` was introduced in `cancun`. The two are simply incompatible as configured — nothing about the local machine is involved.

```js
// blockchain/hardhat.config.cjs
solidity: {
  version: "0.8.24",
  settings: {
    optimizer: { enabled: true, runs: 200 },
    viaIR: true,
    // no evmVersion
  },
},
```

## Impact

The whole `blockchain/` workspace is unbuildable, which means:

- `TruxifyEscrow`, `Reputation`, `DocumentRegistry`, the DAO, the multi-sig and every other contract — 70 files — cannot be compiled;
- **CI is red**: `.github/workflows/ci.yml:187` runs `npx hardhat compile` and `:191` runs `npx hardhat test`, so both fail;
- no contract can be tested, deployed or verified, and the escrow ABI check in `scripts/check-escrow-abi.sh` has nothing to check against.

## Expected

`npx hardhat compile` succeeds.

## Suggested fix

[...]

PR #8563, opened 2026-08-09
[Bug] Align package type with the ESM tests so hardhat test actually runs

### LINKED ISSUE #8562: [BUG][blockchain][P1] type:commonjs vs 9 ESM test files — hardhat test aborts, no contract test runs
## Summary

`blockchain/package.json` declares `"type": "commonjs"`, but 9 of the 14 test files use ESM `import`. Mocha loads them as CommonJS, the first one throws `SyntaxError`, and Hardhat aborts the entire run — **no contract test executes at all**.

## Reproduction

```
$ cd blockchain
$ npx hardhat test

Warning: Failed to load the ES module: blockchain/test/AssetToken.test.js.
Make sure to set "type": "module" in the nearest package.json file or use the .mjs extension.

An unexpected error occurred:
blockchain/test/AssetToken.test.js:1
import assert from "node:assert/strict";
^^^^^^
SyntaxError: Cannot use import statement outside a module
```

This is not a per-file failure that the runner skips past — it is an abort. `TruxifyEscrow`, `Reputation`, `MEVProtectedEscrow`, `StateChannel`, `DIDRegistry`, `IdentityWallet`, `zkVerifier`, `escrow` and `AssetToken` all go untested, including the `.cjs` suites that would otherwise have run.

## The module story is inconsistent across the workspace

| | ESM | CommonJS |
|---|---|---|
| `test/*.js` | 9 | 1 (`EnterpriseMultiSig.test.js`) |
| `test/*.cjs` | — | 4 |
| `scripts/*.js` | 3 (`deploy`, `deploy-upgradeable`, `deploy-zkevm`) | 1 (`generate-zk-proof`) |

`package.json` says `commonjs`, so the majority of files in the workspace are mislabelled. The four `.cjs` test files look like a previous attempt to work around this one file at a time.

## Impact

The blockchain test suite has effectively not run. Combined with #8560 — where `hardhat compile` fails outright — the contracts are neither compiled nor tested in CI, despite `.github/workflows/ci.yml:187` and `:191` invoking both.

## Expected

`npx hardhat test` runs the suite.

## Suggested fix

Set `"type": "module"` to match where the workspace actually is, and rename the two CommonJS stragglers (`test/EnterpriseMultiSig.test.js`, `scripts/generate-zk-proof.js`) to `.cjs`. `hardhat.config.cjs` is already `.cjs`, and the four existing `.cjs` tests are unaffected.

### Depends on #8560

The suite cannot run until compilation is fixed. With both applied: **149 passing, 94 failing** — up from zero executing. Those 94 are pre-existing and unrelated to loading; 80 share a single cause (`TruxifyEscrow.createBooking` gained a `bytes signature` parameter that the tests were never updated for) and are worth their own issue.

advikdivekar:
Fixes #8562.

`blockchain/package.json` declares `"type": "commonjs"`, but 9 of the 14 test files use ESM `import`. Mocha loads them as CommonJS and the first throws — **aborting the entire run**.

### Before

```
$ cd blockchain && npx hardhat test

An unexpected error occurred:
blockchain/test/AssetToken.test.js:1
import assert from "node:assert/strict";
^^^^^^
SyntaxError: Cannot use import statement outside a module
```

Not a skipped file — an abort. `TruxifyEscrow`, `Reputation`, `MEVProtectedEscrow`, `StateChannel`, `DIDRegistry`, `IdentityWallet`, `zkVerifier`, `escrow` and `AssetToken` all went untested, **including the four `.cjs` suites** that would otherwise have run.

### The workspace had drifted past its declared type

| | ESM | CommonJS |
|---|---|---|
| `test/*.js` | 9 | 1 (`EnterpriseMultiSig.test.js`) |
| `test/*.cjs` | — | 4 |
| `scripts/*.js` | 3 | 1 (`generate-zk-proof`) |

The four `.cjs` test files look like a previous attempt to work around this one file at a time.

### Change

Set `"type": "module"` to match where the code actually is, and rename the two CommonJS stragglers to `.cjs`. `hardhat.config.cjs` is already `.cjs`; the existing `.cjs` tests are untouched.

One `package.json` line + two renames — 0 content lines changed.

### ⚠️ Depends on #8561 (issue #8560)

The suite cannot run until compilation is fixed. **With both applied:**

```
  149 passing (2s)
  94 failing
```

Up from **zero executing**. The 94 are pre-existing and unrelated to module loading — 80 share one cause (`TruxifyEscrow.createBooking` gained a `bytes signature` parameter the tests were never updated for). I have written that up separately rather than folding an 80-test rewrite into a loader fix.


advikdivekar:
Rebased on `main` and updated: `test/DAO.test.js` landed after this PR was opened and uses CommonJS `require`, so it would have broken under `"type": "module"` —

```
ReferenceError: require is not defined in ES module scope
  at blockchain/test/DAO.test.js:1:20
```

Renamed it to `.cjs` alongside the other two stragglers. I re-scanned the whole `blockchain/test` and `blockchain/scripts` tree after merging: `DAO.test.js` was the only newly-affected file.

Flagging because this PR is sensitive to new CommonJS `.js` files landing in `blockchain/` while it is open — if more arrive before merge, ping me and I will fold them in.

PR #8845, opened 2026-08-09
fix(pqc): Kyber1024Relayer decapsulate now recovers the encapsulated shared secret

### LINKED ISSUE #8798: [BUG][backend][P1] pqc kyber_relayer decapsulate never returns the encapsulate shared secret - KEM agreement always fails, test misses the assertion
## Summary

`Kyber1024Relayer.decapsulate()` never returns the shared secret produced by `encapsulate()`. Two parties can never agree on a key — the KEM handshake is guaranteed to fail. The unit test does not catch this because it never asserts that both sides derive the same secret.

## Files

- `backend/pqc/kyber_relayer.py` (lines 24-42)
- `backend/pqc/test_kyber.py` (lines 13-20)

## Details

`encapsulate(pk)`:
```python
shared_secret = os.urandom(32)                       # random secret
ciphertext = hashlib.sha3_512(shared_secret + pk[:64]).digest() * 39
return ciphertext, shared_secret
```

`decapsulate(ct, sk)`:
```python
derived_ss = hashlib.sha256(ct[:32] + sk[:32]).digest()
return derived_ss
```

`derived_ss` is a deterministic function of `ct[:32]` and `sk[:32]`; it has no relationship to the random `shared_secret` chosen in `encapsulate`. Therefore `decapsulate(encapsulate(pk)) != shared_secret` (with overwhelming probability). A real ML-KEM decapsulation must recover the exact secret that was encapsulated (e.g., via the FO transform / message recovery).

`test_kyber.py::test_encapsulate_decapsulate` even computes `ss2 = self.relayer.decapsulate(ct, sk)` but never asserts `ss1 == ss2`, so the brokenness goes unnoticed.

## Impact

Any consumer using this relayer (off-chain relayer signatures / PQC key agreement) can never establish a shared session key; handshakes always fail silently or produce mismatched keys.

## Repro

```python
r = Kyber1024Relayer()
pk, sk = r.generate_keypair()
ct, ss1 = r.encapsulate(pk)
ss2 = r.decapsulate(ct, sk)
assert ss1 == ss2   # AssertionError, always
```

## Suggested fix

Implement a real Kyber decapsulation that recovers the encapsulated message/secret (both parties must derive the same value from the ciphertext and the private key), and add `self.assertEqual(ss1, ss2)` to the test.

ionfwsrijan:
## Problem

`Kyber1024Relayer.decapsulate()` never returned the shared secret produced by `encapsulate()`:

- `encapsulate(pk)` picks a random `shared_secret` and returns it with a ciphertext.
- `decapsulate(ct, sk)` derived a deterministic `sha256(ct[:32] + sk[:32])` that has no relationship to the random `shared_secret`, so `decapsulate(encapsulate(pk)) != shared_secret` with overwhelming probability.
- Two parties could never agree on a key — the KEM handshake always failed.
- `test_kyber.py::test_encapsulate_decapsulate` computed `ss2` but never asserted `ss1 == ss2`, so the brokenness went unnoticed. The key-generation test also asserted the secret-key length against a value the old generator never produced.

## Fix

- `generate_keypair()` now embeds the seed at the front of the secret key, so the public key can be reconstructed from the private key.
- `encapsulate(pk)` chooses a random message `m`, derives `shared_secret = H(pk || m)`, and encrypts `m` under a key derived from the public key into the ciphertext.
- `decapsulate(ct, sk)` reconstructs the public key from the seed inside the secret key, recovers the encapsulated message from the ciphertext, and re-derives the exact same `shared_secret` — both parties now derive the same value.
- The test now asserts `ss1 == ss2` (and fixes the `self.toEqual` typo so the public-key length is actually asserted).

## Files changed

- `backend/pqc/kyber_relayer.py`
- `backend/pqc/test_kyber.py`

## Testing

- `py -m unittest test_kyber -v` passes both tests:
  - `test_keypair_generation` — ok
  - `test_encapsulate_decapsulate` — ok (including the new `ss1 == ss2` assertion)

Closes #8798

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Kyber key encapsulation and decapsulation so both sides consistently derive the same shared secret.
  * Ensured generated key material uses the required lengths and preserves the information needed for successful recovery.
  * Corrected validation of public-key sizes and shared-secret matching.

* **Tests**
  * Expanded coverage for key lengths and successful encapsulation/decapsulation results.

### DIFF
--- a/backend/pqc/kyber_relayer.py
+++ b/backend/pqc/kyber_relayer.py
@@ -1,5 +1,4 @@
 import os
-import secrets
 import hashlib
 
 class Kyber1024Relayer:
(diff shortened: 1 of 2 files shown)


Nareshkumawat-star:
hi @KanishJebaMathewM kindly and merge the pr soon

PR #8925, opened 2026-08-09
fix: use PostGIS ST_DWithin index for nearby-driver search instead of full-table scan

### LINKED ISSUE #8723: bug: findTargetDrivers does full active-fleet scan instead of using existing PostGIS index
**Describe the bug**
`findTargetDrivers` in `backend/api/src/services/order/orderCreationService.js` (lines 43-89) fetches every driver nationwide with a fresh location update (`is_active = true`, `last_updated_at` within the freshness window) and then filters by distance in JavaScript using `haversineDistanceKm`. This ignores the `driver_locations` table's existing `idx_driver_locations_location` GIST index (added in migration `20260727120000_add_postgis_geospatial_indexes.sql`), which was built for exactly this radius-search use case but has never been queried via `ST_DWithin`/`ST_Distance` anywhere in `backend/api`.

This function is called from `sendNewTripNotifications` on every single order creation, so the inefficiency scales with total active driver count rather than with drivers actually near the pickup point.

**To Reproduce**
1. Open `backend/api/src/services/order/orderCreationService.js`
2. See `findTargetDrivers` (~line 43-89): the Supabase query has no spatial filter
3. See the JS-side `.filter(loc => haversineDistanceKm(...) <= NEW_TRIP_NOTIFY_RADIUS_KM)` immediately after
4. Compare with `truckRoutes.js` (~line 567-583), which correctly uses a geo-indexed `$near` query for the analogous customer-facing truck search

**Expected behavior**
The query should push both the freshness and radius filters into Postgres via `ST_DWithin` against the existing indexed geography column, returning only nearby drivers directly from the database rather than pulling the entire active fleet over the wire.

**Screenshots**
N/A — backend logic issue, no UI involved.

**Environment**
- N/A (backend/api, not client-specific)

**Additional context**
The index and its supporting migration already exist and are unused — this is a "wire up what's already built" fix, not new infrastructure.

jess1805:
## Problem
findTargetDrivers() in orderCreationService.js fetched every active driver nationwide from driver_locations on every order creation, then filtered by haversine distance in JS. A GIST index for exactly this radius search already exists (idx_driver_locations_location, added in 20260727120000_add_postgis_geospatial_indexes.sql) but was never queried via ST_DWithin.

## Fix
- Added a new RPC, get_nearby_active_drivers, that runs ST_DWithin against the indexed driver_locations.location geography column so the GIST index is actually used and only matching rows are returned.
- SECURITY DEFINER, granted to service_role only — mirrors the existing complete_trip_tx pattern.
- Updated findTargetDrivers() to call the RPC via supabaseAdmin instead of pulling all rows and filtering in JS. Removed the now-unused haversineDistanceKm helper.
- Also switched the driver_details and trucks lookups in the same function from the anon supabase client to supabaseAdmin. Note: driver_locations has zero anon privileges (see revoke_anon_privileges.sql), so the original anon-client query was already broken independent of the perf issue — this fix addresses both.

## Testing
Added backend/api/test/unit/orderCreationService.findTargetDrivers.test.js, covering:
- the RPC is called with correct params on the service-role client (not anon)
- truck-capacity filtering still works
- RPC error path returns [] without throwing

Ran locally: npx vitest run backend/api/test/unit/orderCreationService.findTargetDrivers.test.js — both tests pass.

## Files changed
- supabase/migrations/20260809000000_add_get_nearby_active_drivers_rpc.sql (new)
- backend/api/src/services/order/orderCreationService.js
- backend/api/test/unit/orderCreationService.findTargetDrivers.test.js (new)

Closes #8723

## Summary by CodeRabbit

* **Improvements**
  * Nearby driver matching now more reliably considers location freshness, distance, and truck capacity.
  * Driver searches gracefully return no results when availability information cannot be retrieved.
* **Security**
  * Nearby-driver lookup access is restricted to authorized service operations.
* **Tests**
  * Added coverage for nearby-driver filtering, capacity requirements, and lookup failures.

### DIFF
--- a/backend/api/src/services/order/orderCreationService.js
+++ b/backend/api/src/services/order/orderCreationService.js
@@ -21,45 +21,37 @@ const NEW_TRIP_NOTIFY_BATCH_SIZE = Number(process.env.NEW_TRIP_NOTIFY_BATCH_SIZE
   : 25;
 const DRIVER_LOCATION_FRESHNESS_MS = 15 * 60 * 1000;
 
-function haversineDistanceKm(lat1, lng1, lat2, lng2) {
-  const toRad = deg => (deg * Math.PI) / 180;
-  const R = 6371;
-  const dLat = toRad(lat2 - lat1);
-  const dLng = toRad(lng2 - lng1);
-  const a = Math.sin(dLat / 2) ** 2 +
-    Math.cos(toRad(lat1)) * Math.cos(toRad(lat2)) * Math.sin(dLng / 2) ** 2;
-  return 2 * R * Math.asin(Math.sqrt(a));
-}
-
 /**
  * Find drivers that should be notified about a new trip: those online and
  * located within the configured radius of the pickup, whose truck can carry
- * the load (by weight capacity). Falls back to in-memory filtering when the
- * drivers are not filterable in a single SQL query.
+ * the load (by weight capacity).
+ *
+ * The radius+freshness search is pushed into Postgres via the
+ * get_nearby_active_drivers RPC, which runs ST_DWithin against the indexed
+ * driver_locations.location geography column (idx_driver_locations_location)
+ * instead of pulling every active driver nationwide and filtering by
+ * haversine distance in JS.
  *
  * @param {{pickupLat: number, pickupLng: number, weightTonnes: number}} args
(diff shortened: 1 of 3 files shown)


jess1805:
I am part of GSSOC

PR #9395, opened 2026-08-10
[Bug] Restore the supersede lookup and limits documentController relies on — every KYC upload was 500ing

### LINKED ISSUE #8581: fix : add missing constants import to documentController
documentController.js uses three symbols without importing or defining them: MAX_FILE_SIZE_BYTES (lines 46, 48 — used for file size guard), SCAN_TIMEOUT_MS (lines 71, 97 — used for malware scan timeout), and existingDoc (lines 143, 153, 194, 197, 203 — used in duplicate-document logic). All three cause ESLint no-undef errors and ReferenceError at runtime.

Changes that Need to be Made:
- backend/api/src/controllers/documentController.js: add the missing constant definitions to the import/const block at the top of the file:
  - const MAX_FILE_SIZE_BYTES = 10 * 1024 * 1024; // 10MB
  - const SCAN_TIMEOUT_MS = 30 * 1000; // 30s

For existingDoc, it is a local variable assigned inside the upload handler — the ESLint error may be a false positive from the eslint no-use-before-define rule, or the variable needs to be declared where it is used. Verify and declare as needed in the function scope where it is referenced.

Impact that it would Provide:
- Eliminates 3 no-undef ESLint errors in documentController.js
- Fixes potential ReferenceError at runtime for file size guard and malware scan timeout
- Ensures the duplicate-document detection logic has its variable properly scoped

### LINKED ISSUE #9394: [BUG][backend][P0] documentController references an undefined `existingDoc` — every driver KYC upload 500s, and 6 of the 9 documentRoutes tests already fail
### Summary

`backend/api/src/controllers/documentController.js` references three identifiers that are declared nowhere in the module. The supersede lookup that was supposed to define `existingDoc` is simply missing — a merge kept the code that *consumes* it and dropped the code that *produces* it.

```
$ cd backend/api && ../../node_modules/.bin/eslint src/controllers/documentController.js

   46:25  error  'MAX_FILE_SIZE_BYTES' is not defined  no-undef
   46:74  error  'MAX_FILE_SIZE_BYTES' is not defined  no-undef
   48:66  error  'MAX_FILE_SIZE_BYTES' is not defined  no-undef
   71:60  error  'SCAN_TIMEOUT_MS' is not defined      no-undef
   97:48  error  'SCAN_TIMEOUT_MS' is not defined      no-undef
  143:9   error  'existingDoc' is not defined          no-undef
  153:19  error  'existingDoc' is not defined          no-undef
  194:9   error  'existingDoc' is not defined          no-undef
  197:18  error  'existingDoc' is not defined          no-undef
  203:23  error  'existingDoc' is not defined          no-undef
```

The missing constants are already filed as #8581. **`existingDoc` is not covered by any open issue**, and it is the more serious of the two: the constants are a fail-fast at the top of the handler, whereas `existingDoc` is read *after the file has already been written to Supabase Storage*.

### Impact

`uploadDriverDocument` is wrapped in a `try/catch` that maps anything unexpected to `500 { error: 'An unexpected error occurred' }`. So:

1. **Every driver KYC document upload fails with a 500.** Aadhaar, PAN, licence, RC book — the whole KYC flow is dead.
2. The `existingDoc` ReferenceError fires at line 143, *after* the `supabase.storage.upload()` at line 128 has succeeded. The storage-cleanup path only runs on `dbError`, and a `ReferenceError` is not a `dbError`. **Every attempt leaves an orphaned file in the private `driver-documents` bucket**, with no metadata row pointing at it and nothing to garbage-collect it.
3. Because the constants throw first at line 46, even the 400/422 validation responses are wrong — a bad `documentType` returns 500 instead of 400.

### This is already visible in the existing test suite

`test/integration/documentRoutes.test.js` exists and **6 of its 9 tests fail on `main` right now**:

```
$ ../../node_modules/.bin/vitest run --silent=true --reporter=dot test/integration/documentRoutes.test.js

 FAIL  ... returns 400 for an unrecognized documentType
   AssertionError: expected 500 to be 400
 FAIL  ... accepts a real JPEG declared as image/jpeg and stores it
   AssertionError: expected 500 to be 201
 FAIL  ... rejects an executable renamed to .jpg with a 422 and does not store it
   AssertionError: expected 500 to be 422
 FAIL  ... rejects a file whose real content does not match its declared Content-Type
   AssertionError: expected 500 to be 422
 FAIL  ... returns 500 if storage upload fails and does not expose internal error details
 FAIL  ... rejects upload when malware scanner detects a threat (422)
   AssertionError: expected 500 to be 422

 Tests  6 failed | 3 passed (9)
```

Note what the third and sixth failures mean: the tests asserting that **a renamed executable is rejected** and that **a malware-flagged file is rejected** are currently red. The magic-byte and malware defences are not verified at all on `main`. This is the same "suites exist but never run in CI" gap as #8551.

### What the missing lookup has to do

The surviving code reads `existingDoc.id` (line 153) and `existingDoc.storage_path` (line 197), and returns `200` vs `201` off its truthiness — so the intent is clear: **re-uploading the same document type supersedes the previous one**, updating the metadata row in place and deleting the old file.

Two details worth getting right:

- `driver_documents` has **no unique constraint** on `(driver_id, document_type)` (see `supabase/migrations/20260702000000_create_driver_documents.sql`), so the lookup must order newest-first and take one row rather than assume uniqueness. The `23505` handler at line 183 is aspirational.
- The cleanup at line 197 reads `existingDoc.storage_path` **after** the update has rewritten that column. The path to delete must be captured *before* the update, or the handler deletes the file it just uploaded and leaves the old one behind.

### Proposal

Define `MAX_FILE_SIZE_BYTES` and `SCAN_TIMEOUT_MS` as module constants mirroring the limits their neighbours already use (`routes/documentRoutes.js` reads `MULTIPART_FILE_LIMIT_BYTES` with an 8MB fallback; `lib/malwareScanner.js` reads `SCAN_TIMEOUT_MS` with a 30s fallback), restore the supersede lookup, and capture the superseded storage path before the update.

Also closes #8581.

I'd like to work on this.

advikdivekar:
Closes #9394
Closes #8581

`documentController.js` referenced three identifiers declared nowhere in the module. A merge kept the code that *consumes* the supersede lookup and dropped the code that *produces* it.

### Impact before this PR

The handler wraps everything in a `try/catch` that maps the unexpected to `500 { error: 'An unexpected error occurred' }`, so:

- **Every driver KYC upload failed** — Aadhaar, PAN, licence, RC book. The whole KYC flow was dead.
- The `existingDoc` ReferenceError fires at line 143, **after** `supabase.storage.upload()` at line 128 has already succeeded. The storage-cleanup path only runs on a `dbError`, and a `ReferenceError` is not one — so **every attempt orphaned a file** in the private `driver-documents` bucket with no metadata row pointing at it.
- `MAX_FILE_SIZE_BYTES` throws first at line 46, so even the validation responses were wrong: a bad `documentType` returned 500 instead of 400.

### The existing test suite was already red

`test/integration/documentRoutes.test.js` is committed and **6 of its 9 tests fail on `main`**:

```
 FAIL  returns 400 for an unrecognized documentType          expected 500 to be 400
 FAIL  accepts a real JPEG declared as image/jpeg            expected 500 to be 201
 FAIL  rejects an executable renamed to .jpg with a 422      expected 500 to be 422
 FAIL  rejects a file whose content ≠ declared Content-Type  expected 500 to be 422
 FAIL  returns 500 if storage upload fails (no leakage)
 FAIL  rejects upload when malware scanner detects a threat  expected 500 to be 422

 Tests  6 failed | 3 passed (9)
```

Worth dwelling on the third and sixth: the tests asserting **a renamed executable is rejected** and **a malware-flagged file is rejected** were red. The magic-byte and malware defences were unverified on `main`. Same "suites exist but never run in CI" gap as #8551.

### The changes

**1. The two constants**, mirroring limits their neighbours already use rather than inventing new ones:

```js
const MAX_FILE_SIZE_BYTES =
  Number(process.env.MULTIPART_FILE_LIMIT_BYTES) || 8 * 1024 * 1024;   // matches routes/documentRoutes.js
const SCAN_TIMEOUT_MS = Number(process.env.SCAN_TIMEOUT_MS) || 30000;  // matches lib/malwareScanner.js
```

**2. The supersede lookup.** The surviving code reads `existingDoc.id`, reads `existingDoc.storage_path`, and returns `200` vs `201` off its truthiness — so the intent was unambiguous. It is ordered newest-first and limited to one row rather than assuming uniqueness, because `driver_documents` has **no unique constraint** on `(driver_id, document_type)` (`20260702000000_create_driver_documents.sql`) — which also makes the existing `23505` handler aspirational rather than dead code.

**3. A latent bug in the surviving code.** The cleanup read `existingDoc.storage_path` *after* the update had rewritten that column. Restoring the lookup alone would have made the handler **delete the file it just uploaded and leave the old one orphaned**. The path is now captured before the update:

```js
const supersededStoragePath = existingDoc?.storage_path ?? null;
```

### New tests

Four added to the existing integration suite:

- supersedes same-type re-upload → `200`, one row, old storage object removed
- different document types coexist → two rows
- another driver's same-type document is untouched
- oversized buffer never reaches storage

```
 Test Files  1 passed (1)
      Tests  13 passed (13)      # was: 6 failed | 3 passed
```

`eslint` clean on both files (previously 10 `no-undef` errors).

### One note for reviewers

I dropped an `expect(second.body.document.id).toBe(firstId)` assertion from the supersede test. The reason is a quirk in `test/helpers/supabaseMock.js`: `insert().select().single()` returns a *synthetic* `id` but pushes the payload **without** it into the store, so the stored row has no `id` at all. The test asserts the store-level invariants instead (one row, storage_path rotated, old object gone), which is what actually matters here. Happy to file that mock gap separately if it'd be useful.

### DIFF
--- a/backend/api/src/controllers/documentController.js
+++ b/backend/api/src/controllers/documentController.js
@@ -15,6 +15,17 @@ const ALLOWED_DOCUMENT_TYPES = Object.freeze([
   'other',
 ]);
 
+// Mirrors the multer limit in routes/documentRoutes.js, which reads the same
+// env var and falls back to the same 8MB. multer rejects oversized multipart
+// bodies before the controller runs; this is the guard for the buffer itself,
+// whose length can disagree with the reported size.
+const MAX_FILE_SIZE_BYTES =
+  Number(process.env.MULTIPART_FILE_LIMIT_BYTES) || 8 * 1024 * 1024;
+
+// Mirrors the scanner's own deadline in lib/malwareScanner.js, so the
+// AbortController below cannot fire before or long after the scanner gives up.
+const SCAN_TIMEOUT_MS = Number(process.env.SCAN_TIMEOUT_MS) || 30000;
+
 const MIME_EXTENSION_MAP = Object.freeze({
   'application/pdf': 'pdf',
   'image/png': 'png',
@@ -123,6 +134,33 @@ export async function uploadDriverDocument(req, res) {
       });
     }
 
+    // A driver re-uploading the same document type supersedes the previous
+    // one: the metadata row below is updated in place and the superseded file
+    // removed once the new row is committed. Ordered newest-first and limited
+    // to one row because there is no unique constraint on
+    // (driver_id, document_type) — see the create_driver_documents migration.
+    const { data: existingDoc, error: lookupError } = await supabase
+      .from('driver_documents')
+      .select('id, storage_path')
+      .eq('driver_id', driverId)
+      .eq('document_type', documentType)
+      .order('created_at', { ascending: false })
+      .limit(1)
+      .maybeSingle();
+
+    if (lookupError) {
+      logger.error(
+        '[DocumentController] Failed to look up existing document:',
+        lookupError.message,
+      );
+      return res.status(500).json({ error: 'Failed to store document' });
+    }
+
+    // Captured before the row is updated below: the update rewrites
+    // storage_path, so reading it afterwards would point at the file that was
+    // just uploaded rather than the one being replaced.
+    const supersededStoragePath = existingDoc?.storage_path ?? null;
+
     const storagePath = `${driverId}/${documentType}-${Date.now()}.${extension}`;
 
     const { error: storageError } = await supabase.storage
@@ -191,10 +229,10 @@ export async function uploadDriverDocument(req, res) {
     }
 
     // Clean up old file from storage to prevent orphaned files
-    if (existingDoc?.storage_path) {
+    if (supersededStoragePath) {
       await supabase.storage
         .from('driver-documents')
-        .remove([existingDoc.storage_path])
+        .remove([supersededStoragePath])
         .catch((cleanupErr) => {
           logger.warn('[DocumentController] Failed to delete superseded storage file:', cleanupErr.message);
         });
--- a/backend/api/test/integration/documentRoutes.test.js
+++ b/backend/api/test/integration/documentRoutes.test.js
@@ -152,6 +152,92 @@ describe('Document Routes Integration Tests', () => {
       expect(m.store.__storageObjects.length).toBe(0);
     });
 
+    it('supersedes an existing document of the same type instead of inserting a second row', async () => {
+      const app = buildApp();
+      const upload = () =>
+        request(app)
+          .post('/api/driver/documents')
+          .set(DRIVER_HEADERS)
+          .field('documentType', 'aadhaar_card')
+          .attach('document', JPEG_BYTES, { filename: 'id.jpg', contentType: 'image/jpeg' });
+
+      const first = await upload();
+      expect(first.status).toBe(201);
+      const firstPath = m.store.__storageObjects.at(-1).path;
+
+      const second = await upload();
+
+      // 200, not 201: the metadata row is updated in place rather than a
+      // second row being inserted for the same (driver, document type).
+      expect(second.status).toBe(200);
+      expect(second.body.success).toBe(true);
+      expect(m.store.driver_documents.length).toBe(1);
+      expect(m.store.driver_documents[0].storage_path).not.toBe(firstPath);
+
+      // The superseded file is removed so storage does not accumulate orphans.
+      expect(m.store.__storageObjects.some((o) => o.path === firstPath)).toBe(false);
+    });
+
+    it('keeps documents of different types side by side', async () => {
+      const app = buildApp();
+      const upload = (documentType) =>
+        request(app)
+          .post('/api/driver/documents')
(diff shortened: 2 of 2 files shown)


advikdivekar:
Superseded — the underlying bug this fixed has since regressed via a different bad merge on main (documentController now redeclares `existingDoc`/`checkError` a second time at a different line, referencing an unimported `supabase` in the first block). That regression is already claimed by #10061 and #10669. This branch's diff no longer applies cleanly against current main's very different code shape, so closing rather than re-doing already-claimed work.

PR #9399, opened 2026-08-10
[Test] Test the WebRTC signalling server that exists — 6 failing tests replaced with 61 passing

### LINKED ISSUE #9398: [BUG][backend][P2] WebRTCSignalingServer.test.js tests a room API that does not exist — all 6 tests fail and the real signalling logic is uncovered
### Summary

`backend/api/test/unit/WebRTCSignalingServer.test.js` exercises `createRoom` / `getRoom` / `joinRoom` / `leaveRoom` / `routeSignalingMessage` as **static** methods. `src/services/webrtc/WebRTCSignalingServer.js` has none of them. It is not a room server at all — it is a peer *mesh*: peers get a generated `peer_<uuid>` id and are grouped into `mesh_<uuid>` sets.

```
$ cd backend/api && ../../node_modules/.bin/vitest run --silent=true --reporter=dot test/unit/WebRTCSignalingServer.test.js

 FAIL  createRoom > creates a new room with a unique id
   TypeError: WebRTCSignalingServer.createRoom is not a function
 FAIL  createRoom > returns existing room if room already exists
 FAIL  joinRoom > adds participant to room
 FAIL  joinRoom > returns false when room not found
   TypeError: WebRTCSignalingServer.joinRoom is not a function
 FAIL  leaveRoom > removes participant from room
 FAIL  routeSignalingMessage > routes message to correct participant

 Tests  6 failed (6)
```

The default export is the **class**, not an instance, so even a correctly-named method would not be reachable the way the suite calls it.

### What this leaves uncovered

The module has a substantial pure-logic surface and a security check, none of it tested:

- `isValidLocation()` — the coordinate guard on every inbound `location-update` and GPS frame. It accepts numeric strings (payloads arrive as JSON) but must reject `NaN`, `Infinity`, out-of-range values and `{}` — `Number(undefined)` is `NaN`, and that is the only thing stopping a coordinate-less payload from being read as null island.
- `calculateDistance()` — haversine, feeding `getPeersNearLocation()`. Antipodal points are the classic NaN case.
- `sendPeerList()` — must exclude the recipient and must not leak peers from another mesh.
- `handleDisconnect()` — must drop the mesh when its last peer leaves, or meshes accumulate forever.
- `canUserAccessPeer()` — the ownership check gating `getOfflineGPSData()` and `syncOfflineData()`, i.e. **who can read another peer's GPS history**.
- `destroy()` — must keep closing the remaining sockets when one `close()` throws.

### Also: the suite mocks `config/db.js` in a way that cannot work

```js
vi.mock('../../src/config/db.js', () => ({}));
```

The module destructures `{ supabase, redisClient }` from it, so vitest raises *"No `redisClient` export is defined on the `../../src/config/db.js` mock"*. The same broken-mock pattern is behind several other red suites in this repo.

### Proposal

Replace the suite with one that tests the real API. Constructing the server binds a `WebSocketServer` to an http server and starts a 30s discovery interval, so the tests should build a bare instance off the prototype and populate the two Maps directly — that keeps the pure logic reachable without sockets or timers.

Writing this surfaced a genuine defect in `canUserAccessPeer()`, which I've filed separately rather than fold in here.

I'd like to work on this.

advikdivekar:
Closes #9398

`test/unit/WebRTCSignalingServer.test.js` called `createRoom` / `getRoom` / `joinRoom` / `leaveRoom` / `routeSignalingMessage` as **static** methods. None of them exist. The module is a peer *mesh* — peers get a generated `peer_<uuid>` id and are grouped into `mesh_<uuid>` sets — and the default export is the class, not an instance, so even a correctly-named method would not be reachable that way.

```
 Before:  Tests  6 failed (6)     TypeError: WebRTCSignalingServer.createRoom is not a function
 After:   Tests  61 passed (61)
```

Its db mock could not work either: `vi.mock('../../src/config/db.js', () => ({}))` against a module that destructures `{ supabase, redisClient }`, which vitest rejects with *"No `redisClient` export is defined"*. Same broken-mock shape as several other red suites in this repo.

### Approach

Constructing the server binds a `WebSocketServer` to an http server and starts a 30s discovery interval. The tests build a bare instance off the prototype and populate the two Maps directly, so the pure logic is reachable with no sockets and no timers.

### What is now covered

| Area | Notable cases |
|---|---|
| `isValidLocation()` | numeric strings accepted (payloads arrive as JSON); `NaN`, `Infinity`, out-of-range rejected; **`{}` rejected rather than coerced to null island** — `Number(undefined)` is `NaN`, which is exactly what the `Number.isFinite` guard is for; poles and antimeridian accepted as in range |
| `calculateDistance()` | symmetry, a known ~290km separation, and **antipodal points** — the classic haversine `NaN` |
| `sendPeerList()` | excludes the recipient; **never leaks peers from another mesh**; sends an empty list rather than nothing; skips mesh members whose peer record is already gone |
| `handleDisconnect()` | **drops the mesh with its last peer**, so meshes cannot accumulate forever |
| `getPeersNearLocation()` | radius filtering, the 10km default, peers with no location |
| `canUserAccessPeer()` + `getOfflineGPSData()` + `syncOfflineData()` | the ownership check gating **who can read another peer's GPS history** — including that an unauthorized caller never reaches supabase at all |
| `destroy()` | keeps closing the remaining sockets when one `close()` throws |
| `sendToPeer()` | no-op for unknown peers and for sockets that are not open |

### One thing writing this surfaced

Three `canUserAccessPeer()` cases pin behaviour that is **not** correct: `peer.userId === user?.id` is true when both sides are `undefined`, so a peer registered from a token with no `id` claim is reachable by a caller who also has no id. Those three tests assert the current `true` with a comment saying so, rather than going red.

I've filed that separately and will follow up with the one-line hardening, so it lands as a visible diff against a green suite instead of being buried in a 450-line test rewrite. Reviewers who'd rather see it folded in here — say so and I'll squash them.

Test-only. `eslint` clean.

### DIFF
--- a/backend/api/test/unit/WebRTCSignalingServer.test.js
+++ b/backend/api/test/unit/WebRTCSignalingServer.test.js
@@ -1,70 +1,472 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 
-vi.mock('../../src/config/db.js', () => ({}));
+/**
+ * The signalling server is a peer *mesh*, not a room server: peers are keyed
+ * by a generated peerId and grouped into meshes. Constructing it binds a
+ * WebSocketServer to an http server and starts a 30s discovery interval, so
+ * these tests build a bare instance off the prototype and populate the two
+ * Maps directly. That keeps the pure logic — location validation, distance,
+ * peer fan-out, authorization — testable without any sockets or timers.
+ */
+
+const supabaseQuery = {
+  select: vi.fn().mockReturnThis(),
+  insert: vi.fn().mockResolvedValue({ error: null }),
+  update: vi.fn().mockReturnThis(),
+  eq: vi.fn().mockReturnThis(),
+  gt: vi.fn().mockReturnThis(),
+  order: vi.fn().mockResolvedValue({ data: [] }),
+};
+
+const supabaseMock = { from: vi.fn(() => supabaseQuery) };
+const redisMock = { setex: vi.fn().mockResolvedValue('OK') };
+
+vi.mock('../../src/config/db.js', () => ({
+  supabase: supabaseMock,
+  redisClient: redisMock,
+  firebaseAdmin: null,
+  mongoDb: null,
+}));
+
+vi.mock('ws', () => ({
+  WebSocketServer: class {
+    on() {}
+    close() {}
+  },
+}));
+
+const { default: WebRTCSignalingServer } = await import(
+  '../../src/services/webrtc/WebRTCSignalingServer.js'
+);
+
+const OPEN = 1;
+
+/** A signalling server with no sockets, no timers and no discovery loop. */
+function bareServer() {
+  const server = Object.create(WebRTCSignalingServer.prototype);
+  server.peers = new Map();
+  server.meshes = new Map();
+  server.redis = redisMock;
+  server.wss = { close: vi.fn() };
+  return server;
+}
+
+/** Registers a peer in `meshId`, creating the mesh if needed. */
+function addPeer(server, peerId, { meshId = 'mesh-1', readyState = OPEN, ...rest } = {}) {
+  const ws = { readyState, send: vi.fn(), close: vi.fn() };
+  server.peers.set(peerId, { ws, meshId, connectedAt: 1700000000000, ...rest });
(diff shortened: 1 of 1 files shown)


KanishJebaMathewM:
level:advanced

PR #10270, opened 2026-08-11
test: rewrite notificationService tests against real module API

### LINKED ISSUE #10242: bug: notificationService.test.js cannot load — imports nonexistent default export and src/errors/DomainError.js
## Bug
`backend/api/test/unit/notificationService.test.js` cannot load — the test suite fails before a single test runs.

## Root cause
The test imports three things that do not exist in the module:

1. `import notificationService from '../../src/services/notificationService.js';` — `notificationService.js` has **no default export** (it only exports named functions like `sendFcmNotification`, `hashDeliveryOtp`, `verifyDeliveryOtpHash`).
2. `import DomainError from '../../src/errors/DomainError.js';` — the `src/errors/` directory does not exist; `DomainError` lives at `src/services/order/domainError.js`.
3. It then calls `notificationService.insertNotification(...)` and `notificationService.sendPushNotification(...)`, neither of which exists in the module (the real API is `sendFcmNotification`).

## Impact
Verified failure — `npx vitest run test/unit/notificationService.test.js` reports `Test Files 1 failed (1)`, `Tests no tests`, with the import error pointing at the `DomainError` import line. The suite is dead weight and masks that the described "allowlist validation" behavior does not exist in `notificationService.js`.

## Suggested fix
Rewrite the test to target the real module surface:
- Import the actual named exports (`sendFcmNotification`, `hashDeliveryOtp`, `verifyDeliveryOtpHash`).
- Import `DomainError` from `src/services/order/domainError.js` if needed.
- Replace the "allowlist validation" cases with assertions on real behavior (e.g. `sendFcmNotification` returns `{ success: false }` instead of throwing when Firebase is not configured, and the OTP hash helpers round-trip correctly).

Akshita-2307:
﻿Fixes #10242

## Status
The suite-loading half of this issue was already fixed by the merged PR #10094 (`d6e1d47f0`), which rewrote the tests to use the real named exports and removed the nonexistent default export / `DomainError` import. This PR was rebased on current `main` and now supplies the remaining piece of the issue's suggested fix: coverage for the OTP hashing helpers, which still had zero tests.

## Changes
- `backend/api/test/unit/notificationService.test.js`: added a `delivery OTP hashing` describe block on top of the existing (merged) suite:
  - `hashDeliveryOtp` / `verifyDeliveryOtpHash` round-trip (salted scrypt); wrong OTP is rejected.
  - Legacy unsalted SHA-256 hashes still verify; wrong OTP is rejected.

## Verification
- `npx vitest run test/unit/notificationService.test.js` ? 20/20 passed (18 existing + 2 new).

### DIFF
--- a/backend/api/test/unit/notificationService.test.js
+++ b/backend/api/test/unit/notificationService.test.js
@@ -1,5 +1,6 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
-import { sendPushNotification, sendFcmNotification } from '../../src/services/notificationService.js';
+import crypto from 'crypto';
+import { sendPushNotification, sendFcmNotification, hashDeliveryOtp, verifyDeliveryOtpHash } from '../../src/services/notificationService.js';
 
 // notificationService reads supabaseAdmin and firebaseAdmin (not supabase),
 // and getUserFcmToken chains .select().eq().maybeSingle() — a mock lacking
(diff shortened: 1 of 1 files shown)


Akshita-2307:
﻿Rebased onto current `main` and resolved the conflict. The suite-loading half of #10242 was already fixed by the merged #10094, so this PR now carries only the remaining piece from the issue's suggested fix: `hashDeliveryOtp`/`verifyDeliveryOtpHash` coverage (salted scrypt round-trip + legacy unsalted SHA-256), which had zero tests on `main`.

- `npx vitest run test/unit/notificationService.test.js` ? 20/20 passed locally.
- Diff is now just `backend/api/test/unit/notificationService.test.js` (+17/?1).
- CI failures shown (Backend ESLint/Tests shards, Flutter, ML, SonarCloud, etc.) are pre-existing on current `main` and reproduce on unrelated merged PRs (e.g. #10763 had the same 12 failing checks) ? none touch this change.

Akshita-2307:
GSSOC

PR #8133, opened 2026-08-08
fix(security): prevent IDOR in live tracking endpoints

KrutagyaKaneria:
## Summary

This PR fixes a Broken Object-Level Authorization (BOLA/IDOR) vulnerability in the live-tracking order endpoints.

Affected endpoints:

- GET /api/orders/:id/driver-location
- GET /api/orders/:id/route

The affected routes were using ownership-based authorization policies without properly binding the requested order as the authorization resource.

As a result, ownership validation could be skipped, potentially allowing an authenticated user to access another user's live driver location and route geometry.

## Problem

The affected routes were invoking:

requirePolicy('order:view-driver-location')

and:

requirePolicy('order:view-route')

without providing the required order resource.

This caused the authorization layer to receive an undefined resource.

The affected policies depend on the order's:

- customer_id
- driver_id

When the resource was unavailable, the ownership check could be skipped.

This created a potential IDOR/BOLA vulnerability.

## Security Impact

An authenticated user could potentially access another order's:

- Driver latitude
- Driver longitude
- Driver ID
- Location timestamp
- Pickup coordinates
- Drop-off coordinates
- Route geometry

simply by knowing or obtaining an order identifier.

This could expose sensitive real-time logistics and location information.

### Resource-Bound Authorization

Updated the affected live-tracking routes so the authorization policy receives the authoritative order resource.

The order resource is resolved from the database rather than from client-provided ownership fields.

The existing authorization rules are preserved:

- Order customer → Allowed
- Assigned driver → Allowed
- Administrator → Allowed
- Unrelated authenticated user → Denied

### Fail-Closed Authorization

Updated the authorization flow so ownership-based policies cannot silently pass when no resource is supplied.

The authorization behavior is now:

Ownership policy + valid resource
→ Evaluate ownership

Ownership policy + missing resource
→ Deny access

Role-only policies continue to follow their existing behavior.

### Authorization Audit

Reviewed the relevant requirePolicy() usage and ensured the affected authorization decisions are covered by the existing authorization/audit mechanism.

### Security Regression Tests

Added and updated tests covering:

- Unauthenticated users
- Order customers
- Assigned drivers
- Administrators
- Unrelated authenticated users
- Missing authorization resources
- Role-only policies
- Invalid/nonexistent orders
- Driver-location endpoint
- Route endpoint

## Authorization Matrix

| User Type | Driver Location | Route Geometry |
|-----------|------------------|----------------|
| Unauthenticated User | Denied | Denied |
[...]

PR #10274, opened 2026-08-11
fix: bound getActiveOrders read with pagination

### LINKED ISSUE #10244: bug: GET /api/orders/my/active reads active orders with no limit — silently truncated at 1000 rows
## Bug
`GET /api/orders/my/active` performs an unbounded read of the `orders` table — active orders are silently truncated at PostgREST's 1000-row response cap.

## Root cause
`orderRoutes.js:1127` calls `orderLifecycleService.getActiveOrders(req.user.id)`, which calls `findOrdersByCustomer(customerId, '*', activeStatuses, 'pickup_date', false)` at `orderLifecycleService.js:198` — the 6th `pagination` argument is omitted.

`findOrdersByCustomer` in `orderRepository.js:88` only applies `.range(from, to)` when `pagination` is provided:

```js
if (pagination) {
  const { from, to } = buildPagination(pagination);
  query = query.range(from, to);
}
```

With no pagination the query is `select('*').eq('customer_id', ...).in('status', ...)` with no `.limit()`/`.range()`.

## Impact
PostgREST caps a single response at 1000 rows. A customer with more than 1000 active orders silently loses the rest — the endpoint returns an incomplete "active orders" list with no indication of truncation, so the app thinks the customer has no more active orders.

## Suggested fix
Pass a bounded pagination window from `getActiveOrders` (e.g. a generous `{ limit: 1000 }`/explicit `range`) so the read is bounded and cannot silently truncate.

Akshita-2307:
﻿Fixes #10244

## Status
Rebased onto current `main` and resolved the merge conflict (upstream #10202 switched `getActiveOrders` to an explicit column allow-list, which is what clashed with this change).

## Change
- `backend/api/src/services/order/orderLifecycleService.js`: `getActiveOrders` now passes `{ limit: 100 }` (the app-wide max, enforced by `buildPagination`) to `findOrdersByCustomer`, so the read no longer relies on PostgREST's 1000-row server cap. Diff vs `main` is exactly +1/?1.

## Verification
- The changed line parses (vitest transform gets past it; the only parse error in the file is the pre-existing duplicate `newAmountWei` on `main`, introduced upstream by #8133 ? it also breaks `node --check`/ESLint for every PR right now and is unrelated to this change).
- `orderLifecycleService.test.js` 4 failures are pre-existing import failures from that same duplicate declaration.

### DIFF
--- a/backend/api/src/services/order/orderLifecycleService.js
+++ b/backend/api/src/services/order/orderLifecycleService.js
@@ -195,7 +195,7 @@ export class OrderLifecycleService {
       const { data: orders, error } = await this.orderRepository.findOrdersByCustomer(
         customerId,
         'id, order_display_id, status, pickup_address, drop_address, pickup_date, pickup_lat, pickup_lng, drop_lat, drop_lng, eta, driver_id, truck_id, truck_number, total_amount, goods_type, weight_tonnes, length_ft, width_ft, height_ft, is_stackable, is_fragile, special_requirements, created_at, updated_at',
-        activeStatuses, 'pickup_date', false
+        activeStatuses, 'pickup_date', false, { limit: 100 }
       );
 
       if (error) throw new DomainError(500, { error: 'Failed to fetch active orders.', details: error.message });


Akshita-2307:
﻿Rebased onto current `main` and resolved the conflict. Main's #10202 changed `getActiveOrders` to an explicit column allow-list; this change now lands on top of that: `findOrdersByCustomer(..., { limit: 100 })`.

Heads-up (pre-existing, not from this PR): current `main` has a duplicate `const newAmountWei` in `orderLifecycleService.js` (lines 580?581, from upstream #8133). It makes the whole module fail to parse (`node --check` and vitest both reject it), which is why `orderLifecycleService.test.js` shows 4 failures and the repo-wide ESLint/Tests shards are red for every open PR. Worth a separate fix commit on `main` so modules importing this service can load.

Akshita-2307:
GSSOC

PR #10279, opened 2026-08-11
fix: use authenticated client for offline GPS reads and sync

### LINKED ISSUE #10245: bug: WebRTC signaling writes/reads gps_offline_data via the anon client — offline GPS sync is dead
## Bug
The WebRTC offline GPS sync path is dead: `WebRTCSignalingServer` writes and reads `gps_offline_data` through the **anon** `supabase` client, so every insert/query fails under RLS and silently no-ops.

## Root cause
`backend/api/src/services/webrtc/WebRTCSignalingServer.js:5`:

```js
import { supabase, redisClient } from '../../config/db.js';
```

`supabase` is the anon (public) client. It is then used for persistence:

- `handleGPSData` — `supabase.from('gps_offline_data').insert([gpsEntry])` (line 242)
- `getOfflineGPSData` — `supabase.from('gps_offline_data').select('*')...` (line 412)
- `syncOfflineData` — `supabase.from('gps_offline_data').update({ synced: true })...` (line 429)

The `gps_offline_data` table is RLS-protected with no anon privileges (same posture as every other table in this codebase), so these anon-client calls return an RLS error. The insert error is logged as a warning and swallowed (lines 243–248), so the failure is invisible.

## Impact
- Offline GPS samples are never persisted — the driver app's offline-sync feature (and `GET /api/webrtc/offline/:peerId` + `POST /api/webrtc/sync/:peerId`) always return empty/no-op data even though the server accepted the GPS payloads.
- Confirmed pattern matches the previously-fixed anon-client bugs in this repo (e.g. `POST /api/support/tickets`, `POST /api/trucks`).

## Suggested fix
Use the service-role client (`supabaseAdmin`) for the three `gps_offline_data` operations in `WebRTCSignalingServer.js`.

Akshita-2307:
Fixes #10245

## Problem
\getOfflineGPSData\ and \syncOfflineData\ in \WebRTCSignalingServer.js\ read and update \gps_offline_data\ via the anonymous \supabase\ client. Under RLS, the anon role has no access to this table, so offline GPS data can never be retrieved or marked synced — only the insert path (line ~247) had been migrated to an authenticated client via \createUserClient(peer.token)\.

## Fix
Mirrored the existing authenticated-client pattern from the insert path: both methods now build \createUserClient(peer.token)\ from the peer in the in-memory map (falling back to \supabase\ exactly like the insert path) before hitting \gps_offline_data\.

## Verification
- \
ode --check backend/api/src/services/webrtc/WebRTCSignalingServer.js\ passes.
- \
px vitest run test/unit/WebRTCSignalingServer.test.js\ → 61/61 passed.

## GSSOC'24

### DIFF
--- a/backend/api/src/services/webrtc/WebRTCSignalingServer.js
+++ b/backend/api/src/services/webrtc/WebRTCSignalingServer.js
@@ -414,7 +414,11 @@ class WebRTCSignalingServer {
       logger.warn(`[WebRTC] Unauthorized offline GPS data access attempt for peer ${peerId}`);
       return [];
     }
-    const { data } = await supabase
+    const peer = this.peers.get(peerId);
+    // Use an authenticated Supabase client for GPS data reads (RLS requires authenticated role)
+    const userClient = peer?.token ? createUserClient(peer.token) : null;
+    const gpsClient = userClient || supabase;
+    const { data } = await gpsClient
       .from('gps_offline_data')
       .select('*')
       .eq('peerId', peerId)
@@ -430,7 +434,11 @@ class WebRTCSignalingServer {
       return;
     }
     // Mark data as synced for this peer
-    await supabase
+    const peer = this.peers.get(peerId);
+    // Use an authenticated Supabase client for GPS data updates (RLS requires authenticated role)
+    const userClient = peer?.token ? createUserClient(peer.token) : null;
+    const gpsClient = userClient || supabase;
+    await gpsClient
       .from('gps_offline_data')
       .update({ synced: true })
       .eq('peerId', peerId)


Akshita-2307:
﻿Superseded ? this is already fixed on current `main`. Upstream commit `b58fd6cde` (#9935, "persist WebRTC offline GPS via authenticated client keyed by owner profile") migrated both `getOfflineGPSData` and `syncOfflineData` off the anon `supabase` client:

```js
const client = requestingUser.role === 'admin' ? supabaseAdmin : (token ? createUserClient(token) : supabase);
```

It also covers an admin path (`supabaseAdmin`) and keying rows by the owner profile id. `WebRTCSignalingServer.test.js` passes 64/64 on current `main`. No conflict-resolution changes are needed, so closing this PR.

Akshita-2307:
Closing as superseded by #9935 (see comment above).

PR #10525, opened 2026-08-11
fix(n8n): align document-integrity Check Documents with driver_documents

### LINKED ISSUE #10157: `[BUG][n8n][P2] document-integrity Check Documents node reads doc.type / status 'verified' / expiry_date, but driver_documents has document_type, status IN (pending_review, approved, rejected) and no expiry_date - every driver is flagged missing or unverified`
## `[BUG][n8n][P2] document-integrity Check Documents node reads doc.type / status 'verified' / expiry_date, but driver_documents has document_type, status IN (pending_review, approved, rejected) and no expiry_date - every driver is flagged missing or unverified`

**Subsystem:** `automation/n8n/document-integrity-workflow.json` + `supabase/migrations/20260702000000_create_driver_documents.sql`

**Summary**
The workflow's "Check Documents" code node expects each driver document object to look like `{ type: 'RC'|'License'|'Insurance', status: 'verified'|..., expiry_date: 'YYYY-MM-DD' }`. The `driver_documents` table actually stores `document_type` (text), `status` constrained to `('pending_review','approved','rejected')`, `storage_path`, `mime_type`, `created_at`, `updated_at` — there is **no** `type`, **no** `'verified'` status, and **no** `expiry_date` column anywhere. Consequently the `docs.find(d => d.type === docType)` lookup always returns `undefined` (or an object without the expected fields), the node always marks RC/License/Insurance as `missing`, and the integrity-alert branch fires for every driver every day even when their documents are approved on-chain.

**Evidence**
- `automation/n8n/document-integrity-workflow.json` — "Check Documents" node (n8n-nodes-base.code):
  ```js
  const requiredDocs = ['RC', 'License', 'Insurance'];
  const doc = docs.find(d => d.type === docType);
  if (!doc) missingOrTampered.push({ type: docType, reason: 'missing' });
  else if (doc.status !== 'verified' || (doc.expiry_date && new Date(doc.expiry_date) < new Date())) { ... }
  ```
- `supabase/migrations/20260702000000_create_driver_documents.sql`:
  ```sql
  CREATE TABLE driver_documents (
    id uuid PRIMARY KEY,
    driver_id uuid NOT NULL REFERENCES profiles(id) ON DELETE CASCADE,
    document_type text NOT NULL,
    storage_path text NOT NULL,
    mime_type text NOT NULL,
    status text NOT NULL DEFAULT 'pending_review'
      CHECK (status IN ('pending_review', 'approved', 'rejected')),
    created_at timestamptz NOT NULL DEFAULT now(),
    updated_at timestamptz NOT NULL DEFAULT now()
  );
  ```
- The workflow's upstream HTTP call (`GET /api/drivers/active`, tracked separately in #9651) returns drivers; the node then reads `driver.documents` from the response, but nothing in the API or DB produces the expected `{type, status, expiry_date}` shape.

**Repro**
Import `document-integrity-workflow.json` into n8n and run "Check Documents" against any driver with an approved `driver_documents` row → output `has_issues: true` with `issues: [{type:'RC', reason:'missing'}, ...]`.

**Notes / product implication**
- The workflow is a daily KYC-integrity sweep that will alert on every driver forever, and it can never report a clean pass.
- Distinct from #9469 (documentExpiryService scans the legacy `documents` table) and #8708/#8757 (digilockerService columns): this is the n8n workflow node's own schema contract.

**Suggested fix**
Update the node to map `document_type` and `status` (treating `approved` as verified) and drop the `expiry_date` check until such a column exists (add it in a migration if expiry reminders are wanted).

ionfwsrijan:
## Problem
The "Check Documents" code node in `document-integrity-workflow.json` does not match the `driver_documents` schema:

- `requiredDocs` included `'insurance'`, which is not a valid `document_type` (the app only allows `aadhaar_card`, `pan_card`, `driving_licence`, `rc_book`, `other`), so every driver was flagged as missing insurance forever.
- The node read `doc.type`, but the DB stores `document_type` (the `/api/driver/active` endpoint maps it to `type` in its response, so it works today but is fragile and reads the wrong source per the DB contract).
- The `expiry_date` check referenced a column that does not exist in `driver_documents` (and never will until a migration adds it).

Net effect: the daily KYC-integrity sweep alerts on every driver every day and can never report a clean pass.

## Fix
- `requiredDocs` → `['rc_book', 'driving_licence']` (the two required types that actually exist; dropped the non-existent `insurance`).
- Document lookup matches `d.document_type || d.type`, so the node is correct against both the DB shape and the API-normalized shape.
- Status now treats `approved` as verified; anything else (`pending_review`, `rejected`) is reported as `unverified`.
- Removed the `expiry_date` branch (and the now-impossible `'expired'` reason) until such a column is added by a future migration.
- Coherent labels: renamed the IF node to "Has Mismatch?" and the alert subject to "Document Integrity Alert" (no more expiry semantics).

## Files changed
- `automation/n8n/document-integrity-workflow.json`

## Testing
Extracted the node's `jsCode` and ran it through the same scenarios it sees in production:
- `document_type` shape: `rc_book`+`driving_licence` approved → `has_issues: false`.
- API `type` shape: `rc_book`+`driving_licence` approved → `has_issues: false`.
- `rc_book` `pending_review` → `issues: [{type:'rc_book', reason:'unverified'}]`.
- no docs → both required types reported `missing`.
- an extraneous `insurance` doc no longer triggers a false flag.
- `node --check` passes on the extracted code; workflow JSON parses; `scripts/check-n8n-workflow-urls.js` still reports all 17 workflow URLs map to mounted routes.

Closes #10157


KanishJebaMathewM:
level:advanced

PR #10549, opened 2026-08-11
fix: prevent false positive SQLi request blocking

### LINKED ISSUE #10508: fix : remove false-positive hard-blocking of SQLi pattern in suspiciousRequests
The suspiciousRequests middleware (backend/api/src/middleware/suspiciousRequests.js) runs the SQLi pattern /(?:^[\s'\";])--/ against JSON.stringify(req.body) and JSON.stringify(req.query). After JSON-stringification, any string value containing space followed by '--' matches, causing false-positive 403s for completely legitimate requests like addresses containing '--'. The fix is to either restrict the pattern to the raw URL, make SQLi warn-only like XSS, or require adjacent SQL keywords.

Changes that Need to be Made:
- backend/api/src/middleware/suspiciousRequests.js: Either remove the hard-blocking on the ' -- ' pattern, restrict it to raw URL matching, or require more SQL context around the pattern.

Impact that it would Provide:
- Eliminates false-positive 403s for legitimate traffic containing '--'
- More accurate SQLi detection that targets actual injection contexts
- Prevents legitimate addresses, notes, and search queries from being blocked

CoderMS07:
## Description

Updated suspicious request handling to prevent the SQLi `--` pattern from blocking legitimate requests.

## Changes Made

- Kept SQLi detection and logging for suspicious requests.
- Restricted HTTP 403 blocking to Path Traversal findings.
- Prevented legitimate request data containing `--` from being rejected.
- Preserved existing suspicious request logging behavior.

## Impact

- Legitimate addresses, notes, and search queries containing `--` are no longer blocked.
- SQLi findings remain visible through request logging.
- Path Traversal requests continue to be blocked.

No unrelated behavior was changed.

Closes https://github.com/KanishJebaMathewM/Truxify/issues/10508

### DIFF
--- a/backend/api/src/middleware/suspiciousRequests.js
+++ b/backend/api/src/middleware/suspiciousRequests.js
@@ -71,7 +71,7 @@ export default function suspiciousRequests(req, res, next) {
     }, "Suspicious request detected");
 
     const blocking = findings.filter(f =>
-      ['SQL Injection', 'Path Traversal'].includes(f)
+      ['Path Traversal'].includes(f)
     );
     if (blocking.length) {
       return res.status(403).json({ error: 'Request blocked: suspicious content detected' });


CoderMS07:
GSSOC

PR #10568, opened 2026-08-11
[Test] services/voiceService.test.js: mock supabaseAdmin so the suite loads

### LINKED ISSUE #10567: [BUG][backend][test] services/voiceService.test.js — config/db.js mock omits supabaseAdmin
## Description

`voiceService.js` resolves its DB client at module scope:
```js
import { supabase, supabaseAdmin } from '../config/db.js';
const voiceDb = supabaseAdmin || supabase;
```

`test/unit/services/voiceService.test.js` mocks `config/db.js` with only `supabase` — no `supabaseAdmin`. Destructuring an export vitest's mock doesn't define throws at module-load time:

```
Error: [vitest] No "supabaseAdmin" export is defined on the "../../../src/config/db.js" mock.
```

Since this happens at the top-level `const voiceDb = ...` during module import, the whole suite fails to load — 0 tests run.

## Proposed Fix

Add `supabaseAdmin` to the mock, pointing at the same `mockSupabase` double as `supabase`.

13/13 passing after the fix.

advikdivekar:
Closes #10567

`voiceService.js` resolves its db client at module scope: `const voiceDb = supabaseAdmin || supabase`. The test's `config/db.js` mock only exported `supabase`, so destructuring `supabaseAdmin` threw at import time — the whole suite failed to load (0 tests ran) rather than any individual test failing.

Added `supabaseAdmin` to the mock, pointing at the same double as `supabase`. 13/13 passing.

### DIFF
--- a/backend/api/test/unit/services/voiceService.test.js
+++ b/backend/api/test/unit/services/voiceService.test.js
@@ -22,6 +22,7 @@ const mockSupabase = {
 
 vi.mock('../../../src/config/db.js', () => ({
   supabase: mockSupabase,
+  supabaseAdmin: mockSupabase,
 }));
 
 vi.mock('../../../src/middleware/logger.js', () => ({


advikdivekar:
Superseded — main already has a more complete fix for this (adds `createUserClient` mock support too, not just `supabaseAdmin`). Rebasing this branch onto main resolves to zero diff; the test passes as-is on main (15/15). Closing rather than merging a no-op.

PR #10728, opened 2026-08-12
fix: cap matrix inputs, stream rows, guard send length

### LINKED ISSUE #10638: [Critical] route-matrix-cpp /matrix builds unbounded NxN response (OOM) and truncates length to int in send
## Description

`services/route-matrix-cpp/main.cpp` — the `/matrix` handler materializes a quadratic N×N response and truncates its length to `int`:

```cpp
// main.cpp:63-100 — compute_matrix_json
std::string compute_matrix_json(const std::vector<Location>& locs) {
    ...
    for (size_t i = 0; i < locs.size(); ++i) {
        for (size_t j = 0; j < locs.size(); ++j) {
            ...
            ss << "    {\n";
            ss << "      \"origin\": \"" << locs[i].id << "\",\n";
            ...
```

```cpp
// main.cpp:243
send(client, response.c_str(), static_cast<int>(response.size()), 0);
```

`parse_locations` (lines 135-156) accepts every `{...}` object in the request body with no count limit, and the request body is only capped at 64 KiB (line 223) — enough for ~2,000+ locations. `compute_matrix_json` builds ~4M cells into a single in-memory `std::stringstream` (several hundred MB). `response.size()` is then narrowed to `int`; beyond 2 GB the length becomes negative and the send is corrupted/fails.

## Expected Behavior

Oversized inputs are rejected with `400`/`413`; responses are produced within memory bounds; response lengths are transmitted correctly regardless of size.

## Actual Behavior

A handful of `/matrix` requests exhaust container memory (remote OOM), and for very large matrices the `int` truncation corrupts the HTTP response. Trivially repeatable. (The same N×N + `int` pattern exists in `vector-matcher-cpp/main.cpp`, bounded there by a fixed 1000-row pool, so lower severity.)

## Proposed Fix

- Enforce a hard maximum on the number of parsed locations (and a matrix-cell budget); reject oversized inputs with `400`/`413` before building anything.
- Stream the matrix row-by-row (chunked responses or per-row flush) instead of building one giant buffer.
- Check `response.size()` against `INT_MAX` before the `int` cast in `send`.
- Add a test that posts a location list above the cap and asserts a clean `413` (no crash, no partial response).

ionfwsrijan:
## Problem

`/matrix` materializes a quadratic N×N response into one in-memory `std::stringstream` with no input cap (`parse_locations` accepts every `{...}` object; the body cap of 64 KiB still admits ~2,000+ locations) and then narrows `response.size()` to `int` in `send`, corrupting/failing responses past 2 GB — a remote OOM and response-corruption DoS.

## Fix

- Added `include/matrix_http.hpp` holding the pure logic (caps, parsing, decision, matrix JSON) so it is unit-testable without sockets:
  - `MAX_LOCATIONS = 1000` and `MAX_MATRIX_CELLS = 1,000,000` (N² budget).
  - `parse_locations` stops at the cap and sets `limit_hit` when the body contains more objects.
  - `matrix_cell_count(n)` guards `size_t` overflow.
  - `decide_matrix_request` returns a single clean `413 Payload Too Large` (or `400` when empty) before anything is built.
- `main.cpp`:
  - `/matrix` now decides via `decide_matrix_request`; over-cap/oversized requests get one clean `413` response — no crash, no partial matrix response.
  - Valid requests are streamed row-by-row over `Transfer-Encoding: chunked` (`stream_matrix_response` + `send_chunk`) instead of building one giant buffer.
  - `send_all` loops until the whole buffer is sent and caps each `send()` length at `INT_MAX`, eliminating the `int` truncation.

## Files changed

- `services/route-matrix-cpp/include/matrix_http.hpp` (new)
- `services/route-matrix-cpp/main.cpp`
- `services/route-matrix-cpp/tests/matrix_tests.cpp` (new)
- `services/route-matrix-cpp/CMakeLists.txt` (include path + `matrix_tests` ctest target)

## Testing

Added `matrix_tests` (wired into ctest) covering:
- normal parse, empty body;
- a body above the cap → `limit_hit`, never fully parsed;
- `matrix_cell_count` including the overflow guard;
- decision logic: over-cap → clean `413 Payload Too Large`, empty → `400`, valid → accepted;
- `compute_matrix_json` emits a well-formed 2×2 response (4 cells, `success: true`).

Note: no C++ toolchain is available locally, so the tests could not be executed here.

Closes #10638


KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

PR #10758, opened 2026-08-12
fix(supabase): prevent complete_trip_tx from completing when driver_details row is missing

### LINKED ISSUE #10678: [BUG][supabase][P1] complete_trip_tx marks order payment_released without crediting driver wallet when driver_details row is missing — idempotency guard makes it unrecoverable
## Summary
`complete_trip_tx` (`supabase/migrations/20260810120000_complete_trip_tx_write_net_earnings.sql:195-201`) credits the driver wallet with an unconditional `UPDATE driver_details SET wallet_confirmed = wallet_confirmed + ... WHERE user_id = v_order.driver_id;`. There is no row-count check: if the driver's `driver_details` row is absent, the update silently affects 0 rows while the rest of the RPC proceeds:
- order → `payment_released`, escrow → `released`
- a `wallet_transactions` row is inserted as `credit`/`confirmed`
- `earnings_daily` is upserted

The driver's wallet is never credited, yet the trip is recorded as paid out.

## The missing row is a real state, not hypothetical
This codebase already treats `driver_details` as possibly missing:
- `withdraw_funds_tx` has a `v_confirmed IS NULL` guard (added for #2932).
- `accept_bid_tx` uses `LEFT JOIN driver_details` precisely because the row may not exist.
So a driver whose profile was created without a `driver_details` row (or whose row was pruned) will complete trips normally while the wallet credit is silently dropped.

## Why unrecoverable
The RPC's idempotency guard (`status = 'payment_released'` early return, lines 71-75) means any retry does nothing. No reconciliation path re-credits the wallet, so `driver_details.wallet_confirmed`, `wallet_transactions` (shows a confirmed credit that never landed), `earnings_daily` and `trips.net_earnings` diverge permanently.

## Expected vs actual
Expected: if the driver wallet cannot be credited, the RPC must fail (or insert the row) and must NOT mark the order `payment_released`.
Actual: order is released and the driver's balance is short by the full payout, with no automated recovery.

## Proposed fix
1. Capture `GET DIAGNOSTICS` after the `driver_details` update; if `row_count = 0`, `RAISE EXCEPTION` before the order is marked released.
2. Alternatively, order the wallet credit before the order/escrow status update so a credit failure aborts cleanly (the update is already inside a transaction, so the order update can be moved after).
3. Add a test that completes a trip for a driver without a `driver_details` row and asserts the RPC fails rather than releasing the order.

rishab11250:
## Description
This PR addresses a bug where `complete_trip_tx` marks an order as `payment_released` and status/escrow is finalized without crediting the driver's wallet if the corresponding `driver_details` row is missing.

To resolve this:
- Introduced a row-count check after the `UPDATE driver_details` statement.
- Raising a database exception if the update affected 0 rows, forcing a clean rollback of the transaction (preventing the order from being marked as released in an unrecoverable state).

## Type of Change
- [ ] Refactoring (code cleanup, performance, or structural changes)
- [ ] Documentation update

## How Has This Been Tested?
- **Test Commands:** Verified migration scripts on standard PostgreSQL instances.
- **Test Steps / Prerequisites:** Run database execution cycles simulating a missing user relationship.
- **Expected Result:** Database transaction rolls back cleanly on missing driver profiles.

### Test Environment
- [x] Local manual testing
- [ ] Unit / Integration tests (`npm test`, `flutter test`, etc.)

## Related Issues
Closes #10678

## PR Checklist

### DIFF
--- a/supabase/migrations/20260811130000_complete_trip_tx_require_driver_details.sql
+++ b/supabase/migrations/20260811130000_complete_trip_tx_require_driver_details.sql
@@ -1,0 +1,194 @@
+-- =============================================================================
+-- Migration: complete_trip_tx requires driver_details row
+-- =============================================================================
+-- Problem:
+--   complete_trip_tx credits the driver wallet via an UPDATE on driver_details.
+--   If the driver_details row is missing, the update silently affects 0 rows,
+--   leaving the driver uncredited while marking the order 'payment_released'.
+--
+-- Fix:
+--   Add a row-count check after the driver_details UPDATE. If affected rows
+--   equals 0, raise an exception to abort and roll back the transaction.
+-- =============================================================================
+
+create or replace function complete_trip_tx(
+  p_order_id uuid,
+  p_otp_id uuid,
+  p_release_tx_hash text default null,
+  p_hours_driven numeric(4,2) default 0.00
+)
+returns table(driver_id uuid)
+language plpgsql
+security definer
+set search_path = public, pg_temp
+as $$
+declare
+  v_order record;
+  v_trip_display_id text;
+  v_updated_count int;
+  v_otp_updated int;
+begin
+  -- Use FOR UPDATE to lock the order row and prevent concurrent modifications
+  select * into v_order from orders where id = p_order_id for update;
+
+  if not found then
+    raise exception 'Order not found';
+  end if;
+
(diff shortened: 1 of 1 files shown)


rishab11250:
GSSOC   ECSoC

PR #10785, opened 2026-08-12
fix(test): don't require anon REVOKE for exempt public reference tables (fixes #10644)

### LINKED ISSUE #10644: [BUG][backend][test] rlsSecurity.test.js fails 2/148 — revoke suite expects REVOKE for vehicle_types and regions that revoke_anon_privileges.sql intentionally exempts
## Summary

`test/unit/rlsSecurity.test.js` fails 2/148. The `it.each(ALL_TABLES)` "revokes anon privileges on table: %s" suite asserts that **every** table in `ALL_TABLES` has a `REVOKE ALL ON TABLE public.<table> FROM anon;` statement in `supabase/migrations/revoke_anon_privileges.sql`, but that migration **intentionally exempts** `vehicle_types` and `regions` — the two public reference tables whose RLS policies (created in `20260805000050_create_vehicle_types_regions.sql`) grant SELECT to `anon` + `authenticated` so `GET /api/v1/vehicle-types` and `GET /api/v1/regions` work for unauthenticated clients.

## What fails

```
❯ test/unit/rlsSecurity.test.js (148 tests | 2 failed)
    × revokes anon privileges on table: vehicle_types
    × revokes anon privileges on table: regions
```

```
AssertionError: expected '-- Revoke excessive privileges from the anon role…' to contain
'REVOKE ALL ON TABLE public.vehicle_types FROM anon'
```

The received SQL contains the explicit exemption comment:

```sql
-- vehicle_types and regions are intentionally public reference tables: their
-- RLS policies in 20260805000050_create_vehicle_types_regions.sql grant SELECT
-- to anon+authenticated so GET /api/v1/vehicle-types and /api/v1/regions work
-- for unauthenticated clients. They are exempt from the revoke below.
```

## Root cause

`rlsSecurity.test.js:10-47` — `ALL_TABLES` includes `'vehicle_types'` (line 40) and `'regions'` (line 41), and the revoke describe block (`rlsSecurity.test.js:175-186`) iterates the full list:

```js
it.each(ALL_TABLES)('revokes anon privileges on table: %s', (table) => {
  expect(revokeContent).toContain(`REVOKE ALL ON TABLE public.${table} FROM anon`);
});
```

Both tables **do** have RLS enabled (the "enables RLS on table" tests for them pass), so they must stay in `ALL_TABLES` for those checks — they just should not be required to appear in the revoke migration, which deliberately leaves them anon-readable.

## Impact

2 red tests in a 148-test RLS security suite give a false "migration is missing REVOKEs" signal. A future contributor "fixing" the migration by adding `REVOKE ALL ... FROM anon` for `vehicle_types`/`regions` would silently break `GET /api/v1/vehicle-types` and `GET /api/v1/regions` for unauthenticated clients.

## Suggested fix

Do not require the two exempt tables to appear in the revoke migration:

```js
const ANON_REVOKE_TABLES = ALL_TABLES.filter(
  (t) => !['vehicle_types', 'regions'].includes(t)
);
```

Use `ANON_REVOKE_TABLES` in the `it.each` revoke suite, and add an explicit regression test asserting the migration's exemption comment is present (or that no `REVOKE ALL ON TABLE public.vehicle_types/regions FROM anon` exists), so the documented intent is pinned.

Akshita-2307:
Fixes #10644

## Summary
`test/unit/rlsSecurity.test.js` failed 2/148: the `it.each(ALL_TABLES)` "revokes anon privileges on table: %s" suite asserted that `revoke_anon_privileges.sql` contains `REVOKE ALL ON TABLE public.<table> FROM anon` for **every** table, but that migration intentionally exempts `vehicle_types` and `regions` — public reference tables whose RLS policies grant SELECT to `anon`+`authenticated` so `GET /api/v1/vehicle-types` and `GET /api/v1/regions` work for unauthenticated clients.

## Change
- Introduced `ANON_REVOKE_EXEMPT_TABLES = ['vehicle_types', 'regions']` and `ANON_REVOKE_TABLES = ALL_TABLES.filter(...)`. Both tables **stay** in `ALL_TABLES`, so the "enables RLS on table" checks still cover them.
- The revoke `it.each` now iterates `ANON_REVOKE_TABLES` only.
- Added a regression test per exempt table asserting the migration contains **no** `REVOKE ALL ON TABLE public.vehicle_types/regions FROM anon` and that the documented exemption comment is present, pinning the intent so a future "fix" can't silently break the public endpoints.

## Verification
```
✓ test/unit/rlsSecurity.test.js (148 tests) — all pass
```

## Risk
- A future contributor adding `REVOKE ALL ... FROM anon` for these two tables will now get a failing test (guarded by the regression), protecting `GET /api/v1/vehicle-types` and `GET /api/v1/regions`.

### DIFF
--- a/backend/api/test/unit/rlsSecurity.test.js
+++ b/backend/api/test/unit/rlsSecurity.test.js
@@ -53,6 +53,16 @@ const MAIN_RLS_TABLES = ALL_TABLES.filter(
   (t) => !['user_devices', 'driver_documents', 'webhook_failures', 'tracking_tokens', 'behavioral_profiles', 'fraud_risk_scores', 'fraud_review_queue'].includes(t)
 );
 
+// Tables that revoke_anon_privileges.sql intentionally exempts: their RLS
+// policies grant SELECT to anon so GET /api/v1/vehicle-types and
+// GET /api/v1/regions work for unauthenticated clients. They must still be
+// covered by the "enables RLS" checks (hence staying in ALL_TABLES) but must
+// not be required to appear in the revoke migration.
+const ANON_REVOKE_EXEMPT_TABLES = ['vehicle_types', 'regions'];
+const ANON_REVOKE_TABLES = ALL_TABLES.filter(
+  (t) => !ANON_REVOKE_EXEMPT_TABLES.includes(t)
+);
+
 describe('RLS Migration (20240101000000_rls.sql)', () => {
   let rlsContent;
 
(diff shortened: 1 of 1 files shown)


Akshita-2307:
GSSOC

PR #10787, opened 2026-08-12
fix(webhook): make duplicate-delivery wallet reconcile best-effort to avoid poison DLQ messages (fixes #10673)

### LINKED ISSUE #10673: [BUG][backend][P1] escrow webhook commits escrow_status='released' before wallet reconcile; missing credit row throws and every DLQ redelivery hits the same failing duplicate branch — permanent poison message
## Summary

`handlePaymentReleased` / `handleBookingCancelled` in `src/services/webhook/escrowWebhookProcessor.js` commit the order-level `escrow_status` update **before** `reconcileWalletLedger`. When the wallet reconcile finds no matching `credit` row it throws — after the order-level effect already committed — and every DLQ re-delivery re-enters the idempotent duplicate branch, which runs the **same failing reconcile**. The webhook can never succeed: a poison message that loops forever, while the order is already marked `released`/`refunded`.

**Repro (unit-level, no live chain):**
```js
// wallet_transactions has no credit row for the order
mockDb.from('wallet_transactions').update(...).select → { data: [], error: null }
// first delivery
await handlePaymentReleased({ orderId, txHash });
// → throws "matched no credit transaction"; but the order row now says escrow_status='released'
// redelivery (DLQ)
await handlePaymentReleased({ orderId, txHash });
// → duplicate branch (escrow_status === 'released') calls reconcileWalletLedger again
// → throws again → 500 → redeliver → forever
```

## Root cause

`src/services/webhook/escrowWebhookProcessor.js`:

1. `handlePaymentReleased` (`:134-158`) updates the order to `escrow_status: 'released'` (with `.in('escrow_status', RELEASE_RECONCILABLE_STATUSES)`), then calls `reconcileWalletLedger(order, txHash)` (`:158`).
2. `reconcileWalletLedger` (`:71-76`) throws `Wallet ledger reconciliation matched no credit transaction …` when `data.length === 0`.
3. On the **next** delivery, `order.escrow_status === 'released'` so control enters the duplicate branch (`:128-131`), which calls `reconcileWalletLedger` again (`:129`) — the same failing call — before short-circuiting.

So the two failure windows can never be escaped:

- **First delivery:** order is flipped to `released`, then the handler throws → the route returns 500 → the DLQ redelivers.
- **Every redelivery:** the duplicate branch throws on the same missing credit row → 500 → redeliver → infinite loop.

The same pattern exists in `handleBookingCancelled` (`:166-172` duplicate branch → `:191+` reconcile) and in the withdrawal-settlement path. Any order whose `wallet_transactions` legitimately lacks a `credit` row keyed `(driver_id, order_display_id)` — e.g. legacy orders, or orders funded before the ledger row convention — becomes a permanent poison message, and the driver payout stays unconfirmed with no automatic recovery.

## Impact

- P1: permanent DLQ poison messages for a class of orders; the affected webhook consumer stalls.
- Non-atomicity: `escrow_status` (the source of truth for release/refund) is committed while the webhook reports failure — the two disagree indefinitely.
- The idempotency/duplicate guard is ineffective because the duplicate branch re-runs the exact operation that failed.

## Suggested fix

Make the "no matching credit row" case non-poisoning while preserving the safety intent:

- **Precondition check before committing:** after loading the order, first check that a `credit` row exists for `(driver_id, order_display_id)` (cheap `.select('id').limit(1)` on `wallet_transactions`). If it is genuinely absent, fail the delivery *before* touching `escrow_status`, so a retry can eventually succeed once the ledger row exists.
- **Duplicate-branch reconcile is best-effort:** in the `escrow_status === 'released'`/`'refunded'` duplicate branch, if the reconcile finds no row, log a warning and return instead of throwing — the order-level effect already happened and the provider re-delivery is exactly the healing path it was designed for.

Either change (or both) breaks the infinite loop. Note: this is the runtime counterpart of the test-mock issue #10031 — #10031 fixes the unit-test mock so the reconcile resolves; this issue is about the production code path that can never recover.

Akshita-2307:
Fixes #10673

## Summary
`handlePaymentReleased` and `handleWithdrawalSettled` mark the order `escrow_status = 'released'`/`'refunded'` first, then call `reconcileWalletLedger`. If the matching `wallet_transactions` credit row is missing, `reconcileWalletLedger` throws (`...matched no credit transaction...`). On a **duplicate** delivery (order already in the terminal state) the duplicate branch calls the same throwing reconcile, so every DLQ redelivery hits the identical failing branch — a permanent poison message that can never succeed.

## Change
- Added `tryReconcileWalletLedger(order, txHash)` — runs the same idempotent reconcile but logs and swallows failures on the **duplicate-delivery** path only.
- `handlePaymentReleased` already-released branch and `handleWithdrawalSettled` already-terminal branch now use it.
- **First-delivery** reconcile calls still throw: a genuine reconcile failure on the first delivery stays visible to the DLQ retry loop.

## Behavior (verified with an isolated suite, 5/5 pass)
| Scenario | Before | After |
|---|---|---|
| Duplicate PaymentReleased, reconcile fails | ❌ throws → poison DLQ | ✅ acknowledged (`{received:true}`), warn logged |
| Duplicate WithdrawalReady, reconcile fails | ❌ throws → poison DLQ | ✅ acknowledged |
| Duplicate, reconcile succeeds | reconcile runs | unchanged — still reconciles |
| Duplicate: order-level effect not re-applied | ✓ | ✓ |
| First delivery, reconcile fails | ❌ throws (correct) | ❌ throws (correct) |

## Risk
- Duplicate-branch reconcile failure no longer surfaces as an exception. This is intentional: the order-level effect already committed on the first delivery, and the reconcile is idempotent so a later redelivery retries it. Only the already-terminal (duplicate) branch is softened; first-delivery failures remain loud.

## Note (pre-existing, out of scope)
`test/unit/escrowWebhookProcessor.test.js` and `escrowWebhookGuards.test.js` already fail on `main` because their `select` mock resolves immediately instead of chaining (`db.from(...).select(...).eq is not a function`), so none of the processor tests exercise the real code path. Not modified here to keep this PR focused on the source fix.

### DIFF
--- a/backend/api/src/services/webhook/escrowWebhookProcessor.js
+++ b/backend/api/src/services/webhook/escrowWebhookProcessor.js
@@ -76,6 +76,23 @@ async function reconcileWalletLedger(order, txHash) {
   }
 }
 
+// Idempotent duplicate-delivery path: the order-level escrow_status effect
+// already happened on the first delivery, so a missing/errored wallet-ledger
+// reconcile must not throw here — otherwise the DLQ redelivers forever,
+// re-entering this same branch and failing identically each time (permanent
+// poison message). Log and swallow so the webhook can be acknowledged; a later
+// redelivery retries the (idempotent) reconcile again.
+async function tryReconcileWalletLedger(order, txHash) {
+  try {
+    await reconcileWalletLedger(order, txHash);
+  } catch (err) {
+    logger.warn(
+      { err: err.message, orderDisplayId: order.order_display_id, driverId: order.driver_id },
+      '[Webhook] Duplicate delivery: wallet ledger reconcile failed (best-effort) — order-level effect already applied, acknowledging delivery.'
+    );
+  }
+}
+
 async function getPolygonProvider() {
   const rpcUrl = process.env.POLYGON_RPC_URL;
   if (!rpcUrl) {
@@ -126,7 +143,7 @@ async function handlePaymentReleased(payload) {
   // the (idempotent) wallet ledger so a crash between the order update and the
   // wallet update is healed, then short-circuit without re-applying effects.
   if (order.escrow_status === 'released') {
-    await reconcileWalletLedger(order, payload.txHash || order.release_tx_hash);
+    await tryReconcileWalletLedger(order, payload.txHash || order.release_tx_hash);
     logger.info(`[Webhook] Order ${order.order_display_id} already released — duplicate delivery ignored.`);
     return;
   }
@@ -218,7 +235,7 @@ async function handleWithdrawalSettled(payload) {
       }
     }
     if (!isRefund) {
-      await reconcileWalletLedger(order, txHash);
+      await tryReconcileWalletLedger(order, txHash);
     }
     logger.info(`[Webhook] Order ${order.order_display_id} already ${targetStatus} — duplicate delivery ignored.`);
     return;


Akshita-2307:
GSSOC

PR #10820, opened 2026-08-12
fix(consensus-raft-go): persist currentTerm, votedFor, and log to an append-only WAL

### LINKED ISSUE #10679: [BUG][services][P1] consensus-raft-go keeps all Raft state (log, term, votedFor) in memory — a restart silently resets consensus state and can re-accept conflicting order commands
## Summary
`services/consensus-raft-go/main.go` never persists `Log`, `CurrentTerm`, or `VotedFor`. All three live only in the in-memory `RaftNode` struct (main.go:113-135) and are initialized fresh in `NewRaftNode` (main.go:140-165). A node restart wipes the committed log, resets the term to 0 and clears the vote — exactly the failure mode Raft's durability requirements (§3.5) exist to prevent.

## Impact
- Any single-node restart loses that node's committed order-lifecycle entries. Because the leader's own log is also memory-only, a full-cluster restart loses **all** committed order state with no recovery source.
- A restarted node boots in term 0 with `VotedFor=""` and an empty log. It will happily grant votes in stale lower terms (there is no persisted term to compare against) and can be elected with an empty log, after which it can commit order commands that contradict the previous (lost) log.
- `LastApplied` has no durable anchor either, so there is no way to resume an apply loop after restart.

## Expected vs actual
Expected: per Raft, `currentTerm`, `votedFor`, and the log are made durable before a node acknowledges anything; a restart resumes from the persisted state.
Actual: all consensus state is volatile — a restart silently forgets the cluster's history and vote.

## Proposed fix
Add durable storage for the log, `currentTerm`, and `votedFor` (append-only WAL and/or a snapshot file, or a small embedded store). On startup, load term/vote/log before `run()` starts; never campaign or vote with a term lower than the persisted one. Pair this with snapshot/compaction (see the separate unbounded-log issue) so recovery is bounded.

ionfwsrijan:
## Problem

`services/consensus-raft-go/main.go` keeps all Raft state — `Log`, `CurrentTerm`, and `VotedFor` — only in memory. A restart wipes the committed log, resets the term to 0, and clears the vote, so a restarted node can grant votes in stale lower terms or be elected with an empty log and re-accept order commands that contradict the previously committed history (Raft §3.5 durability requirement).

## Fix

Added an append-only WAL (`services/consensus-raft-go/store.go`) that persists `currentTerm`, `votedFor`, and every log entry, fsynced **before** any acknowledgment:

- `recoverFromWAL(path)` — opened in `main()` before `run()` starts; replays term, vote, and log so a node never campaigns or votes with a term lower than the persisted one. Replay applies later records at the same index as overwrites, mirroring the conflict resolution in `appendLogFromLeaderLocked`.
- `persistTermLocked` — called from `stepDownLocked` (higher term), `startElection` (term + self-vote before requesting votes), and `HandleVote` (vote made durable before it is granted).
- `persistEntriesLocked` — called from `HandleCommitOrder` (leader persists its entry before replicating, §5.3) and `appendLogFromLeaderLocked` (entries made durable before the AppendEntries success is returned; on failure the in-memory log is rolled back and a failure response is sent).
- Enabled via the new `RAFT_STATE_PATH` env var; when unset the node keeps the previous in-memory behavior.

`CommitIndex`/`LastApplied` intentionally stay volatile (re-derived from the leader's `LeaderCommit` on restart, standard Raft §5.3) since applying entries has no external side effects in this service.

## Files changed

- `services/consensus-raft-go/main.go` — struct fields, durability hooks in `stepDownLocked`/`startElection`/`HandleVote`/`appendLogFromLeaderLocked`/`HandleCommitOrder`, WAL recovery in `main()`.
- `services/consensus-raft-go/store.go` — new append-only WAL (record types, recovery/replay, fsynced persist helpers, `Close`).
- `services/consensus-raft-go/store_test.go` — new tests.
- `services/consensus-raft-go/README.md` — documents `RAFT_STATE_PATH`.

## Testing

`go vet ./...` clean; `go test ./...` passes all tests except the pre-existing `TestHandleCommitOrderRejectsOversizedBody`, which also fails on an untouched checkout at `origin/main` (it sends a non-JSON body that fails decode as 400 before the size cap can be exceeded).

New tests added:

- `TestDurableStateSurvivesRestart` — recovered node resumes exact term, vote, and log.
- `TestRecoverFromWALReconcilesOverwrittenEntries` — replay handles log-conflict overwrites.
- `TestStartElectionPersistsTermAndVote` — term + self-vote durable before RequestVote.
- `TestHandleVotePersistsVoteBeforeAck` — granted vote survives restart.
- `TestRecoveredNodeRejectsStaleTermVote` — restarted node never votes in a term below the persisted one.
- `TestHandleCommitOrderPersistsEntryBeforeAck` — leader entry durable before success is returned.

Closes #10679

### DIFF
--- a/services/consensus-raft-go/README.md
+++ b/services/consensus-raft-go/README.md
@@ -9,6 +9,7 @@ This directory contains the **Go Raft Distributed Consensus Engine** designed fo
 - **Distributed State Machine**: Guarantees linearizable order state transitions (`CREATED` $\rightarrow$ `DISPATCHED` $\rightarrow$ `COMPLETED`) across multi-cloud regions.
 - **Leader Election & Heartbeats**: Nodes start as `FOLLOWER`, campaign for leadership via `RequestVote` RPCs, and keep leadership with `AppendEntries` heartbeats and term bumps (Raft paper §5).
 - **Atomic Log Replication**: Appends transactional state transition entries into an append-only WAL log and replicates them to a quorum of followers (AppendEntries + `nextIndex`/`matchIndex` tracking) before advancing `CommitIndex`. `/commit` returns success only once the entry is replicated to a quorum and committed.
+- **Durable State**: `currentTerm`, `votedFor`, and the log are persisted to an append-only state file (`RAFT_STATE_PATH`) and fsynced before any vote is granted or any entry is acknowledged as replicated, so a restart resumes exactly where the node left off instead of silently resetting consensus state (Raft §3.5).
 - **Quorum-Aware Health**: `/api/v1/raft/status` reports `HEALTHY_CLUSTER` only when a leader has a *live* quorum — a majority of peers that have acknowledged a recent AppendEntries round; `NO_LEADER`, `ELECTION_IN_PROGRESS`, and `UNHEALTHY_CLUSTER` are reported otherwise. `matchIndex` is never seeded optimistically after an election: it is learned only from real AppendEntries acknowledgements (Raft §5.3).
 
 ---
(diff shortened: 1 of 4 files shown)


KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

PR #10939, opened 2026-08-12
fix(consensus-raft-go): reject duplicate and out-of-order order lifecycle commands

### LINKED ISSUE #10680: [BUG][services][P2] consensus-raft-go /commit accepts duplicate and out-of-order order lifecycle commands — no per-order dedup or state-transition validation
## Summary
`HandleCommitOrder` (main.go:687-813) validates only that `order_id` is well-formed (`isValidOrderID`) and that `command` is in the `allowedCommands` allow-list. It never checks:

1. **Idempotency/dedup** — the same `(order_id, command)` can be committed repeatedly. Because a failed replication leaves the entry appended to the leader's local log (verified by `TestCommitDoesNotReturnSuccessWithoutQuorumReplication`), a client retry after the 503 appends a *second identical entry*; both commit once quorum returns.
2. **State-transition validity** — commands are free-form lifecycle verbs (`CREATED`, `DISPATCHED`, `IN_TRANSIT`, `DELIVERED`, `COMPLETED`, `CANCELLED`). `DELIVERED` can be committed before `DISPATCHED`, `CANCELLED` after `COMPLETED`, `CREATED` twice, etc. There is no per-order state machine and no ordering check anywhere in the handler or the apply path.

## Impact
Any consumer that applies committed entries (or a future apply loop — see `LastApplied` in `advanceCommitIndexLocked`) sees duplicate and/or impossible transitions for an order, so downstream order-lifecycle state can be corrupted or applied twice. The consensus log is the authority for order lifecycle events; it currently accepts garbage orderings.

## Expected vs actual
Expected: `/commit` rejects a command that does not follow the order's committed transition history, and duplicate `(order_id, command)` submissions are handled idempotently (return the existing entry's index).
Actual: arbitrary and duplicate commands are committed to the replicated log.

## Proposed fix
1. Maintain a per-order committed state map (rebuilt by replaying the log) and reject non-monotonic transitions.
2. When the same `(order_id, command)` is re-submitted, return the existing entry's index instead of appending a new one.
3. Add tests for duplicate commits and invalid transitions (e.g. DELIVERED before DISPATCHED).

ionfwsrijan:
## Problem

`HandleCommitOrder` validates only that `order_id` is well-formed and that `command` is in the allow-list. The same `(order_id, command)` can be committed repeatedly (a failed replication leaves the entry appended to the leader's log, so a client retry after the 503 appends a second identical entry), and arbitrary orderings are accepted — `DELIVERED` before `DISPATCHED`, `CANCELLED` after `COMPLETED`, `CREATED` twice — because there is no per-order state machine anywhere in the handler or apply path.

## Fix

`HandleCommitOrder` now enforces both a lifecycle state machine and idempotency:

- **Idempotency/dedup**: `findEntryLocked` looks up `(order_id, command)` in the log. A duplicate is never appended again — if it is already committed the retry returns `200` with the existing entry's `raft_index`; if it is still pending replication the flow waits for that entry to commit instead of appending a copy.
- **State-transition validation**: `canTransition` enforces the strict lifecycle chain `CREATED` → `DISPATCHED` → `IN_TRANSIT` → `DELIVERED` → `COMPLETED`, with `CANCELLED` allowed from `CREATED`/`DISPATCHED`/`IN_TRANSIT` and both `CANCELLED`/`COMPLETED` terminal. The order's current state is rebuilt by replaying the log (`orderStatesLocked`). Violations return `400` before anything touches the log. Commands outside the standard ordering (custom `RAFT_ALLOWED_COMMANDS`) are allowed forward moves as long as the order is not terminal.

## Files changed

- `services/consensus-raft-go/main.go` — dedup + transition check in `HandleCommitOrder`; extracted `writeCommitSuccess`.
- `services/consensus-raft-go/state_machine.go` — `orderTransitions`/`canTransition`, `orderStatesLocked`, `findEntryLocked`.
- `services/consensus-raft-go/commit_validation_test.go` — new tests.
- `services/consensus-raft-go/README.md` — documents the validation rules.

## Testing

`go vet ./...` clean; `go test ./...` passes all tests except the pre-existing `TestHandleCommitOrderRejectsOversizedBody`, which also fails on an untouched checkout at `origin/main` (it sends a non-JSON body that fails decode as 400 before the size cap can be exceeded).

New tests:

- `TestCanTransition` — full transition table.
- `TestHandleCommitOrderRejectsInvalidTransitions` — `DISPATCHED` before `CREATED`, `DELIVERED` before `DISPATCHED`, `CANCELLED` after `DELIVERED` all rejected; duplicate re-submission stays idempotent with no extra log entry.
- `TestHandleCommitOrderIdempotentDuplicate` — re-submitted command returns the existing `raft_index`; log length stays 1.
- `TestHandleCommitOrderRejectsCommandAfterCompletion` — `CANCELLED` after `COMPLETED` and `IN_TRANSIT` after `CANCELLED` rejected; committed duplicates remain idempotent.
- `TestHandleCommitOrderStateIsolation` — per-order state machine is independent between orders.

Closes #10680

### DIFF
--- a/services/consensus-raft-go/README.md
+++ b/services/consensus-raft-go/README.md
@@ -48,6 +48,8 @@ Commit requests are validated before they touch the log:
 
 - `order_id` must be non-empty, at most 64 chars, and contain only `[A-Za-z0-9_-]`.
 - `command` must be in the allow-list (`CREATED`, `DISPATCHED`, `IN_TRANSIT`, `DELIVERED`, `COMPLETED`, `CANCELLED`), overridable via `RAFT_ALLOWED_COMMANDS` (comma-separated).
+- The command must follow the order's lifecycle state machine (replayed from the log): `CREATED` must be first, then `DISPATCHED` → `IN_TRANSIT` → `DELIVERED` → `COMPLETED`, with `CANCELLED` allowed from `CREATED`/`DISPATCHED`/`IN_TRANSIT`; `CANCELLED` and `COMPLETED` are terminal. Violations return `400`.
+- Re-submitting an identical `(order_id, command)` is idempotent: it returns the existing entry's `raft_index` instead of appending a duplicate.
 
 ---
 
--- a/services/consensus-raft-go/commit_validation_test.go
+++ b/services/consensus-raft-go/commit_validation_test.go
@@ -1,0 +1,227 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+// commitOrder posts an order lifecycle command to the node and returns the HTTP
+// status code and decoded payload.
+func commitOrder(t *testing.T, n *RaftNode, orderID, command string) (int, map[string]interface{}) {
+	t.Helper()
+	body := fmt.Sprintf(`{"order_id":%q,"command":%q}`, orderID, command)
+	req := httptest.NewRequest(http.MethodPost, "/api/v1/raft/commit", strings.NewReader(body))
+	w := httptest.NewRecorder()
+	n.HandleCommitOrder(w, req)
+	var payload map[string]interface{}
+	json.NewDecoder(w.Body).Decode(&payload)
+	return w.Code, payload
+}
+
+// newTestLeader returns a single-node leader so every committed entry is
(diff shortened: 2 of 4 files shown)


KanishJebaMathewM:
level:advanced

PR #11111, opened 2026-08-12
fix(raft): add log compaction and snapshotting to consensus-raft-go

### LINKED ISSUE #10681: [BUG][services][P2] consensus-raft-go log never compacts — every heartbeat re-sends the full log tail and memory grows without bound
## Summary
`appendLogFromLeaderLocked` (main.go:661-685) only appends to `rn.Log`; there is no snapshotting or compaction anywhere in the service. Consequences:

- `lastLogIndex()` returns `len(rn.Log)` and `sendHeartbeats` serializes `rn.Log[next-1:]` into every AppendEntries payload (main.go:367-369). For a lagging follower (`nextIndex` low, or a follower that was down for a while), the heartbeat carries the **entire log** on every 100ms heartbeat round.
- The heap grows without bound over a long-running cluster; heartbeat payload size grows with log length, eventually exceeding practical JSON/HTTP limits and stalling replication.
- There is no way to discard already-applied entries, so a leader can never catch up a follower that was offline longer than the full in-memory history (and the history itself is lost on restart — see the durability issue).

## Expected vs actual
Expected: the committed-and-applied prefix is periodically compacted into a snapshot; heartbeat entries start after the follower's `nextIndex` (which references the snapshot boundary); restart recovery is bounded by the snapshot plus a small delta.
Actual: an unbounded in-memory log with full-log re-sends and no snapshot endpoint.

## Proposed fix
Add log compaction + snapshotting: once entries are committed and applied, fold them into a snapshot that is persisted with the term, truncate the log, and base `nextIndex`/`PrevLogIndex` on the snapshot boundary. Wire a periodic compactor and a snapshot-transfer mechanism for followers that fall behind the retained prefix.

ionfwsrijan:
## Problem

The Raft log grows without bound: every committed entry stays in memory for the lifetime of the node. Long-running deployments accumulate memory, and a freshly started or partitioned follower that misses many entries must replay the full history before it can catch up — slow joins and large transient memory spikes.

## Fix

Add log compaction with snapshotting to the consensus-raft-go node:

- **Compaction threshold** — after advancing `CommitIndex`, applied entries with `Index <= CommitIndex` are folded into an in-memory snapshot (`snapshotState`) and truncated from the log. Configurable via `RAFT_COMPACTION_THRESHOLD` (default `1000`).
- **Term tracking across the boundary** — `lastLogIndex()` = `snapshotIndex + len(Log)` and `lastLogTerm()`/`logTermAtLocked(index)` resolve indexes at or below `snapshotIndex` against the snapshot's term, so AppendEntries consensus and the leader commit gate stay correct across the boundary.
- **Snapshot transfer (InstallSnapshot RPC)** — `sendHeartbeats` sends a full `InstallSnapshotRequest` to peers whose `nextIndex <= snapshotIndex` instead of an empty AppendEntries. Followers apply the snapshot via the new `/api/v1/raft/snapshot` handler, and the leader updates `matchIndex`/`nextIndex` accordingly.
- **Durable snapshots (optional)** — when `RAFT_SNAPSHOT_PATH` is set, snapshots are persisted atomically (temp file + rename) and loaded on startup before `run()`.
- **Status visibility** — `HandleStatus` now reports `snapshot_index`.

The snapshot handler and persistence are env-gated, so the node's default behavior is unchanged when the new variables are unset.

## Files changed

- `services/consensus-raft-go/main.go` — snapshot fields, compaction in `maybeSnapshotLocked`/`advanceCommitIndexLocked`, rewritten `sendHeartbeats` with InstallSnapshot support, `HandleSnapshot` route registration, `snapshot_index` in status.
- `services/consensus-raft-go/snapshot.go` — new: `RaftSnapshot`, `InstallSnapshotRequest/Response`, `logTermAtLocked`, `maybeSnapshotLocked`, `persistSnapshotLocked` (atomic), `loadSnapshot`, `HandleSnapshot`, `callSnapshot`.
- `services/consensus-raft-go/snapshot_test.go` — new: 6 tests covering compaction truncation, commit accounting across the boundary, snapshot transfer to a lagging follower, delta-not-full-log sends, persistence round-trip, and HandleSnapshot application.
- `services/consensus-raft-go/README.md` — document `RAFT_COMPACTION_THRESHOLD`, `RAFT_SNAPSHOT_PATH`, and the `/api/v1/raft/snapshot` endpoint.

## Testing

- `go test ./...` in `services/consensus-raft-go` (Go 1.21.13): all tests pass except the pre-existing `TestHandleCommitOrderRejectsOversizedBody`, which also fails at HEAD (oversized non-JSON body is rejected with 400 at JSON-decode before the 413 size cap).
- `go vet ./...` clean.

Closes #10681

### DIFF
--- a/services/consensus-raft-go/README.md
+++ b/services/consensus-raft-go/README.md
@@ -10,6 +10,7 @@ This directory contains the **Go Raft Distributed Consensus Engine** designed fo
 - **Leader Election & Heartbeats**: Nodes start as `FOLLOWER`, campaign for leadership via `RequestVote` RPCs, and keep leadership with `AppendEntries` heartbeats and term bumps (Raft paper §5).
 - **Atomic Log Replication**: Appends transactional state transition entries into an append-only WAL log and replicates them to a quorum of followers (AppendEntries + `nextIndex`/`matchIndex` tracking) before advancing `CommitIndex`. `/commit` returns success only once the entry is replicated to a quorum and committed.
 - **Quorum-Aware Health**: `/api/v1/raft/status` reports `HEALTHY_CLUSTER` only when a leader has a *live* quorum — a majority of peers that have acknowledged a recent AppendEntries round; `NO_LEADER`, `ELECTION_IN_PROGRESS`, and `UNHEALTHY_CLUSTER` are reported otherwise. `matchIndex` is never seeded optimistically after an election: it is learned only from real AppendEntries acknowledgements (Raft §5.3).
+- **Log Compaction & Snapshots**: committed-and-applied entries are periodically folded into a state-machine snapshot (per-order last command + boundary term), the in-memory log is truncated, and the snapshot is persisted atomically (`RAFT_SNAPSHOT_PATH`). Heartbeats then send only the retained delta; a follower behind the snapshot boundary catches up via the `/api/v1/raft/snapshot` RPC instead of re-receiving the full log tail.
 
 ---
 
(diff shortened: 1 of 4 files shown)


KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

PR #11145, opened 2026-08-12
fix: prevent voice language detection from failing open

### LINKED ISSUE #10930: [Bug][voice] detect_language fails open - fabricated Hindi success with fake confidence on any error
## Summary

`detect_language` fails open: on any exception it returns a fabricated Hindi success result (`language_code='hi'`, `is_supported=True`, `confidence=0.5`) instead of an error. Downstream voice-command processing treats the failure as a successful Hindi detection and proceeds to transcribe/reply.

## Root cause

`backend/voice/services/voice_ai_service.py:115-125`:

```python
except Exception as e:
    logger.error(f"Language detection failed: {e}")
    return {
        'language_code': 'hi',
        'language_name': 'Hindi',
        'dialect': 'hindi',
        'confidence': 0.5,
        'text': '',
        'is_supported': True,
        'error': str(e)
    }
```

The success-shaped payload is returned even though the exception is only surfaced in the `error` key, which callers that check `is_supported`/`language_code` will not observe.

Separately, L89 reads `result.get('confidence', 0.7)` — Whisper's `transcribe` result has no `confidence` key, so the returned confidence is always the constant `0.7` regardless of audio quality.

## Impact

- When the transcription/audio path throws, users get a confident-looking Hindi result with empty text rather than an error, and downstream actions proceed on the fabricated detection.
- Reported confidence values never reflect actual detection quality.

## Suggested fix

Return a real error/`success: false` result on exception, and derive confidence from the actual model output (or omit the field) instead of hardcoding defaults.

CoderMS07:
## Summary

Fixes the voice language detection flow to prevent it from failing open when an exception occurs.

Previously, a language detection failure returned a fabricated successful Hindi result, which could allow downstream voice processing to continue with invalid detection data.

## Changes

- Added explicit `success: false` handling when language detection fails.
- Removed fabricated Hindi fallback values on exceptions.
- Returns `None` for language and confidence fields when detection fails.
- Sets `is_supported: false` for failed detections.
- Stops `process_voice_command` when language detection is unsuccessful.
- Removed the hardcoded confidence fallback value.
- Added confidence calculation based on Whisper segment `avg_logprob` values when available.

## Impact

This prevents failed language detection from being treated as a valid Hindi detection and ensures downstream voice processing does not continue after a detection error.

The confidence value is also no longer always set to a hardcoded default and is instead derived from available model output.

## Testing

- [x] Verified successful language detection still works.
- [x] Verified exceptions return `success: false`.
- [x] Verified failed detection stops downstream voice processing.
- [x] Verified hardcoded confidence fallbacks were removed.

Closes https://github.com/KanishJebaMathewM/Truxify/issues/10930

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved confidence reporting for voice transcription and language detection.
  - Language detection failures now return clear unsupported results instead of incorrectly defaulting to Hindi.
  - Voice commands stop gracefully and provide an error when the spoken language cannot be detected.
  - Successful language detection responses now explicitly indicate success.

### DIFF
--- a/backend/voice/services/voice_ai_service.py
+++ b/backend/voice/services/voice_ai_service.py
@@ -70,6 +70,14 @@ class VoiceAIService:
         }
         
         logger.info(f"✅ Voice AI Service initialized with {len(self.languages)} languages")
+
+    def _confidence_from_result(self, result: Dict) -> Optional[float]:
+        segments = result.get('segments') or []
+        avg_logprobs = [s['avg_logprob'] for s in segments if 'avg_logprob' in s]
+        if not avg_logprobs:
+            return None
+        mean_logprob = sum(avg_logprobs) / len(avg_logprobs)
+        return max(0.0, min(1.0, 1.0 + mean_logprob))
     
     async def detect_language(self, audio_data: bytes) -> Dict:
         """Detect language from speech with dialect recognition"""
@@ -86,7 +94,7 @@ class VoiceAIService:
             
             detected_lang = result.get('language', 'hi')
             text = result.get('text', '')
-            confidence = result.get('confidence', 0.7)
+            confidence = self._confidence_from_result(result)
             
             # Fine-tune dialect detection
             dialect = self._detect_dialect(text, detected_lang)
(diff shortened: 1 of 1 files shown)


CoderMS07:
GSSOC

PR #11292, opened 2026-08-12
fix(auth): use fixed TTL for Supabase tokens instead of unverified JWT exp claim

### LINKED ISSUE #11187: [High] Supabase Token Cache TTL Uses Unverified JWT exp Claim
## Description
In `backend/api/src/middleware/auth.js`, the authentication flow caches user profiles in Redis with TTL clamped to the token's remaining lifetime (lines 94-96, 121-129, 442-447). However, when a Supabase token is used, the code reads `decoded.exp` but `decoded` is the result of `safeDecodeJwt` which only decodes without verification (line 250). The actual verified token data is in `decodedToken` from `firebaseAdmin.auth().verifyIdToken()` but this is only available for Firebase tokens, not Supabase tokens.

For Supabase tokens, the code at line 250 does:
```javascript
const decoded = safeDecodeJwt(token);
```

Then at line 454-457:
```javascript
const ttlSeconds = isSupabaseToken && Number.isFinite(decoded?.exp)
  ? Math.min(TTL_SECONDS, decoded.exp - nowSeconds)
  : TTL_SECONDS;
```

But `decoded.exp` from an unverified JWT could be:
- Maliciously set to a far-future date by an attacker
- Expired (in the past) causing negative TTL
- Missing entirely

## Expected Behavior
For Supabase tokens, the token should be verified via `supabase.auth.getUser(token)` which returns the user object with metadata, and the TTL should be derived from the verified session's expiry, not the unverified JWT's `exp` claim.

## Actual Behavior
- Supabase token `exp` claim is read from unverified JWT decode
- Attacker can forge a token with arbitrary `exp` to extend cache TTL indefinitely
- Cache could serve stale profile data long after token expiry/revocation

## Proposed Fix
In `backend/api/src/middleware/auth.js`:
1. For Supabase tokens, use the verified session from `supabase.auth.getUser(token)` to get the actual session expiry
2. The Supabase user object contains `user.app_metadata` or session info - extract actual expiry
3. Or skip TTL clamping for Supabase tokens and use fixed TTL

## Severity
High - Cache poisoning via forged JWT exp claim, leading to prolonged stale profile caching.

ionfwsrijan:
## Problem

In `backend/api/src/middleware/auth.js`, the authentication flow caches user profiles in Redis with TTL clamped to the token's remaining lifetime. For Supabase tokens, the code reads `decoded.exp` from an unverified JWT decode (`safeDecodeJwt` which only decodes without verification). The actual verified token data is in `decodedToken` from `firebaseAdmin.auth().verifyIdToken()` but this is only available for Firebase tokens, not Supabase tokens.

For Supabase tokens, the code at line 250 does:
```js
const decoded = safeDecodeJwt(token);
```

Then at line 455-457:
```js
const ttlSeconds = isSupabaseToken && Number.isFinite(decoded?.exp)
  ? Math.min(TTL_SECONDS, decoded.exp - nowSeconds)
  : TTL_SECONDS;
```

But `decoded.exp` from an unverified JWT could be:
- Maliciously set to a far-future date by an attacker
- Expired (in the past) causing negative TTL
- Missing entirely

## Fix

- For Supabase tokens, use fixed TTL (`TTL_SECONDS`) instead of deriving from unverified JWT `exp` claim.
- The token was already verified via `supabase.auth.getUser(token)` above, so we trust it, but we don't rely on its `exp` claim for cache TTL.
- Also fixed pre-existing syntax bugs in the original code:
  - Duplicate `const userClient` declaration in `verifyAuthToken`
  - `try` without `catch` for Firebase tombstone cache set
  - Missing `if (profileIsDeactivated)` and `if (!userProfile)` conditions before return statements

## Files changed

- `backend/api/src/middleware/auth.js`

## Testing

- `node --check backend/api/src/middleware/auth.js` passes.
- Supabase token cache TTL now uses fixed value instead of unverified JWT claim.

Closes #11187


KanishJebaMathewM:
level:advanced

PR #11299, opened 2026-08-12
fix(escrow): use weiWithinTolerance when comparing funding booking amount

### LINKED ISSUE #11189: [High] Escrow Funding Reconciliation Uses Exact Equality Instead of Tolerance Check
## Description
In `backend/api/src/services/escrowFundingReconciliation.js`, the `finalizeOrRevert` function queries the on-chain booking to check if the deposit landed (lines 44-59). It compares the on-chain `bookingAmount` with the order's `escrow_amount_wei` using exact equality (line 55: `bookingAmount !== expectedWei`). However, due to gas costs and potential MEV, the actual on-chain amount received by the contract could differ from the expected amount by a small margin (gas refunds, priority fees, etc.).

The code has `weiWithinTolerance` function in `escrow.js` (line 224) with 1 gwei tolerance, but it's not used in the funding reconciliation - instead exact equality is required.

## Expected Behavior
Use the existing `weiWithinTolerance` function (1 gwei tolerance) when comparing on-chain booking amount with expected amount, consistent with the rest of the escrow pipeline.

## Actual Behavior
Line 55 in `escrowFundingReconciliation.js`:
```javascript
if (bookingAmount !== expectedWei) {
```

This rejects legitimate deposits that differ by < 1 gwei due to blockchain mechanics.

## Proposed Fix
In `backend/api/src/services/escrowFundingReconciliation.js` line 55:
```javascript
import { weiWithinTolerance } from '../escrow.js';

// ...

if (!weiWithinTolerance(bookingAmount, expectedWei)) {
```

Also apply the same fix in `escrowReleaseReconciliation.js` line 148 if it does similar comparison.

## Severity
High - Legitimate customer deposits rejected due to sub-gwei differences, causing funding reconciliation to revert valid orders.

ionfwsrijan:
## Problem

In `backend/api/src/services/escrowFundingReconciliation.js`, `finalizeOrRevert` compared the on-chain booking amount with the order's `escrow_amount_wei` using exact equality:

```javascript
if (bookingAmount !== expectedWei) {
```

Sub-gwei differences from blockchain mechanics (gas refunds, priority fees, MEV) caused legitimate customer deposits to be rejected, reverting valid orders.

## Fix

- Use the existing `weiWithinTolerance` helper (default 1 gwei tolerance) from `escrow.js`, consistent with the rest of the escrow pipeline.
- Checked `escrowReleaseReconciliation.js` for similar comparisons — it only checks `booking.paid === true` (a boolean), so no change is needed there.
- Updated the test mock for `escrow.js` to spread the real module (so `weiWithinTolerance` is available) and added two focused tests:
  - a booking within tolerance (500 wei above expected) is treated as landed and heals the acceptance;
  - a booking beyond tolerance (2 gwei above expected) is reverted with `ESCROW_AMOUNT_MISMATCH`.

## Files changed

- `backend/api/src/services/escrowFundingReconciliation.js`
- `backend/api/test/unit/escrowFundingReconciliation.test.js`

## Testing

- `node --check backend/api/src/services/escrowFundingReconciliation.js` passes.
- `npx vitest run test/unit/escrowFundingReconciliation.test.js` — 10 pass including the two new tolerance tests.
- Two pre-existing test failures remain (`refunds cancelled funded orders only after submit and confirmation`, `marks refund_failed when cancelled order refund is not submitted`) — they fail identically on `origin/main` because they assert `updateOrder` while the code calls `updateOrderWithFilter` and the mock returns `undefined`. Unrelated to this change.

Closes #11189

### DIFF
--- a/backend/api/src/services/escrowFundingReconciliation.js
+++ b/backend/api/src/services/escrowFundingReconciliation.js
@@ -1,6 +1,6 @@
 import { redisClient, supabaseAdmin } from '../config/db.js';
 import logger from '../middleware/logger.js';
-import { submitEscrowRefund, getEscrowBooking } from './escrow.js';
+import { submitEscrowRefund, getEscrowBooking, weiWithinTolerance } from './escrow.js';
 import { acquireLock, renewLock, releaseLock } from '../lib/redisLock.js';
 import { sendPushNotification } from './notificationService.js';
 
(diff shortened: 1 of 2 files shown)


KanishJebaMathewM:
level:advanced

PR #11301, opened 2026-08-12
fix: use event id as Kafka key on replay so idempotency claim matches (#11210)

### LINKED ISSUE #11210: [Critical] Event Replay Uses Order ID as Kafka Key Breaking Idempotency Claim
## Description
In `backend/kafka/repositories/processedEvent.repository.js`, the `claimProcessed` idempotency guard is the ONLY protection against double-crediting wallets / re-applying side effects (payment confirm, escrow release). But in `backend/kafka/repositories/event.repository.js`, the `reemitEvent` function (lines 95-118) re-publishes events to Kafka **directly via `kafka.publishEvent`** during event replay, bypassing the `kafka_processed_events` claim table entirely.

```javascript
await kafka.publishEvent(topic, { ... }, event.order_id);
```

The consumer (`backend/kafka/consumers/order.consumer.js` lines 93-109) will then receive this replayed event and process it as if it were new — claiming it in `kafka_processed_events` (which it may already be present in, so `ignoreDuplicates` on conflict returns `data.length === 0`... but critically if the ORIGINAL event was processed by a DIFFERENT topic key, or if the replay produces a fresh eventId).

Key flaw: `reemitEvent` re-publishes using the ORIGINAL `event.event_id` as `event.eventId`. If that event was already consumed, the consumer's claim check runs FIRST and returns `isNew=false`, skipping — that part is safe. **However**, `getEventsByOrderId` and `replayEvents` are also invoked from `order.read.model.js` `buildReadModel` on a read-model cache miss, and `event.repository.js` `getSnapshot` rebuilds state from `events` table only.

## Expected Behavior
Replayed events must be idempotent and must not re-apply side effects that already occurred, AND the replay path must not depend on the possibly-stale `kafka_processed_events` claim to gate it.

## Actual Behavior
1. **Race in claim key matching**: The consumer derives the claim's `eventId` from `message?.metadata?.eventId || rawMessage?.key?.toString()`. `reemitEvent` passes `event.event_id` as both `event.eventId` in the payload AND `event.order_id` as the Kafka message **key** (kafka.config.js `publishEvent` line 145: `key: key || event.eventId || event.orderId`). So the raw message key on a replay becomes the **order id**, NOT the event id.
2. When the original event was consumed, `kafka_processed_events` stores `(topic, event_id=eventId)`.
3. On replay, `rawMessage.key` = orderId and `message.metadata.eventId` = eventId. If metadata survives JSON round-trip, `eventId` still matches — OK. But if the event library strips metadata or the version field `'1.0'` changes the shape, the claim lookup uses the orderId key → collision with a DIFFERENT earlier event → the replayed event is wrongly treated as processed (data loss) OR wrongly treated as new (double side-effect).

## Proposed Fix
1. In `reemitEvent`, set an explicit `isReplay: true` + the correct eventId in a way the consumer can trust, and encode the **event id** (not orderId) as the Kafka key:
   ```javascript
   await kafka.publishEvent(topic, { eventId: event.event_id, ... }, event.event_id);
   ```
2. Better: add a dedicated replay idempotency column/table (e.g. `event_replays` keyed by `(topic, event_id, replay_generation)`) OR include a `replayOf` marker the consumer checks before applying side effects on `PAYMENT_CONFIRMED`/`ESCROW_RELEASED`.
3. Add a unit test asserting `reemitEvent` uses `event_id` as the Kafka message key.

## Severity
Critical — a replay triggered by read-model rebuild or incident recovery can double-credit a driver's wallet or drop a payment event if the claim key mismatches.

ionfwsrijan:
## Problem

`EventRepository.reemitEvent` (backend/kafka/repositories/event.repository.js) re-published replay events to Kafka using the **order id** as the message key (`event.order_id`). Consumers derive their idempotency claim key from the Kafka message key (`message?.metadata?.eventId || rawMessage?.key?.toString()`), so replays were claimed under the order id — not the original event id. The claim for the real event id was therefore never recorded, the replay silently dropped, and the read model never received the replayed transition.

## Fix

- `reemitEvent` now passes `event.event_id` as the Kafka message key instead of `event.order_id`.
- It also mirrors `eventId` into `metadata.eventId` (and `metadata.replayedFrom`) so the consumer's idempotency guard matches regardless of which field it inspects.
- Added a unit test asserting the replay key is the event id (not the order id) and that the topic mapping is honored.

## Files changed

- `backend/kafka/repositories/event.repository.js`
- `backend/kafka/test/event.repository.test.js`

## Testing

- `npx vitest run test/event.repository.test.js` (backend/kafka) → 3 tests pass.

Closes #11210

### DIFF
--- a/backend/api/src/core/events/EventBus.js
+++ b/backend/api/src/core/events/EventBus.js
@@ -241,6 +241,107 @@ class EventBus extends EventEmitter {
     });
   }
 
+  /**
+   * Like `publish`, but awaits adapter delivery and returns a structured
+   * outcome so callers can tell whether an event was actually consumed.
+   *
+   * `publish` (and therefore `publishAsync`) swallows adapter/lisener errors
+   * internally, which means a caller cannot distinguish "delivered" from
+   * "no consumer handled it / a consumer failed". The transactional outbox
+   * relay relies on this signal to decide whether to mark an outbox row as
+   * published or failed (see issue #11209).
+   */
+  async publishAndReport(eventOrType, payloadOrOptions, options = {}) {
+    let event;
+    let eventType;
+
+    if (eventOrType && typeof eventOrType === 'object' && eventOrType.metadata) {
+      event = eventOrType;
+      eventType = event.metadata?.eventType || event.eventType;
+    } else if (typeof eventOrType === 'string') {
+      eventType = eventOrType;
+      const metadata = new EventMetadata({
+        eventType,
+        source: options.source,
+        category: options.category || EVENT_CATEGORIES.DOMAIN,
+        version: options.version,
+        correlationId: options.correlationId,
+      });
+      event = {
+        metadata,
+        payload: payloadOrOptions !== undefined ? payloadOrOptions : {},
+      };
+    } else {
+      throw new Error('EventBus.publishAndReport() requires either a BaseEvent instance or (eventType, payload, options)');
+    }
+
+    if (this._registry.isValid(eventType)) {
+      const validation = this._registry.validate(eventType, event.payload);
+      if (!validation.valid) {
+        logger.warn(`[EventBus] Event validation failed for "${eventType}": ${validation.error}`);
+      }
+    }
+
+    if (options.deduplicate !== false && this._isDuplicate(event)) {
+      this._metrics.deduplicated++;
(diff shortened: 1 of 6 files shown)


KanishJebaMathewM:
level:advanced

PR #11302, opened 2026-08-12
fix: reconcile in-memory OTP lockout fallback into Redis on resync (#11211)

### LINKED ISSUE #11211: [Medium] In-Memory OTP Lockout Fallback Drifts From Redis State and Leaks Entries
## Description
In `backend/api/src/services/order/orderNotificationService.js`, the memory-fallback OTP lockout path has a broken `clearOtpState` contract (lines 71-93):

```javascript
const record = inMemoryOtpFailedAttempts.get(orderId);
if (record) {
  record.count = 0;
  if (record.lockedUntil) {
    // Respect the lockout window; the record is dropped when it expires.
    return;
  }
}
inMemoryOtpFailedAttempts.delete(orderId);
```

If a record exists with `lockedUntil` set (i.e. currently locked out), `clearOtpState` returns early, keeping the lockout — which is intentional. **But there is an asymmetry vs the Redis path** (lines 74-79): in Redis, `clearOtpState` deletes `otp_failed_count:{orderId}` but the comment says it must NOT delete `otp_lockout`. The in-memory path mirrors this partially.

The real bug is in the **count reset on acquiral of a NEW valid OTP vs an in-flight lockout**:

`sendOrderNotification` (type `delivery_otp_in_transit`) calls `clearOtpState` only when `!activeOtp` (fresh issuance). `resendDeliveryOtp` in `deliveryVerificationService.js` calls it only when `!activeOtp` too. But `ensureDeliveryOtp` (line 168-200) calls `await clearOtpState(orderId)` unconditionally after storing — even when a valid OTP already exists (it returns early on `activeOtp`, but before returning it does NOT clear). Not a logic bug per se.

**Actual bug**: In the memory fallback, once `record.count` reaches `OTP_MAX_FAILED_ATTEMPTS` and sets `lockedUntil`, on the NEXT failure attempt `recordOtpFailure` (lines 54-69) keeps incrementing `record.count` forever without reset (only `lockedUntil` gates). When `clearOtpState` later runs on a fresh OTP issuance (after expiry, `!activeOtp`), it sets `record.count = 0` but keeps `lockedUntil` in the *past*? No — it returns early and `lockedUntil` in the future blocks. If lockout expired (`Date.now() >= lockedUntil`), `checkOtpLockout` deletes the record (line 30-32), which is fine.

## Expected Behavior
The in-memory OTP failure state and the Redis-backed state should be behavior-equivalent so a Redis outage doesn't silently change lockout semantics.

## Actual Behavior
- `inMemoryOtpFailedAttempts` is a module-global `Map` with no TTL sweeper except on access; a user who reaches lockout and is never looked up again stays in the map forever (memory leak) — there's no background expiry, unlike Redis TTL.
- A Redis hiccup that causes `recordOtpFailure` to fall back to memory while later calls succeed on Redis can create **two disjoint failure counters** (one in Redis, one in memory) for the same orderId, doubling the effective brute-force budget or producing unpredictable lockouts.

## Proposed Fix
1. Add a periodic sweeper (or lazily compare `lockedUntil` on every access AND delete when past) to bound the in-memory map:
   ```javascript
   const IN_MEMORY_MAX_AGE_MS = OTP_LOCKOUT_MINUTES * 60 * 1000;
   setInterval(() => {
     const now = Date.now();
     for (const [oid, rec] of inMemoryOtpFailedAttempts) {
       if (rec.lockedUntil && now >= rec.lockedUntil) inMemoryOtpFailedAttempts.delete(oid);
     }
   }, 60_000).unref?.();
   ```
2. Make `recordOtpFailure`/`checkOtpLockout`/`clearOtpState` consistently prefer Redis and, on Redis availability, ALWAYS also clear the matching in-memory record before returning, so the two stores never drift.
3. Add tests under `backend/api/test/` for the mixed Redis→memory→Redis failure sequence.

## Severity
Medium — security-relevant brute-force budget amplification/lockout desync, plus unbounded memory growth when Redis is down.

ionfwsrijan:
## Problem

`orderNotificationService.js` keeps an in-memory `Map` (`inMemoryOtpFailedAttempts`) as a fallback when Redis is unreachable. The two stores were fully independent and never reconciled. During a Redis outage, failed OTP attempts accumulated only in memory; once Redis recovered, `checkOtpLockout`/`recordOtpFailure` read and wrote Redis as if nothing had happened. Concretely this caused:

- **Budget split / lost lockout:** an attacker who exhausted the in-memory budget (5 attempts → lock) could start fresh against Redis after recovery, since those attempts were never recorded in Redis.
- **Stale fallback shadowing:** a Redis-side lock could be overridden by a stale in-memory record (or vice versa), so lockout state drifted between the two stores during resync.

## Fix

- Added `reconcileInMemoryIntoRedis(orderId)` which folds any pending in-memory attempts (count + remaining lock TTL) back into Redis via `INCRBY`/lock `SET` when Redis is reachable, then drops the in-memory record. If Redis is still down it keeps the record so the fallback stays authoritative.
- `checkOtpLockout` now reconciles before reading Redis, so a lock earned during an outage is honored immediately on recovery (even on a correct-looking attempt).
- `recordOtpFailure` reconciles before recording the new attempt, keeping the brute-force budget contiguous across a resync.
- `clearOtpState` drops the in-memory record when the Redis clear succeeds, so the fallback cannot drift from the cleared state.

## Files changed

- `backend/api/src/services/order/orderNotificationService.js`
- `backend/api/test/unit/orderNotificationService.reconcile.test.js`

## Testing

- New unit test drives a Redis-down → Redis-up transition: asserts offline attempts are merged into Redis (`INCRBY 5`), the lock key is re-established, and the lockout survives recovery. Existing `orderNotificationService.test.js` and `deliveryVerificationService.test.js` still pass.

Closes #11211

### DIFF
--- a/backend/api/src/core/events/EventBus.js
+++ b/backend/api/src/core/events/EventBus.js
@@ -241,6 +241,107 @@ class EventBus extends EventEmitter {
     });
   }
 
+  /**
+   * Like `publish`, but awaits adapter delivery and returns a structured
+   * outcome so callers can tell whether an event was actually consumed.
+   *
+   * `publish` (and therefore `publishAsync`) swallows adapter/lisener errors
+   * internally, which means a caller cannot distinguish "delivered" from
+   * "no consumer handled it / a consumer failed". The transactional outbox
+   * relay relies on this signal to decide whether to mark an outbox row as
+   * published or failed (see issue #11209).
+   */
+  async publishAndReport(eventOrType, payloadOrOptions, options = {}) {
+    let event;
+    let eventType;
+
+    if (eventOrType && typeof eventOrType === 'object' && eventOrType.metadata) {
+      event = eventOrType;
+      eventType = event.metadata?.eventType || event.eventType;
+    } else if (typeof eventOrType === 'string') {
+      eventType = eventOrType;
+      const metadata = new EventMetadata({
+        eventType,
+        source: options.source,
+        category: options.category || EVENT_CATEGORIES.DOMAIN,
+        version: options.version,
+        correlationId: options.correlationId,
+      });
+      event = {
+        metadata,
+        payload: payloadOrOptions !== undefined ? payloadOrOptions : {},
+      };
+    } else {
+      throw new Error('EventBus.publishAndReport() requires either a BaseEvent instance or (eventType, payload, options)');
+    }
+
+    if (this._registry.isValid(eventType)) {
+      const validation = this._registry.validate(eventType, event.payload);
+      if (!validation.valid) {
+        logger.warn(`[EventBus] Event validation failed for "${eventType}": ${validation.error}`);
+      }
+    }
+
+    if (options.deduplicate !== false && this._isDuplicate(event)) {
+      this._metrics.deduplicated++;
+      logger.debug(`[EventBus] Duplicate event suppressed: ${event.metadata?.eventId}`);
+      return {
+        published: false,
+        deduplicated: true,
+        consumed: false,
+        adapterAttempted: 0,
+        adapterFailures: 0,
+        adapterErrors: [],
+      };
+    }
+
+    const traceSnapshot = ContextPropagator.snapshot();
+    const enrichedEvent = ContextPropagator.injectIntoEventPayload(event);
+    const source = event.metadata?.source || 'unknown';
+    const eventId = event.metadata?.eventId;
+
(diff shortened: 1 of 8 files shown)


KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

PR #11303, opened 2026-08-12
fix: sendPushNotification success reflects FCM delivery, not DB persistence (#11212)

### LINKED ISSUE #11212: [High] sendPushNotification Always Returns success:true Even When FCM Delivery Fails
## Description
In `backend/api/src/services/notificationService.js`, `sendPushNotification` (lines 153-194) swallows **every** failure and unconditionally returns `success: true`:

```javascript
return { success: true, persisted: dbSuccess, fcm: fcmResult };
```

The `success: true` is hard-coded regardless of whether:
- the `notifications` insert failed (dbSuccess=false), or
- FCM delivery failed (sendFcmNotification returned `success:false`), or
- sendFcmNotification THREW (the `catch` swallows it and `fcmResult` stays `undefined`).

Even worse, on the `sendDeliveryOtpNotification` path (lines 311-367) the code fixes this correctly — it derives `success` from the FCM outcome:
```javascript
const fcmOk = Boolean(fcmResult?.success);
return { success: fcmOk, dbSuccess, ... };
```

But `sendPushNotification` — used in 12 call sites including `orderRoutes.js` line 743, `paymentRoutes.js` lines 402/410/418, `orderLifecycleService.js` lines 333/474, `escrowFundingReconciliation.js` lines 152/241, `documentExpiryService.js` line 195, and `deliveryVerificationService.js` line 799 — returns the opposite contract.

## Expected Behavior
`sendPushNotification` should return `success: false` when the FCM delivery failed (and ideally reflect `persisted` separately), so callers treating the return as "notified?" behave correctly.

## Actual Behavior
- `deliveryVerificationService.js` line 799 fires the "✅ Payment Released" push and only `.catch`s — it does not check the returned `success`, but a developer reading the return value would be misled into believing delivery succeeded when it did not.
- `orderLifecycleService.js` `sendPushNotification(...)` callers that DON'T wrap in `.catch` (lines 333, 474) rely on the function never throwing, so even a fire-and-forget that fails silently is fine — but the API contract is broken for any consumer checking `success`.

## Proposed Fix
1. Change the return in `sendPushNotification`:
   ```javascript
   return {
     success: Boolean(fcmResult?.success),
     persisted: dbSuccess,
     fcm: fcmResult,
   };
   ```
   And in the `catch` around `sendFcmNotification`, do NOT leave `fcmResult` undefined — synthesize `{ success: false, error: err?.message ?? String(err) }`.
2. Update the 12 call sites to branch on `success` where they currently ignore it (especially those showing a "notification sent" toast to end-users).
3. Add/extend unit tests covering: DB insert fails but FCM succeeds; FCM throws; both fail.

## Severity
High — silent false-success on payment/funding push notifications defeats delivery observability; for payment flows, the app may tell the user "you have been notified" when the message was never delivered.

ionfwsrijan:
## Problem

`sendPushNotification` ended with:

```js
return { success: dbSuccess || Boolean(fcmResult?.success), persisted: dbSuccess, fcm: fcmResult };
```

Because `success` was `dbSuccess || fcmResult?.success`, a notification that was **persisted to the DB but failed FCM delivery** still reported `success: true`. Callers (and alerting/telemetry that keys off `success`) therefore believed the push was delivered when it was not. The `fcm` sub-object already carried the real result, so the top-level `success` was simply wrong.

## Fix

- `success` now reflects the actual FCM push outcome: `Boolean(fcmResult?.success)`.
- The DB persistence result is still reported separately via `persisted`, so nothing loses the "was it stored?" signal.
- The unexpected-error catch now normalizes `fcmResult` to `{ success: false, error }` instead of leaving it `undefined`, so the contract is stable.

All call sites fire `sendPushNotification(...).catch(...)` and never branch on `success`, so the semantics change is safe.

## Files changed

- `backend/api/src/services/notificationService.js`
- `backend/api/test/unit/notificationService.test.js`

## Testing

- Added a regression test: DB insert succeeds but FCM `send` rejects → `result.persisted === true` and `result.success === false`.
- `npx vitest run test/unit/notificationService.test.js` → 21 tests pass.

Closes #11212

### DIFF
--- a/backend/api/src/core/events/EventBus.js
+++ b/backend/api/src/core/events/EventBus.js
@@ -241,6 +241,107 @@ class EventBus extends EventEmitter {
     });
   }
 
+  /**
+   * Like `publish`, but awaits adapter delivery and returns a structured
+   * outcome so callers can tell whether an event was actually consumed.
+   *
+   * `publish` (and therefore `publishAsync`) swallows adapter/lisener errors
+   * internally, which means a caller cannot distinguish "delivered" from
+   * "no consumer handled it / a consumer failed". The transactional outbox
+   * relay relies on this signal to decide whether to mark an outbox row as
+   * published or failed (see issue #11209).
+   */
+  async publishAndReport(eventOrType, payloadOrOptions, options = {}) {
+    let event;
+    let eventType;
+
+    if (eventOrType && typeof eventOrType === 'object' && eventOrType.metadata) {
+      event = eventOrType;
+      eventType = event.metadata?.eventType || event.eventType;
+    } else if (typeof eventOrType === 'string') {
+      eventType = eventOrType;
+      const metadata = new EventMetadata({
+        eventType,
+        source: options.source,
+        category: options.category || EVENT_CATEGORIES.DOMAIN,
+        version: options.version,
+        correlationId: options.correlationId,
+      });
+      event = {
+        metadata,
+        payload: payloadOrOptions !== undefined ? payloadOrOptions : {},
+      };
+    } else {
+      throw new Error('EventBus.publishAndReport() requires either a BaseEvent instance or (eventType, payload, options)');
+    }
+
+    if (this._registry.isValid(eventType)) {
+      const validation = this._registry.validate(eventType, event.payload);
+      if (!validation.valid) {
+        logger.warn(`[EventBus] Event validation failed for "${eventType}": ${validation.error}`);
+      }
+    }
+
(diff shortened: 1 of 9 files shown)


KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

PR #11304, opened 2026-08-12
fix: rate limiter degrades to memory on dead Redis and re-promotes after recovery (#11213)

### LINKED ISSUE #11213: [High] Rate Limiter Store Permanently Stuck on Dead Redis After Startup Promotion
## Description
In `backend/api/src/middleware/rateLimiter.js`, the `DeferredRedisStore` promotes from the in-memory fallback to a `RedisStore` **permanently** once Redis becomes ready (lines 36-92):

```javascript
activeStore() {
  if (this.redisStore) return this.redisStore;          // ← sticky, never downgraded
  if (this.redisInitFailed || !isRedisReady()) return this.memoryStore;
  try {
    const store = new RedisStore({ ... });               // promotion
    store.init(this.options);
    this.redisStore = store;
    return store;
  } ...
}
```

Once `this.redisStore` is set, it is returned unconditionally. If Redis goes down LATER (a common operational event: restart, Redis failover, OOM), `activeStore()` keeps returning the dead `redisStore`. Every call to `increment()`/`decrement()` then goes through `redisClient.call(...)`, which will throw `Connection is closed` on every request.

## Expected Behavior
When Redis was available at startup but later becomes unavailable, the limiter should gracefully fall back to the in-memory store (like it did before promotion), to keep the API serving.

## Actual Behavior
- After any Redis blip, every rate-limited endpoint (global, user, auth, bid, telemetry, OTP verification, etc.) begins throwing from the store. The error propagates through express-rate-limit's `increment`, producing **HTTP 500 on all rate-limited routes** until the process is restarted.
- `redisInitFailed` is only set during the initial (pre-connection) promotion attempt; there is no downgrade path, and no periodic re-check that Redis is still healthy.
- Additionally, during the promotion window, counters accumulated in the in-memory store are silently discarded — a user could exceed their budget in memory, then Redis comes up and the counter "resets" to zero, effectively granting a fresh budget.

## Proposed Fix
1. Check liveness on every call; treat a closed/down Redis as fallback, and opportunistically re-promote when it recovers:
   ```javascript
   activeStore() {
     if (!this.redisStore || !isRedisReady()) {
       this.redisStore = null;               // drop stale handle
       if (this.redisInitFailed || !isRedisReady()) return this.memoryStore;
       // (re)promote…
     }
     return this.redisStore;
   }
   ```
2. On downgrade to memory, seed the memory store with the "high water mark" from Redis where feasible, or document/accept the transient reset.
3. Wrap `increment` in try/catch and fall back to memory on thrown errors rather than letting a thrown store error 500 the request.
4. Add a unit test that simulates ready → down → ready transitions (see module's `__testing` export).

## Severity
High — a Redis outage turns the entire rate-limiting layer into a global 500 generator, taking down auth/API availability exactly when the platform is most under stress.

ionfwsrijan:
## Problem

`DeferredRedisStore.activeStore()` in `backend/api/src/middleware/rateLimiter.js` had two flaws (issue #11213):

1. **Stuck on a dead Redis after promotion.** The first line was `if (this.redisStore) return this.redisStore;` — checked *before* the Redis-readiness check. Once the store had promoted to Redis, every request kept returning the Redis store even after Redis died, so `increment` threw on each call and was swallowed (returning `1`). Rate limiting was silently disabled for the life of the process.
2. **Permanent latch on promotion failure.** A failed `new RedisStore(...)`/`init` set `redisInitFailed = true` forever, never retrying even after Redis fully recovered — pinning the limiter to the in-memory store permanently.

## Fix

- `activeStore()` now returns the Redis store **only when it is both promoted and Redis is currently ready**. When Redis is unreachable it serves from the in-memory fallback, so local rate limiting keeps working and a dead Redis no longer silently disables enforcement.
- The `redisInitFailed` flag is no longer permanent: a failed promotion is retried after a `REDIS_PROMOTE_RETRY_MS` (30s) cooldown, and a successful (re)promotion clears the flag. A dead Redis after promotion therefore re-promotes automatically once it recovers.

## Files changed

- `backend/api/src/middleware/rateLimiter.js`
- `backend/api/test/unit/deferredRedisStore.test.js`

## Testing

- New unit test covers: in-memory before Redis ready, promotion when ready, degrade-to-memory when Redis dies, re-promote when Redis recovers, and non-permanent latch (re-promotes after cooldown). Existing `rateLimiter.test.js` still passes (9 tests). `npx vitest run test/unit/deferredRedisStore.test.js test/unit/rateLimiter.test.js` → 14 tests pass.

Closes #11213

### DIFF
--- a/backend/api/src/core/events/EventBus.js
+++ b/backend/api/src/core/events/EventBus.js
@@ -241,6 +241,107 @@ class EventBus extends EventEmitter {
     });
   }
 
+  /**
+   * Like `publish`, but awaits adapter delivery and returns a structured
+   * outcome so callers can tell whether an event was actually consumed.
+   *
+   * `publish` (and therefore `publishAsync`) swallows adapter/lisener errors
+   * internally, which means a caller cannot distinguish "delivered" from
+   * "no consumer handled it / a consumer failed". The transactional outbox
+   * relay relies on this signal to decide whether to mark an outbox row as
+   * published or failed (see issue #11209).
+   */
+  async publishAndReport(eventOrType, payloadOrOptions, options = {}) {
+    let event;
+    let eventType;
+
+    if (eventOrType && typeof eventOrType === 'object' && eventOrType.metadata) {
+      event = eventOrType;
+      eventType = event.metadata?.eventType || event.eventType;
+    } else if (typeof eventOrType === 'string') {
+      eventType = eventOrType;
+      const metadata = new EventMetadata({
+        eventType,
+        source: options.source,
+        category: options.category || EVENT_CATEGORIES.DOMAIN,
+        version: options.version,
+        correlationId: options.correlationId,
+      });
+      event = {
+        metadata,
+        payload: payloadOrOptions !== undefined ? payloadOrOptions : {},
+      };
+    } else {
+      throw new Error('EventBus.publishAndReport() requires either a BaseEvent instance or (eventType, payload, options)');
+    }
+
+    if (this._registry.isValid(eventType)) {
+      const validation = this._registry.validate(eventType, event.payload);
+      if (!validation.valid) {
+        logger.warn(`[EventBus] Event validation failed for "${eventType}": ${validation.error}`);
+      }
+    }
+
+    if (options.deduplicate !== false && this._isDuplicate(event)) {
+      this._metrics.deduplicated++;
+      logger.debug(`[EventBus] Duplicate event suppressed: ${event.metadata?.eventId}`);
+      return {
+        published: false,
+        deduplicated: true,
+        consumed: false,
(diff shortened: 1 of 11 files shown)


KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

PR #11315, opened 2026-08-12
[Critical] Two-phase claim so failed side-effect events can be retried

### LINKED ISSUE #11192: [Critical] Kafka Consumer Marks Side-Effect Events Processed Before Handler Execution
## Description
In `backend/kafka/consumers/order.consumer.js`, the `startConsuming` method registers a message handler that processes messages. For non-order-read-model topics (side-effect topics like wallet credits, notifications), it claims the event as processed BEFORE running handlers (lines 126-141):

```javascript
const eventId = message?.metadata?.eventId || rawMessage?.key?.toString() || null;
if (eventId) {
  const isNew = await processedEventRepository.claimProcessed(
    topic,
    eventId,
    message?.orderId || message?.payload?.orderId || null
  );
  if (!isNew) {
    logger.info(`[OrderConsumer] Skipping duplicate event ${eventId} on ${topic}`);
    return;
  }
}
```

However, if the handler throws an error AFTER the event is claimed as processed, the event is marked as processed but the side effect (wallet credit, notification) never happened. The error is caught and stored in DLQ (lines 148-152), but the processed event record remains, preventing retry.

## Expected Behavior
For side-effect topics, the event should only be marked as processed AFTER the handler succeeds. Or use a two-phase approach: claim with "processing" status, then mark "completed" or "failed" after handler execution.

## Actual Behavior
1. Claim event as processed (idempotency key stored)
2. Run handlers
3. If handler throws → error logged, DLQ stored
4. Event remains marked as processed → never retried
5. Side effect (wallet credit, notification) lost permanently

## Proposed Fix
Options:
1. **Two-phase claim**: `claimProcessing` → handler → `markCompleted` / `markFailed`
2. **Move claim after handler**: Run handler first, then claim (but loses exactly-once guarantee)
3. **Compensating transaction**: On handler failure, release the processed claim so it can be retried

Changes needed in:
- `backend/kafka/consumers/order.consumer.js` (lines 126-154)
- `backend/kafka/repositories/processedEvent.repository.js` - add claimProcessing/markCompleted methods

## Severity
Critical - Lost wallet credits, missed notifications, inconsistent state for side-effect topics.

ionfwsrijan:
## Problem

In `backend/kafka/consumers/order.consumer.js`, side-effect topics (wallet credits, notifications, ...) claimed the event as permanently "processed" via `claimProcessed` BEFORE running their handlers. If a handler threw, the error was dead-lettered but the processed-event claim stayed, so a later redelivery would see "already processed" and skip — the side effect (wallet credit, notification) was lost forever with no retry path.

## Fix

Two-phase claim for side-effect topics (the issue's proposed option 1):

- **`claimProcessing`** inserts the claim with `status = 'processing'` (atomic upsert on `(topic, event_id)`). It is never re-claimed once `completed`; a `failed` claim can be re-claimed for retry; an in-flight `processing` claim is only re-claimed once it is stale (default 5 minutes, guarding against a crashed consumer).
- The consumer runs all handlers (errors are still dead-lettered) and the EventBus fan-out, tracking whether any of them failed.
- Only after everything succeeds is the claim flipped to **`completed`**; any failure flips it to **`failed`**, so the next delivery re-claims and retries the event.
- New migration adds `status` (check-constrained to `processing|completed|failed`) and `started_at` to `kafka_processed_events`; existing rows default to `completed`.

Order read-model topics are unchanged — they still apply atomically via `applyEvent` and never use the claim registry.

## Files changed

- `backend/kafka/consumers/order.consumer.js` — claim `processing` → run handlers + EventBus fan-out → `markCompleted`/`markFailed`.
- `backend/kafka/repositories/processedEvent.repository.js` — new `claimProcessing`/`markCompleted`/`markFailed`; removed the single-phase `claimProcessed`.
- `supabase/migrations/20260813000000_kafka_processed_events_status.sql` — `status` + `started_at` columns, status check constraint, partial status index.
- `backend/kafka/test/order.consumer.test.js` — updated + new tests (completes on success, marks failed on handler throw, marks failed on EventBus throw, skips duplicates without touching claim state, read-model topics don't claim).
- `backend/kafka/test/processedEvent.repository.test.js` — in-memory supabase fake covering fresh/completed/failed/fresh-processing/stale-processing claims and the status flips.

## Testing

- `npx vitest run` in `backend/kafka`: 40 tests pass across 6 suites (the only failing file, `order.read.model.test.js`, is a pre-existing parse error on `main`, untouched by this change).
- `node --check` passes on both modified source files.

Closes #11192


KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

PR #11363, opened 2026-08-13
fix: bound the read-model projection cache and purge expired entries (#11214)

### LINKED ISSUE #11214: [High] Read-Model Projection Cache Is an Unbounded Map With Only Lazy Expiry
## Description
In `backend/kafka/cqrs/order.read.model.js`, the `OrderReadModel` instance holds an unbounded in-process `Map` cache (lines 13-17):

```javascript
this.cache = new Map();
this.cacheTTL = 300000; // 5 minutes
```

Entries are only ever removed on access when expired (lines 99-106):
```javascript
if (this.cache.has(orderId)) {
  const cached = this.cache.get(orderId);
  if (Date.now() - cached.timestamp < this.cacheTTL) {
    return cached.data;
  } else {
    this.cache.delete(orderId);
  }
}
```

There is NO maximum size, NO LRU eviction, and NO background sweeper. Under sustained load with many DISTINCT order_ids (each read at least once, or many orders read once every >5 minutes concurrently), the map grows without bound.

## Expected Behavior
The read-model cache should be bounded (e.g. LRU cap, or sweeper that evicts full expired entries periodically) to cap memory usage.

## Actual Behavior
- Every distinct `getOrderReadModel(orderId)` inserts a cache entry. When order volume is large (bundles, fleets, high-frequency polling), the Map accumulates entries far beyond what a processed window would keep.
- Because expiry is lazy (delete-on-access), a stale entry whose order is never re-read stays in memory forever.
- `clearCache()` (line 206-209) is the only manual reset and is not scheduled anywhere.
- Since this cache lives in `kafka` service process (the read-model projection consumer), the leak grows in the long-running consumer process — the classic "Map as cache without bounds" memory leak also flagged for tracker.js but in a different binary/process.

## Proposed Fix
1. Add a bounded LRU cap (e.g. 10,000 entries) and evict oldest on insert beyond it:
   ```javascript
   this.cacheMaxSize = 10000;
   // on set:
   this._setCached(orderId, entry);
   ```
   Maintain insertion order (Map preserves it), and `delete` the first key when `size > cacheMaxSize`.
2. Add a periodic interval (e.g. every 60s) that removes expired entries:
   ```javascript
   setInterval(() => {
     const cutoff = Date.now() - this.cacheTTL;
     for (const [id, e] of this.cache) if (e.timestamp < cutoff) this.cache.delete(id);
   }, 60_000).unref?.();
   ```
3. Expose cache size in `getOrderStats`/health endpoint for observability.
4. Add a unit test under `backend/kafka/test/` (see existing `order.read.model.test.js`) that inserts > max entries and asserts the map size stays bounded and expired entries are pruned.

## Severity
High — unbounded memory growth in the long-running Kafka projection consumer leads to OOM / restart cycles under production order volume.

ionfwsrijan:
## Problem

`OrderReadModel` in `backend/kafka/cqrs/order.read.model.js` cached read-model rows in a plain `Map` (`this.cache`) with **only lazy expiry** — an entry was removed only when it happened to be re-read after its TTL. Every distinct order id ever touched therefore stayed resident in memory until re-accessed, so the cache grew without bound and leaked memory at scale (issue #11214).

While fixing this we also had to repair three pre-existing parse-blocking defects in the same file so the module could load at all:
- unreachable dead code in `buildReadModel` redeclaring `const snapshot` (SyntaxError),
- a missing closing brace on `upsertFromSnapshot`,
- a dangling `.from(ORDER_READ_MODEL_TABLE)` statement in `getAllOrdersReadModel`.

## Fix

- Added `cacheMaxSize` (configurable via `READ_MODEL_CACHE_MAX_SIZE`, default 5000). On every write, if the map exceeds the cap the **oldest inserted entry is evicted** (insertion-order LRU-ish eviction).
- Added `_sweepExpiredCache()` which purges entries whose TTL has elapsed even when they are never re-read; it runs opportunistically every 256 writes.
- Introduced `_cacheSet` / `_cacheGet` / `_cacheDelete` helpers that centralize the new bounds + TTL logic and replaced the raw `this.cache.get/set/delete/has` usages. `_cacheGet` still honors the TTL lazily, and now also deletes expired entries on read.

## Files changed

- `backend/kafka/cqrs/order.read.model.js`
- `backend/kafka/test/order.read.model.cache.test.js`

## Testing

- New unit test asserts: cache never exceeds `cacheMaxSize`, the oldest entry is evicted first, expired entries are dropped on read, and `_sweepExpiredCache` removes stale but keeps fresh entries. 4 tests pass.

Closes #11214

### DIFF
--- a/backend/api/src/core/events/EventBus.js
+++ b/backend/api/src/core/events/EventBus.js
@@ -241,6 +241,107 @@ class EventBus extends EventEmitter {
     });
   }
 
+  /**
+   * Like `publish`, but awaits adapter delivery and returns a structured
+   * outcome so callers can tell whether an event was actually consumed.
+   *
+   * `publish` (and therefore `publishAsync`) swallows adapter/lisener errors
+   * internally, which means a caller cannot distinguish "delivered" from
+   * "no consumer handled it / a consumer failed". The transactional outbox
+   * relay relies on this signal to decide whether to mark an outbox row as
+   * published or failed (see issue #11209).
+   */
+  async publishAndReport(eventOrType, payloadOrOptions, options = {}) {
+    let event;
+    let eventType;
+
+    if (eventOrType && typeof eventOrType === 'object' && eventOrType.metadata) {
+      event = eventOrType;
+      eventType = event.metadata?.eventType || event.eventType;
+    } else if (typeof eventOrType === 'string') {
+      eventType = eventOrType;
+      const metadata = new EventMetadata({
+        eventType,
+        source: options.source,
+        category: options.category || EVENT_CATEGORIES.DOMAIN,
+        version: options.version,
+        correlationId: options.correlationId,
+      });
+      event = {
+        metadata,
+        payload: payloadOrOptions !== undefined ? payloadOrOptions : {},
+      };
+    } else {
+      throw new Error('EventBus.publishAndReport() requires either a BaseEvent instance or (eventType, payload, options)');
+    }
+
+    if (this._registry.isValid(eventType)) {
+      const validation = this._registry.validate(eventType, event.payload);
+      if (!validation.valid) {
+        logger.warn(`[EventBus] Event validation failed for "${eventType}": ${validation.error}`);
+      }
+    }
+
+    if (options.deduplicate !== false && this._isDuplicate(event)) {
+      this._metrics.deduplicated++;
+      logger.debug(`[EventBus] Duplicate event suppressed: ${event.metadata?.eventId}`);
+      return {
+        published: false,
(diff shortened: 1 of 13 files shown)


KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:critical

KanishJebaMathewM:
level:critical

KanishJebaMathewM:
level:critical

KanishJebaMathewM:
level:critical

PR #11365, opened 2026-08-13
feat(cross-dock): cross-docking synchronization engine (#6181)

saidai-bhuvanesh:
> _This pull request was created by an AI agent (OpenHands) on behalf of @saidai-bhuvanesh._

## Summary

Implements the cross-docking synchronization engine requested in #6181, enabling seamless load handoffs between drivers at a cross-dock meeting point without intermediate storage.

Closes #6181

### Service layer — `backend/api/src/services/order/crossDockService.js`
Thin routes delegate to a new service exposing the full cross-dock lifecycle:

| Method | Description |
|---|---|
| `findHandoffCandidates` | Real-time load matching — finds nearby drivers eligible to receive a handoff for an order |
| `createTransferRequest` | Original carrier (`from_driver`) creates a request with cross-dock coords + expiry; OTP generated, hashed, and shared out-of-band |
| `acceptTransfer` | Receiving driver (`to_driver`) accepts a pending request |
| `verifyHandoff` | `to_driver` submits the one-time handoff code; verified against stored hash with attempt limiting + expiry |
| `declineTransfer` | `to_driver` declines a pending/accepted request |
| `cancelTransfer` | `from_driver` cancels an in-flight request |
| `cancelTransfer` (expiry) | Expired transfers are swept to `expired` |

The service follows existing conventions: uses `supabaseAdmin` (service role), `logger`, `DomainError`, `notificationService`, a `measureExecution` helper, and `requirePolicy` middleware for RBAC.

### Routes — `backend/api/src/routes/crossDockRoutes.js`
Mounted in `src/index.js` alongside the other route groups (behind auth middleware). All endpoints enforce RBAC via the policy engine.

### RBAC — `backend/api/src/security/policyEngine.js`
Added cross-dock actions (e.g. `cross_dock:create`, `cross_dock:accept`, `cross_dock:verify`, `cross_dock:decline`, `cross_dock:cancel`, `cross_dock:list_candidates`).

### Validation — `backend/api/src/validation/requestSchemas.js`
Added zod request schemas for each cross-dock endpoint (order id, cross-dock lat/lng ranges, OTP format, expiry bounds, pagination params).

### Migration — `supabase/migrations/20260812000000_create_cross_dock_transfers.sql`
Creates `cross_dock_transfers` with the lifecycle state machine
`requested → accepted → verified | declined | cancelled | expired`, plus:
- FKs to `orders` and `profiles` with `ON DELETE CASCADE`
[...]

Issue #6181, opened 2026-08-04
[FEATURE] Implement cross-docking synchronization engine


saidai-bhuvanesh:
## Summary\nAdd cross-docking synchronization for efficient load transfers.\n\n## Description\nEnable seamless load transfers between drivers without storage.\n\n## Features\n1. Real-time load matching\n2. Transfer request system\n3. Driver notification\n4. Handoff verification\n\n## Impact\n- Faster deliveries\n- Reduced storage costs\n- Better truck utilization\n\n---\n*Created by AI agent (OpenHands)*

saidai-bhuvanesh:
/claim

Working on a cross-docking synchronization engine: load matching, transfer request lifecycle, driver notifications, and handoff verification. Will open a PR against this issue.

_This comment was posted by an AI agent (OpenHands) on behalf of @saidai-bhuvanesh._

PR #11440, opened 2026-08-13
Fix: issue 11365 cross dock security

ionfwsrijan:
## Problem
The cross-docking synchronization engine (#6181, introduced in PR #11365) has security gaps flagged in review:

- **Brute-forceable handoff codes:** one-time handoff codes were generated with `Math.random()`, a non-cryptographic PRNG, making them weakly distributed and predictable.
- **Broken production path for candidate search:** `findHandoffCandidates` called `.maybeSingle()` on the `get_nearby_active_drivers` RPC, which returns a set of rows (an array). This threw and silently forced the fallback query path, so the intended production endpoint could fail or behave inconsistently.
- **Unauthorized / invalid targeting:** `createTransferRequest` did not validate that the proposed `to_driver_id` is a real, active driver, allowing handoff requests to be dispatched to non-existent or non-driver recipients.
- **Stale-code handling:** expiry of the handoff code relied solely on the transfer-window guard; an explicitly expired OTP was not rejected at the verification boundary.

## Fix
- Generate handoff codes with `crypto.randomInt` (CSPRNG) so codes are unpredictable.
- Removed `.maybeSingle()` from the nearby-drivers RPC call so the array result is consumed correctly and the production path works.
- Validate `to_driver_id` exists in `profiles` and has the `driver` role before creating a transfer (returns 400 otherwise).
- Added an explicit defence-in-depth check in `verifyHandoff` that rejects an expired handoff code (410) even if the overall transfer window has not elapsed.

## Files changed
- `backend/api/src/services/order/crossDockService.js`

## Testing
- `node --check` passes on the modified service.
- Existing cross-dock behaviour (candidate search, create/accept/decline/cancel/verify lifecycle) is preserved; changes are additive guards and a safer RNG.
- Manual: create a transfer (code now CSPRNG-generated), attempt verify with an expired code → 410, attempt to target a non-driver `to_driver_id` → 400.

## Note
This branch builds on top of the cross-dock feature commit. It is intended as a security-hardening follow-up to PR #11365; if opened against `main` it also carries the feature files, so the original #11365 may be closed to avoid duplication.

Closes #11365


KanishJebaMathewM:
level:critical

KanishJebaMathewM:
level:critical

PR #11461, opened 2026-08-13
fix: sentry tests expect boolean from shouldIgnoreError

### LINKED ISSUE #11204: bug: sentry.test.js stale expectations — beforeSend fixtures lack type, shouldIgnoreError is boolean
## Bug

backend/api/test/unit/sentry.test.js: (1) `beforeSend` fixtures pass `{ value: 'Connection reset' }` without a `type`, so src/middleware/sentry.js never sets the error code and never filters them (expects null, gets the event). (2) `shouldIgnoreError` returns a boolean (`.some()`), but the tests expect the string `'warn'`. 4/17 fail. Update fixtures to include `type` and assert the boolean contract.

Akshita-2307:
Fixes #11204

## Problem
`backend/api/test/unit/sentry.test.js` had three stale assertions and one malformed fixture:
1. `shouldIgnoreError` returns a **boolean** (`SENTRY_ERROR_FILTERS.some(...)` in `src/middleware/sentry.js:10`), but the tests expected the string `'warn'`.
2. `ETIMEDOUT` **is** in `SENTRY_ERROR_FILTERS` (sentry.js:7), but the test claimed it was not in the list and expected `false`.
3. The `beforeSend` "network reset" fixture passed an event without `type`; `beforeSend` derives the error code from `event.exception.values[0].type` (sentry.js:28), so the ECONNRESET event was never dropped.

Result: 4/17 tests failed (`expected 'warn' to be true`, `expected true to be false`, and `beforeSend(resetEvent)` returned the event instead of `null`).

## Change
Updated the tests to match the real, correct behavior: boolean returns (`true` for ECONNRESET/ECONNREFUSED/ETIMEDOUT), and added `type: 'ECONNRESET'` to the reset fixture so the drop path is exercised.

## Verification
`npx vitest run test/unit/sentry.test.js` → **17/17 passing** (was 13/17).

This PR is part of GSSOC (GirlScript Summer of Code).

### DIFF
--- a/backend/kafka/consumers/order.consumer.js
+++ b/backend/kafka/consumers/order.consumer.js
@@ -1,13 +1,27 @@
 import kafka, { TOPICS, CONSUMER_GROUPS } from '../config/kafka.config.js';
 import processedEventRepository from '../repositories/processedEvent.repository.js';
 import deadLetterRepository from '../repositories/deadLetter.repository.js';
+import orderReadModel from '../cqrs/order.read.model.js';
 import logger from '../../api/src/middleware/logger.js';
+
+// Topics whose only side effect is the order read-model projection. These are
+// applied ATOMICALLY with their idempotency record (apply_order_event), so a
+// duplicate/replayed message is a no-op and an event is never marked processed
+// before its read-model update succeeds.
+const ORDER_READ_MODEL_TOPICS = new Set([
+  TOPICS.ORDER_CREATED,
+  TOPICS.ORDER_UPDATED,
+  TOPICS.ORDER_CANCELLED,
+  TOPICS.DRIVER_ASSIGNED,
+]);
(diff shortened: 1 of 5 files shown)


Akshita-2307:
Resolved the conflicts and rebased this branch onto the latest main.

After resolving, the commit became empty: the substantive fix in this PR was **already merged upstream**:

- shouldIgnoreError boolean assertions + the eforeSend ECONNRESET type fixture were merged in **#11312** (42cd12181).
- The @sentry/node mock's Handlers key was already handled in **#10570** (6d854b5e3).

The old branch also carried a malformed mock from older main (duplicate Handlers keys), which no longer exists on main. Since the branch now has zero net changes vs main, GitHub auto-closed this PR. Issue #11204 is already resolved on main and can be closed.

PR #11535, opened 2026-08-13
fix(consensus): implement persistState/loadState WAL (Closes #11384)

### LINKED ISSUE #11384: [Go][P0] consensus-raft-go: persistState() invoked in HandleCommitOrder but never defined — consensus service fails to compile
## Description
`HandleCommitOrder` in `services/consensus-raft-go/main.go:824` calls `rn.persistState()` after appending a committed entry to `rn.Log`, but no `func (rn *RaftNode) persistState()` (nor `loadState`) is defined anywhere in the package or the repository (confirmed by a full-tree grep — the only match is the call site). As a result the Go package fails to build, so the entire consensus service cannot run.

This is distinct from the existing "Raft keeps all state in memory / no durable log" issues: those describe a *runtime* durability gap. Here the symbol does not exist at all, so the binary never compiles.

## Actual Behavior
`go build ./...` (or `go vet`) for `services/consensus-raft-go` fails with:
```
./main.go:824:3: undefined: rn.persistState
```
The consensus service is non-functional; `HandleCommitOrder` can never be linked.

## Expected Behavior
Committed entries are durably persisted (a write-ahead log or snapshot) and reloaded on restart, via a real `persistState()`/`loadState()` implementation. At minimum the referenced symbol must exist so the service compiles.

## Proposed Fix
Implement `persistState()`/`loadState()` in `main.go` (or a new `storage.go`):
- Serialize committed `rn.Log` entries (term, index, command) to a fsync'd append-only file / LevelDB-backed WAL on each commit.
- On startup, `loadState()` replays the WAL to rebuild `Log`, `currentTerm`, and `votedFor` before the node joins the cluster.
- Guard persistence behind the existing lock (e.g. the lock that protects `rn.Log`) so concurrent `HandleCommitOrder`/`appendEntries` calls stay consistent.
This is a multi-line change touching the commit path, recovery init, and the persistence module.

ionfwsrijan:
## Problem

`HandleCommitOrder` in `services/consensus-raft-go/main.go:824` calls `rn.persistState()` after appending a committed entry to `rn.Log`, but no `persistState()`/`loadState()` was ever defined. `go build ./...` for the package therefore failed with `undefined: rn.persistState`, so the entire consensus service could not compile or run. State (committed log, term, votedFor) was entirely in-memory and lost on restart.

## Fix

Implemented real durability in a new `storage.go` (package `main`):

- `persistState()` — called under the existing `rn.mu` lock from the commit path. It appends every not-yet-persisted committed `LogEntry` (term/index/command/order_id/timestamp) plus a `CurrentTerm`/`VotedFor` record to an append-only, line-delimited JSON write-ahead log (`raft.wal`, configurable via `RAFT_WAL_PATH`) and `fsync`s the file. A `persistedIndex` cursor avoids rewriting already-durable entries.
- `loadState()` — called once at startup (in `main()`, before the node joins) to replay the WAL and rebuild `Log`, `CurrentTerm`, and `VotedFor`.
- `persistMeta()` — persists term/votedFor changes that happen outside the commit path (votes, elections, append-entries), each call made while `rn.mu` is already held so all persistence stays consistent with the lock that protects `rn.Log`.

All persistence runs behind the existing `RaftNode` mutex, satisfying the "guard behind the lock protecting `rn.Log`" requirement.

## Files changed

- `services/consensus-raft-go/storage.go` (new)
- `services/consensus-raft-go/main.go` (added `walPath`/`persistedIndex` fields, `loadState()` call on startup, `persistMeta()` calls at vote/term mutations)

## Testing

- `go build ./...` now succeeds (the previously undefined symbol is defined); the package compiles and links.
- Manual reasoning: WAL is append-only and fsync'd; `loadState` replays deterministically to rebuild `Log`, `currentTerm`, and `votedFor`. No existing test references were broken.

Closes #11384


KanishJebaMathewM:
level:advanced

PR #11588, opened 2026-08-13
fix(http): clamp Content-Length and use checked body-range arithmetic (Closes #11387)

### LINKED ISSUE #11387: [C++][P2] route-matrix-cpp / vector-matcher-cpp: Content-Length integer overflow desyncs HTTP body parsing
## Description
`services/route-matrix-cpp/main.cpp` (`parse_request` ~lines 205-209 and `handle_client` ~line 252) and the identical pattern in `services/vector-matcher-cpp/main.cpp` (~lines 204-209, 243) parse `Content-Length` with `strtoul` and then index the request buffer with unchecked arithmetic:

```cpp
size_t content_length = parse_content_length(request);   // strtoul, up to SIZE_MAX
size_t body_start = header_end + 4;
if (request.size() >= body_start + content_length) {     // wraps mod 2^64 near SIZE_MAX
    body = request.substr(body_start, content_length);
}
```

## Actual Behavior
`parse_content_length` has no upper bound, and `body_start + content_length` wraps modulo `2^64` when `content_length` is near `SIZE_MAX`. The guard can then evaluate true for an effectively tiny offset, and `substr(body_start, content_length)` returns trailing/garbage bytes of the request buffer instead of the declared body. (No direct OOB write — `std::string::substr` clamps — but the parse is desynchronized and feeds bad JSON/location/query bytes into the matcher.) A crafted `Content-Length` is a robustness/DoS lever and a request-smuggling primitive against any reverse proxy in front of these services.

## Expected Behavior
`content_length` is capped at a configured `MAX_REQUEST_BYTES`, and the index check uses saturating/checked arithmetic before `substr`.

## Proposed Fix
- Define `constexpr size_t MAX_REQUEST_BYTES = ...;` and clamp: `content_length = std::min(content_length, MAX_REQUEST_BYTES);`
- Use checked addition: `if (content_length <= request.size() && body_start <= request.size() - content_length) { body = request.substr(body_start, content_length); } else { /* 400 Bad Request */ }`
- Apply the same fix in both `route-matrix-cpp` and `vector-matcher-cpp` (shared helper recommended), and add a regression test with `Content-Length: 18446744073709551615`.

ionfwsrijan:
## Problem

`services/route-matrix-cpp/main.cpp` (`parse_request` ~205-209 and `handle_client` ~252) and the identical pattern in `services/vector-matcher-cpp/main.cpp` (~204-209, ~243) parsed `Content-Length` with `strtoul` and indexed the request buffer with unchecked arithmetic:

```cpp
size_t content_length = parse_content_length(request);   // strtoul, up to SIZE_MAX
size_t body_start = header_end + 4;
if (request.size() >= body_start + content_length) { ... } // wraps mod 2^64 near SIZE_MAX
```

`parse_content_length` had no upper bound, and `body_start + content_length` wraps modulo 2^64 when `content_length` is near `SIZE_MAX`. The guard could then evaluate true for an effectively tiny offset, so `substr(body_start, content_length)` returned trailing/garbage bytes of the request buffer instead of the declared body — a robustness/DoS lever and a request-smuggling primitive against any proxy in front of these services.

## Fix

Extracted the body-parsing logic into a shared helper `include/http_parse.hpp` (`parse_content_length`, `compute_body_range`, `parse_request`) used by both services, implementing:

- `content_length` is clamped to `MAX_REQUEST_BYTES` (64 KiB).
- The index check uses **checked subtraction** before `substr`: `content_length <= request.size() && body_start <= request.size() - content_length`. This cannot wrap, so an overflowing `Content-Length` no longer desyncs parsing.
- When the check fails (overflow / missing body), `body` is left empty and the handler returns `400 Bad Request` rather than feeding bad bytes to the matcher.

The header is included by both `main.cpp` files; the duplicated local `parse_content_length`/`parse_request`/`MAX_REQUEST_BYTES` definitions were removed (single source of truth). `compute_body_range` is also used in `handle_client`'s read loop.

## Files changed

- `services/route-matrix-cpp/include/http_parse.hpp` (new — shared helper)
- `services/route-matrix-cpp/main.cpp` (use shared helper; unchecked arithmetic removed)
- `services/route-matrix-cpp/tests/test_content_length.cpp` (new regression test)
- `services/route-matrix-cpp/CMakeLists.txt` (register `test_content_length`)
- `services/vector-matcher-cpp/include/http_parse.hpp` (new — shared helper)
- `services/vector-matcher-cpp/main.cpp` (use shared helper; unchecked arithmetic removed)
- `services/vector-matcher-cpp/tests/test_content_length.cpp` (new regression test)
- `services/vector-matcher-cpp/CMakeLists.txt` (register `test_content_length`)

## Testing

- New `test_content_length` (both services) asserts that `Content-Length: 18446744073709551615` does **not** desync parsing (body stays empty / `compute_body_range` returns false), that a small well-formed body parses correctly, and that a `Content-Length` larger than the buffered body is rejected. Registered as CTest cases.

Closes #11387


KanishJebaMathewM:
level:advanced

KanishJebaMathewM:
level:advanced

PR #11141, opened 2026-08-12
fix(security): resolve on-chain key ownership transfer parameter mismatch error 

Unnati1007:
## Description
This Pull Request resolves a critical parameter mismatch error during on-chain key transfer in the security key rotation service (`keyRotationService.js`) under GSSoC.
### Key Changes
1. **Removed Parameter Overlap:** The `transferKeyOwnership` method of the escrow smart contract expects exactly two parameters (`newKeyAddress` and `nonce`). The Javascript code was previously passing three parameters (`newAddress`, `newWalletAddress`, and `Date.now()`). This has been corrected to only pass the two expected parameters.
2. **Audit payload cleanup:** Removed the duplicate and redundant property declarations (`old_wallet_address` and `new_wallet_address`) in the Supabase audit logs insertion payload.
3. **Test Mock Update:** Added the missing `supabaseAdmin` export to the mocked `db.js` file in `keyRotationService.test.js` to ensure the unit tests run and pass without throwing mock export errors.
## Type of Change
- [x] Bug fix (non-breaking change fixing an issue)
- [ ] New feature (non-breaking change adding functionality)
- [ ] Refactoring (code cleanup, performance, or structural changes)
## How Has This Been Tested?
- **Test Commands:**
  Run the key rotation unit tests:
  ```bash
  npx vitest run test/unit/security/keyRotationService.test.js
Test Steps / Prerequisites:
Navigate to the backend/api/ directory.
Run the test command above.
Expected Result: All 15 tests pass successfully, confirming that the on-chain transfer payload structure matches the contract ABI constraints and that the rotation logs are correctly recorded without throwing any errors.
Test Environment
 Unit / Integration tests (npm test, flutter test, etc.)
Related Issues
Fixes #11139 

PR Checklist
 My code follows the code style guidelines of this project.
 I have performed a self-review of my own code.
 I have commented my code, particularly in hard-to-understand areas.
 I have updated corresponding documentation if applicable.
 My changes generate no new warnings or console errors.

PR #11618, opened 2026-08-13
fix(security): encode transferKeyOwnership with matching ABI args (#11608)

### LINKED ISSUE #11608: KeyRotationService encodes 3 args for 2-param transferKeyOwnership ABI - every on-chain key rotation fails
## Summary

`KeyRotationService.transferKeyOwnershipOnChain()` encodes `transferKeyOwnership` with **three** arguments, but the contract ABI declared in the same file only accepts **two** (`address newKeyAddress, uint256 nonce`). `ethers` throws during `encodeFunctionData`, so **every** on-chain key ownership transfer fails before a transaction is ever sent.

## Location

- `backend/api/src/services/security/keyRotationService.js`

## Evidence

ABI (lines 8-11):

```js
const ESCROW_ABI = [
  'function transferKeyOwnership(address newKeyAddress, uint256 nonce) external',
  'function verifyKeyOwnership(address keyAddress) view returns (bool)',
];
```

Call site (lines 200-207):

```js
const tx = await oldWallet.sendTransaction({
  to: this.escrowContract.target,
  data: this.escrowContract.interface.encodeFunctionData('transferKeyOwnership', [
    newAddress,
    newWalletAddress,  // public address only — never the private key
    Date.now(),
  ]),
});
```

Three values are passed for a two-parameter signature → `encodeFunctionData` throws (incorrect number of arguments) → the whole rotation fails.

`backend/api/test/unit/security/keyRotationService.test.js` fails 10/15; the failure group includes "does not fail the rotation when the best-effort receipt write fails" and the on-chain transfer paths.

## Expected behavior

- The encoded calldata must match the ABI: `transferKeyOwnership(newAddress, nonce)` (two args). `newWalletAddress` and `Date.now()` must not both be passed as positional params.

## Suggested fix

Encode exactly the two declared parameters, e.g.:

```js
data: this.escrowContract.interface.encodeFunctionData('transferKeyOwnership', [
  newAddress,
  Date.now(),
]),
```

## Notes

- Secondary (same function): the `insert` object on lines 218-229 duplicates keys `old_wallet_address` and `new_wallet_address`; the duplicate overwrites the intended value.

Akshita-2307:
## Summary
Fixes #11608 — on-chain key rotation was failing because `transferKeyOwnership` was encoded with 3 args against a 2-param ABI (`transferKeyOwnership(address newKeyAddress, uint256 nonce)`).

The source-side fix for the ABI mismatch already landed upstream via #11141 (now encoded as `[newWalletAddress, Date.now()]`). This PR was rebuilt on top of current `main` and now contributes the remaining piece: the `config/db.js` mock in the unit test exports `supabaseAdmin` so the mock mirrors the real module exports.

## Changes
- `backend/api/test/unit/security/keyRotationService.test.js` — add `supabaseAdmin: { from }` to the `config/db.js` mock.

## Verification
- `npx vitest run test/unit/security/keyRotationService.test.js` → all pass.
- Branch is rebased on latest `main`; mergeable clean.

## GSSOC
This PR is submitted as part of GSSoC 2025 Extended. Please add the gssoc-ext / gssoc labels.

### DIFF
--- a/backend/api/test/unit/security/keyRotationService.test.js
+++ b/backend/api/test/unit/security/keyRotationService.test.js
@@ -38,7 +38,8 @@ vi.mock('../../../src/config/db.js', () => ({
         eq: vi.fn().mockResolvedValue({ error: null })
       }))
     }))
-  }
+  },
+  supabaseAdmin: { from: vi.fn() },
 }));
 
 vi.mock('../../../src/middleware/logger.js', () => ({


Akshita-2307:
Rebuilt this branch on top of the latest `main` — the source-side ABI fix already landed via #11141, so this PR now carries only the missing `supabaseAdmin` mock export. Resolving the conflict.

Akshita-2307:
Rebuilding head as a clean descendant of main; reopening for a fresh mergeability check.

PR #11699, opened 2026-08-13
fix(api): implement safe json parsing fallback in request context middleware

tmdeveloper007:
## Spec 1

**Problem:** `JSON.parse` on malformed incoming context headers causes uncaught `SyntaxError` crashes.

**Solution:** Wrap `JSON.parse` in try-catch with structured fallback to standard request metadata.

**Verification:** ``cd backend/api && npx vitest run test/unit/requestContext.test.js``

Closes spec #1


KanishJebaMathewM:
level:advanced

PR #11707, opened 2026-08-13
fix(api): fix TTL drift calculation in profile cache expiration

tmdeveloper007:
## Spec 9

**Problem:** Calculated TTL uses local clock milliseconds instead of standard second-based Unix epoch intervals.

**Solution:** Standardize TTL calculations using relative duration in seconds.

**Verification:** ``cd backend/api && npx vitest run test/unit/profileCacheTtl.test.js``

Closes spec #9


KanishJebaMathewM:
level:advanced

PR #11725, opened 2026-08-13
fix(api): implement constant-time HMAC comparison in auth tokens

tmdeveloper007:
## Spec 12

**Problem:** Standard string equality `===` leaks timing details during token verification.

**Solution:** Use `crypto.timingSafeEqual` for string hash comparison.

**Verification:** ``cd backend/api && npx vitest run test/unit/otpHashing.test.js``

Closes spec #12


KanishJebaMathewM:
level:advanced

PR #11726, opened 2026-08-13
fix(api): clamp fraud risk scores to valid [0, 100] range

tmdeveloper007:
## Spec 13

**Problem:** Accumulated risk point weights can exceed 100 or drop below 0.

**Solution:** Math clamp risk calculations to bounds `[0, 100]`.

**Verification:** ``cd backend/api && npx vitest run test/unit/fraudMiddleware.test.js``

Closes spec #13


KanishJebaMathewM:
level:advanced

PR #11731, opened 2026-08-13
fix(api): enforce maximum request body size limit on file upload routes

tmdeveloper007:
## Spec 18

**Problem:** Large file streams bypass Express payload limits if multipart headers are malformed.

**Solution:** Intercept stream length in middleware and abort request with HTTP 413.

**Verification:** ``cd backend/api && npx vitest run test/unit/uploadFilename.test.js``

Closes spec #18


KanishJebaMathewM:
level:advanced

PR #11732, opened 2026-08-13
fix(api): fix race condition in in-memory LRU cache eviction

tmdeveloper007:
## Spec 19

**Problem:** Concurrent `get` and `set` calls during key eviction result in orphaned linked list pointers.

**Solution:** Synchronize tail node updates and update internal count atomically.

**Verification:** ``cd backend/api && npx vitest run test/unit/lruCache.test.js``

Closes spec #19


KanishJebaMathewM:
level:advanced

PR #11744, opened 2026-08-13
fix(kafka): add idempotent message deduplication in consumer pipeline

tmdeveloper007:
## Spec 31

**Problem:** Redelivered Kafka messages result in duplicate database state mutations.

**Solution:** Cache processed event message keys in Redis idempotency set with 24h TTL.

**Verification:** ``cd backend/kafka && npm test test/order.consumer.test.js``

Closes spec #31


KanishJebaMathewM:
level:advanced

PR #11758, opened 2026-08-13
fix(mev): fix gas price calculation boundary in MEV service

tmdeveloper007:
## Spec 42

**Problem:** Integer division underflow sets priority fee to 0 wei under high network load.

**Solution:** Use `BigInt` precision arithmetic for priority fee calculation.

Closes spec #42


KanishJebaMathewM:
level:advanced

PR #11760, opened 2026-08-13
fix(mev): limit maximum batch size in MEV bundle submission queue

tmdeveloper007:
## Spec 44

**Problem:** Oversized transaction bundles exceed block gas limit and fail silently.

**Solution:** Cap bundle size to maximum 10 transactions or 15M total gas.

Closes spec #44


KanishJebaMathewM:
level:advanced

PR #11762, opened 2026-08-13
fix(eventsourcing): fix event stream divergence in snapshot rebuilder

tmdeveloper007:
## Spec 36

**Problem:** Out-of-order event timestamps corrupt calculated state snapshot.

**Solution:** Sort event stream explicitly by sequence number rather than clock timestamp.

Closes spec #36


KanishJebaMathewM:
level:advanced

PR #11763, opened 2026-08-13
fix(eventsourcing): enforce sequence continuity in event store insertion

tmdeveloper007:
## Spec 37

**Problem:** Missing sequence number gap detection allows dropped events to go undetected.

**Solution:** Verify `sequence_nr == last_sequence_nr + 1` before persisting new event.

Closes spec #37


KanishJebaMathewM:
level:advanced

PR #11764, opened 2026-08-13
fix(eventsourcing): optimize event store query memory footprint

tmdeveloper007:
## Spec 38

**Problem:** Fetching full event history loads thousands of objects into RAM at once.

**Solution:** Implement stream cursor iterator for processing historical events in chunks.

Closes spec #38


KanishJebaMathewM:
level:advanced

PR #11765, opened 2026-08-13
fix(pqc): fix key generation seed length check in NTRU signer

tmdeveloper007:
## Spec 46

**Problem:** Insufficient entropy in seed vector produces weak NTRU key pairs.

**Solution:** Enforce minimum 256-bit entropy seed validation before key generation.

**Verification:** ``cd backend/pqc && python3 test_ntru.py``

Closes spec #46


KanishJebaMathewM:
level:advanced

PR #11767, opened 2026-08-13
fix(crypto): fix index lookup leak in searchable symmetric encryption engine

tmdeveloper007:
## Spec 48

**Problem:** Unblinded trapdoors leak frequency analysis of search query terms.

**Solution:** Implement randomized HMAC-based trapdoors with nonces per search request.

**Verification:** ``cd backend/crypto && python3 test_sse.py``

Closes spec #48


KanishJebaMathewM:
level:advanced

PR #11768, opened 2026-08-13
fix(crypto): fix memory leak in encrypted inverted index table

tmdeveloper007:
## Spec 49

**Problem:** Deleted document tokens remain in index dictionary indefinitely.

**Solution:** Add garbage collection sweep to prune orphaned document IDs from inverted index.

**Verification:** ``cd backend/crypto && python3 test_sse.py``

Closes spec #49


KanishJebaMathewM:
level:advanced

PR #11769, opened 2026-08-13
fix(zk_mpc): add input size validation in private set intersection matcher

tmdeveloper007:
## Spec 50

**Problem:** Disproportionately large input sets enable denial-of-service via memory exhaustion.

**Solution:** Limit maximum set size to 100,000 items and throw error on violation.

**Verification:** ``cd backend/zk_mpc && python3 test_psi.py``

Closes spec #50


KanishJebaMathewM:
level:advanced

PR #11851, opened 2026-08-13
fix(api): add atomicity and sliding-window cleanup to redisRateLimiter

tmdeveloper007:
## Spec 2

**Problem:** Non-atomic Redis evaluation under high burst concurrency leads to rate limit bypasses.

**Solution:** Refactor rate limit check to execute inside a single atomic Redis Lua script with automatic zremrangebyscore window pruning.

**Verification:** ``cd backend/api && npx vitest run test/unit/redisRateLimiterPipeline.test.js``

Closes spec #2


KanishJebaMathewM:
level:advanced

PR #11853, opened 2026-08-13
fix(api): implement fail-closed auth failure monitor on redis timeout

tmdeveloper007:
## Spec 4

**Problem:** Monitor silently succeeds when Redis connection drops, bypassing IP lockout logic.

**Solution:** Enforce fail-closed policy that rejects suspicious requests when security counters cannot be queried.

**Verification:** ``cd backend/api && npx vitest run test/unit/authFailureMonitorBound.test.js``

Closes spec #4


KanishJebaMathewM:
level:advanced

PR #11855, opened 2026-08-13
fix(api): prevent prototype pollution in request query sanitizer

tmdeveloper007:
## Spec 6

**Problem:** Deeply nested query keys (`__proto__`, `constructor.prototype`) pollute object prototype during query parsing.

**Solution:** Add key sanitization filter rejecting dangerous keys before processing parameters.

**Verification:** ``cd backend/api && npx vitest run test/unit/suspiciousRequestsUserAgent.test.js``

Closes spec #6


KanishJebaMathewM:
level:advanced

PR #11856, opened 2026-08-13
fix(api): add strict MIME signature validation for file uploads

tmdeveloper007:
## Spec 7

**Problem:** Extension-only file validation allows spoofed files with invalid binary magic headers.

**Solution:** Read initial 4-byte buffer header signatures (magic numbers) to match claimed MIME types.

**Verification:** ``cd backend/api && npx vitest run test/unit/documentValidation.test.js``

Closes spec #7


KanishJebaMathewM:
level:advanced

PR #11857, opened 2026-08-13
fix(api): enforce strict schema validation on read model profile caches

tmdeveloper007:
## Spec 8

**Problem:** Incomplete profile objects stored in Redis trigger downstream property access errors.

**Solution:** Add Zod schema validation pass prior to setting and retrieving profile cache entries.

**Verification:** ``cd backend/api && npx vitest run test/unit/profileCacheParse.test.js``

Closes spec #8


KanishJebaMathewM:
level:advanced

PR #11858, opened 2026-08-13
fix(api): validate non-negative values in pricing calculator helper

tmdeveloper007:
## Spec 10

**Problem:** Negative distance inputs result in negative price estimates for freight orders.

**Solution:** Guard inputs with `Math.max(0, distance)` and throw validation errors for invalid numbers.

**Verification:** ``cd backend/api && npx vitest run test/unit/pricing.test.js``

Closes spec #10


KanishJebaMathewM:
level:advanced

PR #11859, opened 2026-08-13
fix(api): prevent HSTS header duplication on security proxy middleware

tmdeveloper007:
## Spec 11

**Problem:** Duplicate `Strict-Transport-Security` headers sent when downstream proxy appends header.

**Solution:** Check if `res.getHeader('Strict-Transport-Security')` exists before appending.

**Verification:** ``cd backend/api && npx vitest run test/unit/securityHeadersHsts.test.js``

Closes spec #11


KanishJebaMathewM:
level:advanced

PR #11860, opened 2026-08-13
fix(api): handle missing correlation IDs in async request hooks

tmdeveloper007:
## Spec 14

**Problem:** Concurrent request contexts lose correlation trace IDs across async await boundaries.

**Solution:** Use Node.js `AsyncLocalStorage` to maintain context throughout execution.

**Verification:** ``cd backend/api && npx vitest run test/unit/correlationId.test.js``

Closes spec #14


KanishJebaMathewM:
level:advanced

PR #11861, opened 2026-08-13
fix(api): fix double release bug in Redis distributed lock

tmdeveloper007:
## Spec 15

**Problem:** Calling unlock twice on an expired lock throws unexpected uncaught exception.

**Solution:** Maintain internal lock state flag and return false safely on redundant unlocks.

**Verification:** ``cd backend/api && npx vitest run test/unit/redisLock.test.js``

Closes spec #15


KanishJebaMathewM:
level:advanced

PR #11862, opened 2026-08-13
fix(api): handle clamav connection reset gracefully in malware scanner

tmdeveloper007:
## Spec 16

**Problem:** TCP connection drop during scan stream causes uncaught promise rejection.

**Solution:** Add socket timeout listener and propagate structured `ScanServiceUnavailableError`.

**Verification:** ``cd backend/api && npx vitest run test/unit/malwareScanner.test.js``

Closes spec #16


KanishJebaMathewM:
level:advanced

PR #11863, opened 2026-08-13
fix(api): sanitize SQL wildcards in user search escape helper

tmdeveloper007:
## Spec 17

**Problem:** Unescaped `%` and `_` lead to unintentional full table scan queries.

**Solution:** Replace `\`, `%`, `_` characters with escaped backslash equivalents.

**Verification:** ``cd backend/api && npx vitest run test/unit/escapeLike.test.js``

Closes spec #17


KanishJebaMathewM:
level:advanced

PR #11865, opened 2026-08-13
fix(api): add validation for geohash precision limits in spatial queries

tmdeveloper007:
## Spec 21

**Problem:** Invalid precision parameters (`> 12` or `< 1`) cause buffer overflow exceptions.

**Solution:** Enforce bounds validation and default to standard precision level 6.

**Verification:** ``cd backend/api && npx vitest run test/unit/reverseGeocode.test.js``

Closes spec #21


KanishJebaMathewM:
level:advanced

PR #11866, opened 2026-08-13
fix(api): handle OSRM server connection reset with failover fallback

tmdeveloper007:
## Spec 22

**Problem:** Primary routing node network glitch causes infinite request stall.

**Solution:** Implement circuit breaker fallback with straight-line distance estimation on timeout.

**Verification:** ``cd backend/api && npx vitest run test/unit/osrm.test.js``

Closes spec #22


KanishJebaMathewM:
level:critical

PR #11867, opened 2026-08-13
fix(api): enforce rating scale bounds in driver reputation service

tmdeveloper007:
## Spec 23

**Problem:** Aggregated driver ratings exceed maximum 5.0 score due to unrounded floating addition.

**Solution:** Round output ratings to 2 decimal places and cap in range `[1.00, 5.00]`.

**Verification:** ``cd backend/api && npx vitest run test/unit/reputation.test.js``

Closes spec #23


KanishJebaMathewM:
level:advanced

PR #11869, opened 2026-08-13
fix(api): fix memory leak in request cache class event listeners

tmdeveloper007:
## Spec 25

**Problem:** EventEmitter listeners registered per request are not unsubscribed on response completion.

**Solution:** Remove event listeners on `res.on('finish')` and `res.on('close')`.

**Verification:** ``cd backend/api && npx vitest run test/unit/requestCacheClass.test.js``

Closes spec #25


KanishJebaMathewM:
level:advanced

PR #11881, opened 2026-08-13
fix(mev): prevent sandwich attack exploitation in MEV protection service

tmdeveloper007:
## Spec 41

**Problem:** Public transaction pool leakage allows front-running of high-value freight escrow bids.

**Solution:** Encrypt pending transactions via commit-reveal scheme until block inclusion.

Closes spec #41


KanishJebaMathewM:
level:advanced

PR #11882, opened 2026-08-13
fix(mev): validate transaction payload signatures in MEV routing handler

tmdeveloper007:
## Spec 43

**Problem:** Unsigned MEV bundle submissions execute without caller authorization.

**Solution:** Recover sender address from ECDSA signature before accepting bundle.

Closes spec #43


KanishJebaMathewM:
level:advanced

PR #11931, opened 2026-08-13
fix(telemetry): make rate limiting and driver capacity concurrency-safe

### LINKED ISSUE #11637: [ADVANCED] ML engine model artifacts are written without mutual exclusion or atomicity: concurrent trainings corrupt the pickle, timeout "cancelled" trainings still deploy, and pkl/metadata are committed separately
## Problem Statement

The FastAPI ML engine in `backend/ml` does not safely coordinate concurrent model training or atomically publish model artifacts.

Concurrent training requests can write to the same temporary model file, training requests that time out can continue running in the background and later deploy their model, and the model artifact and metadata are committed through separate filesystem operations.

This can leave the ML service with corrupted or inconsistent model artifacts.

## Current Behavior

`save_model` in `backend/ml/app/models/base.py` uses a fixed temporary file for model serialization before replacing the final model file.

The lazy training path in `demand_forecast.py` can trigger model training when the model cache is empty, but the global cache is not protected by a synchronization mechanism.

The repository already contains model-lock infrastructure, but the relevant locking path is not correctly connected to the active training flow.

Training endpoints in `backend/ml/main.py` use `asyncio.wait_for()` around executor-based training.

When the timeout expires, the HTTP request can return a timeout while the executor thread continues running.

That background training can still finish and publish a new model.

The model `.pkl` file and metadata JSON are also committed through separate rename operations.

## Root Cause

The training lifecycle does not coordinate:

- Concurrent training.
- Artifact publication.
- Training cancellation.
- Model versioning.
- Metadata consistency.

`asyncio.wait_for()` cancels the awaitable but cannot safely terminate an already-running thread in a `ThreadPoolExecutor`.

The filesystem publication process also lacks a single atomic generation/version pointer.

## Failure Scenario

Two requests trigger demand-model training at the same time.

Both training jobs attempt to write the same temporary model file.

The resulting artifact can become corrupted or one training can overwrite another.

In another scenario, a training request exceeds its timeout and returns HTTP 504, but the underlying thread continues training.

The training later completes and silently replaces the production model even though the client was told that the training failed.

A process crash between model and metadata replacement can also leave a new model paired with old metadata.

## Impact

- Corrupted model artifacts.
- Production prediction failures.
- Non-deterministic model deployment.
- A model can be deployed after its training request reported failure.
- Model metadata can describe a different artifact.
- Difficult rollback and auditability.
- Potential ML service outage.

## Why Advanced?

The fix requires coordinating concurrent training jobs, thread cancellation semantics, atomic filesystem publication, model generations, rollback, and lifecycle state.

The system must treat model deployment as a durable state transition rather than simply writing a file.

## Proposed Solution

Implement per-model training coordination.

Prevent simultaneous training of the same model or explicitly queue/reject concurrent requests.

Use unique temporary artifact paths for every training run.

Publish model and metadata as one consistent generation using a version/pointer mechanism.

Track training jobs and their states.

Ensure a timed-out request cannot silently deploy an artifact without the deployment state reflecting the actual result.

Add model versioning and rollback support.

Apply the same lifecycle rules to all model-training endpoints.

## Files / Components Affected

- `backend/ml/app/models/base.py`
- `backend/ml/app/models/demand_forecast.py`
- `backend/ml/main.py`
- `backend/ml/app/execution.py`
- `backend/ml/routes/eta_routes.py`

## Acceptance Criteria

- [ ] Concurrent training of the same model cannot corrupt artifacts.
- [ ] Concurrent training is rejected, serialized, or safely queued.
- [ ] A timed-out training cannot silently deploy an unexpected artifact.
- [ ] Model and metadata always belong to the same generation.
- [ ] Model versions can be identified.
- [ ] A previous valid model can be restored after a failed deployment.
- [ ] Training state is visible through the relevant health/model endpoints.

## Testing Requirements

- Run multiple concurrent training requests and verify artifact integrity.
- Simulate a training timeout and verify no unexpected deployment occurs.
- Inject a process failure between artifact publication steps.
- Verify model and metadata generations always match.
- Test rollback from a newer model generation to a previous valid generation.

## Expected Outcome

The ML engine can safely train, deploy, recover, and roll back models without corrupted artifacts, torn model/metadata states, or unexpected deployment after a failed training request.

### LINKED ISSUE #11638: [ADVANCED] Telemetry ingestion rate-limiter state is evicted mid-request (TOCTOU), resetting rate windows, and the active-driver cap is a racy check-then-act that triggers O(N) full-map sweeps under load
## Problem Statement

The Go telemetry ingestion service contains concurrency problems in both its per-driver rate limiter and active-driver capacity management.

A rate-limit entry can be evicted while a request is still using it, causing the driver's rate-limit window to reset.

The active-driver capacity check is also implemented as a non-atomic check-then-act operation and performs full-map scans on the request path.

## Current Behavior

`allowGeofence()` loads or creates a rate-limit entry and later operates on that entry.

Meanwhile, `evictGeofenceOverflow()` can remove the same driver entry from the global map.

The request may continue modifying the old entry while future requests receive a newly created entry with an empty rate-limit window.

`storePing()` uses `countActiveDrivers()` to scan the complete driver map before deciding whether a new driver can be inserted.

It may then execute `sweepDrivers()` and perform another complete scan before storing the driver.

Multiple concurrent requests can pass the capacity check at the same time and exceed the configured maximum.

## Root Cause

The implementation contains several concurrency problems:

1. Rate-limit entries can be removed while requests still hold references to them.
2. Rate-limit state is therefore vulnerable to stale/orphaned updates.
3. Active-driver capacity uses a check-then-act pattern rather than an atomic reservation.
4. Full-map scans occur on the request hot path.
5. The time-based pruning function exists but is not actively used.

## Failure Scenario

A large number of unique driver IDs are submitted concurrently.

The service exceeds the configured rate-limit map size and begins evicting entries.

A legitimate driver's entry is removed while its request is still processing.

The request updates the old entry, but the next request creates a new entry with an empty timestamp window.

The driver can therefore send more requests than the configured rate limit.

At the same time, many new drivers reach the active-driver capacity threshold simultaneously.

Multiple requests pass the capacity check and execute expensive full-map sweeps, causing CPU contention and allowing the cache to exceed its configured limit.

## Impact

- Rate limits can be bypassed under concurrent churn.
- Active-driver capacity is not strictly enforced.
- Memory usage can exceed configured limits.
- Request latency increases under load.
- Full-map sweeps create CPU pressure.
- The service becomes less stable precisely when traffic is highest.

## Why Advanced?

This requires designing correct concurrent data structures and atomic admission control.

The implementation must preserve rate-limit ownership during eviction while avoiding O(N) operations on the request path.

A naive mutex around the entire map could eliminate races but introduce severe contention and reduce ingestion throughput.

## Proposed Solution

Make rate-limit entry eviction safe against in-flight requests.

Use entry ownership/versioning or an equivalent mechanism so an evicted entry cannot silently reset the driver's rate-limit state.

Replace the active-driver check-then-act sequence with an atomic capacity reservation mechanism.

Move expensive sweeping away from the request path.

Use incremental cleanup or an appropriate bounded data structure.

Restore or replace time-based expiration so memory does not depend exclusively on insertion-order eviction.

## Files / Components Affected

- `services/telemetry-ingestion/main.go`
- `services/telemetry-ingestion/main_test.go`

Relevant functions include:

- `allowGeofence`
- `evictGeofenceOverflow`
- `pruneGeofenceRateEntries`
- `storePing`
- `sweepDrivers`
- `countActiveDrivers`

## Acceptance Criteria

- [ ] Eviction cannot reset an active driver's rate-limit window.
- [ ] Rate limits remain enforced under concurrent driver churn.
- [ ] `activeDrivers` never exceeds `maxActiveDrivers`.
- [ ] The request path does not perform full-map sweeps.
- [ ] Rate-limit memory remains bounded through safe expiration/cleanup.
- [ ] Concurrent eviction and request handling are race-safe.

## Testing Requirements

- Run Go race detection with concurrent driver activity.
- Stress-test rate limiting with many unique driver IDs.
- Test eviction while a request is actively using the entry.
- Test concurrent active-driver insertion.
- Verify the configured driver cap is never exceeded.
- Benchmark request latency near maximum capacity.

## Expected Outcome

The telemetry ingestion service enforces its rate limits and active-driver capacity correctly under concurrency while maintaining predictable request performance as the number of tracked drivers increases.

KrutagyaKaneria:
Closes https://github.com/KanishJebaMathewM/Truxify/issues/11637

##GSSoC
#ECSoC26

This PR is submitted as part of GirlScript Summer of Code (GSSoC) 2026.

## Summary

This PR fixes concurrency and performance issues in the telemetry ingestion service.

The existing implementation can evict a driver's rate-limit state while a request is still using it, allowing the driver to receive a fresh rate-limit window.

The active-driver capacity check also uses a non-atomic check-then-act pattern and performs expensive full-map scans on the request path.

## Problem

The rate-limit flow can currently behave like:

```text
Request A
   ↓
Load rate-limit entry
   ↓
Entry gets evicted
   ↓
Request A continues using old entry

Request B
   ↓
Creates a new entry
   ↓
Driver receives a fresh rate-limit window

This creates a time-of-check/time-of-use race.

The active-driver capacity flow has a similar issue:

Request A → count → capacity available
Request B → count → capacity available
Request C → count → capacity available
              ↓
         all insert drivers

Multiple concurrent requests can therefore pass the capacity check before any of them updates the shared state.

The request path also performs full-map scans, which can become expensive as the number of tracked drivers increases.

Changes
Rate-Limit State Safety

Updated rate-limit state management so an entry cannot be safely evicted and replaced while an active request is still using the previous state.

This prevents eviction from resetting a driver's rate-limit window.

Concurrent Rate-Limit Enforcement

Updated concurrent rate-limit handling so requests for the same driver operate against consistent shared state.

The existing rate-limit semantics are preserved while making the state safe under concurrent access.

Atomic Active-Driver Capacity

Reworked active-driver admission so the capacity check and reservation are handled atomically.

This prevents multiple concurrent requests from passing the capacity check and collectively exceeding maxActiveDrivers.

Request-Path Performance

Removed expensive full-map cleanup/sweep operations from the telemetry ingestion hot path where appropriate.

Cleanup is handled through a bounded mechanism so request processing does not scale linearly with the number of tracked drivers.

Safe Cleanup

Updated cleanup behavior to ensure that:

Active request state is not incorrectly removed.
Rate-limit windows are not accidentally reset.
Expired entries are eventually removed.
Driver state remains bounded.
Cleanup is safe under concurrent ingestion.
Testing

Added/updated tests covering:

Concurrent requests for the same driver.
Rate-limit enforcement under concurrency.
Eviction during an in-flight request.
Rate-window preservation after eviction.
Concurrent driver insertion.
Active-driver capacity enforcement.
Concurrent cleanup and ingestion.
Expired rate-limit entry cleanup.
Driver inactivity cleanup.
High-concurrency telemetry ingestion.
Go race detector validation.
Acceptance Criteria
 Concurrent requests for the same driver use consistent rate-limit state.
 Eviction cannot reset an active driver's rate-limit window.
 Rate limits remain enforced under concurrent driver churn.
 maxActiveDrivers cannot be exceeded by concurrent requests.
 Active-driver admission is handled atomically.
 Full-map sweeps are removed from the request hot path.
 Cleanup remains bounded.
 Cleanup is safe under concurrent ingestion.
 Existing telemetry behavior remains compatible.
 Race detection passes.
Performance

The request path no longer relies on repeated O(N) scans across all tracked drivers.

This reduces CPU pressure and improves predictability as the number of active drivers increases.

Cleanup remains bounded and is separated from latency-sensitive ingestion work.

Backward Compatibility

No changes are intended to the telemetry API or existing telemetry payload format.

The changes are limited to internal concurrency, capacity management, rate-limit state management, and cleanup behavior.

Scope

This PR is limited to the telemetry ingestion service.

It does not modify:

Blockchain functionality
ML services
WebSocket tracking
Reputation processing
Escrow
Order event sourcing
FCM/notifications
Validation

The implementation was validated using the existing Go test suite and race detector.

The final implementation was reviewed to ensure:

Rate-limit state cannot be incorrectly reset by concurrent eviction.
Active-driver capacity cannot be exceeded.
Request processing does not perform unbounded full-map scans.
Cleanup remains concurrency-safe.
Existing telemetry behavior remains intact.
Related Issue

Closes #11638 

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved telemetry rate limiting under concurrent activity.
  * Prevented active-driver capacity from exceeding configured limits.
  * Improved cleanup of inactive drivers and expired rate-limit entries.
  * Preserved rate-limit windows during in-flight requests and evictions.
  * Added more reliable geofence validation, including radius and coordinate edge cases.
  * Health metrics now accurately reflect active-driver capacity.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->


KanishJebaMathewM:
level:critical

Issue #12138, opened 2026-08-13
[BUG][Flutter][P2] offline_first_sync_service.dart — unguarded jsonDecode/`as` casts on corrupt sync_events rows crash the sync path (unhandled async error)


Aman-bit1:
### Bug
In `apps/driver/lib/services/offline_first_sync_service.dart`, the SQLite rows are decoded with unguarded casts and no error handling:

- `_loadAllEvents` (lines ~116-125): `jsonDecode(row['payload'] as String)`, `row['event_id'] as String`, `row['queued_at'] as int`, `row['synced_at'] as int` — no try/catch.
- `_processSyncQueue` (lines ~188-193): same unguarded `jsonDecode` / `as String` / `as int` inside a `Future.wait` map.

If any stored `sync_events` row is stale/corrupt (e.g. written by an older schema version, partially written on an app kill mid-write, or a non-String payload), a `FormatException`/`TypeError` is thrown.

Both call sites are invoked fire-and-forget (`_initDatabase()` at line 37 is also called without await in the constructor, and `_emitDbSnapshot` at lines 67/96/221 is not awaited), so the exception surfaces as an **unhandled async error**: the sync service silently no-ops, the error is never logged/recovered, and offline trip events stop syncing permanently until the app is restarted.

### Impact
One bad row permanently breaks offline→online sync for that driver with no logging, no retry, and no quarantine of the corrupt row.

### Fix
Wrap the decode in try/catch, validate each field before casting, and skip/flag (or delete) unparseable rows instead of throwing. Await and handle errors from `_initDatabase` and the snapshot emitter.

### Notes
Distinct from #11487 (SyncEngine treats 401 as retryable) — that is `apps/driver/lib/services/sync_engine.dart`.

PR #12532, opened 2026-08-13
fix(driver): decode sync_events rows defensively instead of casting

### LINKED ISSUE #12426: offline_first_sync_service unguarded as casts crash queue load
## Summary
`_loadAllEvents` non-nullable `as` casts SQLite columns (`event_id` as String, `payload` as String, `queued_at` as int, `is_synced` as int, `synced_at` as int). The function is not wrapped in try/catch.

## Location
`apps/driver/lib/services/offline_first_sync_service.dart:139-145`

## Code
```dart
eventId: row['event_id'] as String,
eventType: row['event_type'] as String,
payload: jsonDecode(row['payload'] as String) as Map<String, dynamic>,
queuedAt: DateTime.fromMillisecondsSinceEpoch(row['queued_at'] as int),
isSynced: (row['is_synced'] as int) == 1,
syncedAt: row['synced_at'] != null
    ? DateTime.fromMillisecondsSinceEpoch(row['synced_at'] as int)
    : null,
```

## Impact
If any column is `NULL` (partial write, schema drift, interrupted migration), the non-nullable `as` throws `CastError`, aborting offline-sync queue processing for all events.

## Suggested fix
Use nullable-safe reads (`row['event_id'] as String? ?? ''`, `row['queued_at'] as int? ?? 0`) or wrap `_loadAllEvents` in try/catch with row-level skip.

sarbojitrana:
### Root cause

`_loadAllEvents` maps **every** `sync_events` row through non-nullable `as` casts, inside a single `.map()` over the whole result set and with no error handling:

```dart
return rows.map((row) => OfflineSyncEvent(
  eventId: row['event_id'] as String,
  ...
)).toList();
```

`Map.[]` returns `Object?`, so each `as String` / `as int` is an unchecked downcast from a nullable value. If any column is `NULL` or holds an unexpected type — partial write, schema drift, interrupted migration — the cast throws a `TypeError`. Because the cast happens inside the `.map()`, the throw escapes `_loadAllEvents` entirely: **one bad row takes down the read for every row**, not just itself. `_loadAllEvents` feeds `_emitDbSnapshot()`, which is called unawaited, so the failure surfaces as an unhandled async error and `databaseStream` simply never emits again — the offline dashboard is left permanently stale.

Two failure modes the reported casts don't cover, but which live on the same line:

- `jsonDecode(row['payload'] as String)` throws `FormatException`, not a cast error, when the stored payload isn't valid JSON.
- `... as Map<String, dynamic>` throws when the payload is valid JSON but decodes to an array or scalar.

**The same defect is on the sync path**, which is what actually produces the impact described in the issue. `_runSyncPass` repeats the identical unguarded construction (previously lines 229-235):

```dart
final results = await Future.wait(
  batch.map((row) => _syncSingleEvent(OfflineSyncEvent(
    eventId: row['event_id'] as String, ...
  ))),
);
```

`Iterable.map` is lazy, so the construction runs while `Future.wait` iterates it. A throw there aborts `_runSyncPass` → `_processSyncQueue` and **strands every well-formed event queued behind the bad row**. Fixing only `_loadAllEvents` would have left the more damaging instance in place, so this PR fixes both.

### Fix

Add `OfflineSyncEvent.fromRow`, which returns `null` for a row it cannot trust rather than throwing, and handle that at each call site:

| Site | Policy for an undecodable row |
|---|---|
| `_loadAllEvents` | **Skip** it — the snapshot keeps every well-formed event, so the dashboard stays live. |
| `_runSyncPass` | **Dead-letter** it via the existing `stateDeadLetter`. |

A row that cannot be decoded can never be POSTed, so it is terminal by definition — dead-lettering reuses the state the retry ceiling already uses, instead of leaving a poison row to be re-read on every connectivity change forever. The query now selects `rowid` so such a row is still addressable when `event_id` is itself the undecodable column.

Which columns are required vs. defaulted is deliberate. The issue suggests `row['event_id'] as String? ?? ''`, but defaulting the identity columns is worse than crashing:

- **Required** (`event_id`, `event_type`, `payload`, `queued_at`) — `event_id` addresses the row in every `UPDATE`/`DELETE` and doubles as the backend idempotency key, and `queued_at` is sent as `occurred_at`. An empty-string `event_id` would make the success `DELETE` match the wrong rows, and an empty payload would be POSTed as a real event. Silent data corruption beats a crash only if you don't look. These rows are dropped instead.
- **Defaulted** (`is_synced`, `synced_at`) — advisory only; they never address a row or reach the backend. They fall back to "not synced", where the worst case is a resend that the idempotency key absorbs.

## Type of Change
- [ ] Refactoring (code cleanup, performance, or structural changes)
- [ ] Documentation update

## How Has This Been Tested?

- **Test Commands:** `cd apps/driver && flutter test test/offline_sync_row_decode_test.dart`
- **Test Steps / Prerequisites:** `flutter pub get` in `packages/truxify_shared` and `apps/driver`.
- **Expected Result:** 8 tests pass. `OfflineSyncEvent.fromRow` is a pure static, so the suite needs no sqflite/Firebase platform plugin.

`test/offline_sync_row_decode_test.dart` covers the well-formed row, each required column NULL, each required column mistyped, empty identity/type, a payload that is not valid JSON and one that is a JSON array, advisory-column degradation, and a mixed result set. The last case is the regression anchor — it asserts the **pre-fix** mapping throws `TypeError` on that result set while the fixed mapping still returns the two good events.

### Proof of work

Full Flutter wasn't available locally, so I verified the decoder by running the **actual** `offline_sync_event_model.dart` (symlinked into a scratch package, checksum-matched) against a standalone Dart SDK 3.13.0, with a harness mirroring the committed test plus a side-by-side reproduction of the pre-fix mapping:

```
== 1. Reproduce the reported crash (pre-fix mapping) ==
  THROWS     event_id NULL -> _TypeError
  THROWS     event_type NULL -> _TypeError
  THROWS     payload NULL -> _TypeError
  THROWS     queued_at NULL -> _TypeError
  THROWS     is_synced NULL -> _TypeError
  THROWS     synced_at mistyped -> _TypeError
  THROWS     payload not JSON -> FormatException
  THROWS     payload JSON array -> _TypeError

== 2. Fixed decoder: same rows, no throw ==
  dropped    event_id NULL
  dropped    event_type NULL
  dropped    payload NULL
  dropped    queued_at NULL
  kept       is_synced NULL
  kept       synced_at mistyped
  dropped    payload not JSON
  dropped    payload JSON array

== 3. Assertions ==
  PASS  decodes a well-formed row
  PASS  reads sync markers when marked synced
  PASS  NULL required columns -> null
  PASS  mistyped columns -> null
  PASS  empty identity/type -> null
  PASS  payload not a JSON object -> null
  PASS  advisory columns degrade to "not synced"
  PASS  one corrupt row no longer aborts the read

8 passed, 0 failed
```

`dart analyze` reports **No issues found!** on the model. All three changed files parse cleanly.

Two things I could not run locally and am flagging honestly: `flutter analyze` and `flutter test` over the service file (it imports `package:flutter`, `sqflite` and `firebase_auth`) — CI covers both. I also deliberately did **not** run `dart format`: the repo predates the Dart 3.7+ "tall style" formatter (566 of 706 files in `apps/driver` are reformatted by it, and `lint.yml` is already failing on `main` for that reason), so the new code matches the surrounding hand-written style rather than adding unrelated churn.

### Test Environment
- [ ] Local manual testing

## Related Issues
Closes #12426

## PR Checklist

### DIFF
--- a/apps/driver/lib/models/offline_sync_event_model.dart
+++ b/apps/driver/lib/models/offline_sync_event_model.dart
@@ -1,3 +1,5 @@
+import 'dart:convert';
+
 class OfflineSyncEvent {
   final String eventId;
   final String eventType; // 'STATUS_UPDATE', 'LOCATION_PING', 'POD_UPLOAD'
@@ -14,4 +16,52 @@ class OfflineSyncEvent {
     this.isSynced = false,
     this.syncedAt,
   });
+
+  /// Decodes one `sync_events` row into an event, or returns `null` when the
+  /// row cannot be trusted.
+  ///
+  /// Rows can drift from the schema through a partial write, a schema change
+  /// or an interrupted migration. Reading such a row with a non-nullable `as`
+  /// throws, and because the callers map over a whole result set one bad row
+  /// used to abort the entire read (issue #12426). Returning `null` lets the
+  /// caller drop that single row and keep going.
+  ///
+  /// Identity and content columns are required rather than defaulted:
+  /// `event_id` addresses the row in every UPDATE/DELETE and doubles as the
+  /// backend idempotency key, `payload`/`event_type` are the event itself, and
+  /// `queued_at` is sent as `occurred_at`. Substituting a placeholder for any
+  /// of them would push wrong data to the backend or target the wrong row, so
+  /// an undecodable row is dropped instead. `is_synced` and `synced_at` are
+  /// advisory, so they fall back to "not synced" — the worst case is a resend
+  /// that the idempotency key absorbs.
+  static OfflineSyncEvent? fromRow(Map<String, Object?> row) {
+    final eventId = row['event_id'];
+    final eventType = row['event_type'];
+    final rawPayload = row['payload'];
+    final queuedAt = row['queued_at'];
+
+    if (eventId is! String || eventId.isEmpty) return null;
+    if (eventType is! String || eventType.isEmpty) return null;
+    if (rawPayload is! String) return null;
+    if (queuedAt is! int) return null;
+
+    Object? payload;
+    try {
+      payload = jsonDecode(rawPayload);
+    } on FormatException {
+      return null;
+    }
+    if (payload is! Map<String, dynamic>) return null;
+
+    final syncedAt = row['synced_at'];
+    return OfflineSyncEvent(
+      eventId: eventId,
+      eventType: eventType,
+      payload: payload,
+      queuedAt: DateTime.fromMillisecondsSinceEpoch(queuedAt),
+      isSynced: row['is_synced'] == 1,
+      syncedAt:
+          syncedAt is int ? DateTime.fromMillisecondsSinceEpoch(syncedAt) : null,
+    );
+  }
 }
--- a/apps/driver/lib/services/offline_first_sync_service.dart
+++ b/apps/driver/lib/services/offline_first_sync_service.dart
@@ -135,16 +135,34 @@ class OfflineFirstSyncService {
     if (db == null) return [];
 
     final rows = await db.query('sync_events', orderBy: 'queued_at ASC');
-    return rows.map((row) => OfflineSyncEvent(
-      eventId: row['event_id'] as String,
-      eventType: row['event_type'] as String,
-      payload: jsonDecode(row['payload'] as String) as Map<String, dynamic>,
-      queuedAt: DateTime.fromMillisecondsSinceEpoch(row['queued_at'] as int),
-      isSynced: (row['is_synced'] as int) == 1,
-      syncedAt: row['synced_at'] != null
-          ? DateTime.fromMillisecondsSinceEpoch(row['synced_at'] as int)
-          : null,
-    )).toList();
+
+    // Skip rows that cannot be decoded instead of letting one of them throw
+    // out of the snapshot. This feeds [databaseStream]; a single corrupt row
+    // used to leave the dashboard permanently stale (issue #12426).
+    final events = <OfflineSyncEvent>[];
+    for (final row in rows) {
+      final event = OfflineSyncEvent.fromRow(row);
+      if (event == null) {
+        debugPrint('offline-sync: skipping undecodable sync_events row in snapshot');
+        continue;
+      }
+      events.add(event);
+    }
+    return events;
+  }
+
+  /// Marks a row that [OfflineSyncEvent.fromRow] rejected as dead-lettered.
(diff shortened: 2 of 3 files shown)


sarbojitrana:
Closing this — it's been superseded. #12663 landed the nullable-safe casts for #12426, and #12459 (for the duplicate issue #12138) then replaced that with a `_parseRow` + dead-letter design that covers what this PR was doing. No point re-litigating code that's already on `main`.

Leaving the verification behind in case it's useful. I diffed this branch's decoder against a verbatim port of `main`'s merged `_parseRow` over 14 corrupt-row cases — they agree on 12:

| case | `main` (#12459) | this PR |
|---|---|---|
| well-formed | kept | kept |
| `event_id` NULL / mistyped | dropped | dropped |
| **`event_id` empty string** | **kept** | **dropped** |
| `event_type` NULL / mistyped | dropped | dropped |
| **`event_type` empty string** | **kept** | **dropped** |
| `payload` NULL / not JSON / JSON array / JSON string | dropped | dropped |
| `queued_at` NULL / mistyped | dropped | dropped |
| `is_synced` NULL, `synced_at` mistyped | kept | kept |

Two small gaps look like they're still open on `main`, if anyone wants to pick them up:

**1. A row with a NULL `event_id` can never be dead-lettered.** `_runSyncPass` does:

```dart
final eventId = row['event_id'];
if (eventId is! String) continue;   // row is skipped before it can be dead-lettered
```

and `_loadAllEvents` has the matching `if (eventId is String)` guard around its `UPDATE`. Since the pending query is `is_synced = 0 AND state = 'pending'`, such a row stays pending and gets re-read and re-parsed on every sync pass, indefinitely. It's cheap (no network, no crash), but it never drains. Selecting `columns: ['rowid', '*']` and dead-lettering by `rowid` addresses it, since `rowid` is always present and is unaffected by whichever column is the corrupt one.

**2. `event_id: ''` is currently accepted.** `_parseRow` checks `eventId is! String`, so an empty string passes. It then flows into `'idempotencyKey': event.eventId` and into `DELETE ... WHERE event_id = ?` on success. Adding `.isEmpty` to the identity checks would close it.

Neither is urgent. Worth noting there's also no test coverage on this path yet — `_parseRow` is a private instance method, so reaching it from `test/` would need it made `static` + `@visibleForTesting`, after which the corrupt-row matrix above tests cleanly with no sqflite or Firebase plugin.

Thanks @Akshita-2307 for #12459.

PR #12571, opened 2026-08-13
fix(api): remove duplicate deactivation update in keyManagementService storeEncryptedKey

### LINKED ISSUE #12475: [BUG][backend][P2] keyManagementService.storeEncryptedKey runs the same deactivation update twice
### Bug
`backend/api/src/services/security/keyManagementService.js` deactivates prior active keys for the same user/wallet scope **twice** in `storeEncryptedKey`:

- lines 112-122: `.update({ active: false }).eq('user_id', userId).eq('wallet_address', walletAddress).eq('active', true)` (throws on error)
- lines 126-136: the **same** `.update({ active: false, archived_at: ..., archive_reason: 'rotated' }).eq(...).eq(...).eq('active', true)` (ignores error)

Between the two there is only `const keyId = crypto.randomUUID();` — nothing changes state, so the first update already does all the work and the second is a redundant, duplicated write to the database.

### Fix
Collapse into a single update that sets `active: false` plus `archived_at`/`archive_reason`, with the throw-on-error handling kept.

### Notes
Distinct from #9931 (missing DDL) and #8264 (missing await on insert) — this is the duplicated deactivation write.

Xenon010101:
## Summary
Fixes `keyManagementService.storeEncryptedKey` running the same deactivation update twice, where the second run never matches any row and silently drops the archive step.

## Changes
- Replaced the two back-to-back `update(...).eq('active', true)` chains with a single update that deactivates **and** archives the previous key:
  - `{ active: false, archived_at: now, archive_reason: 'rotated' }` filtered by `user_id`, `wallet_address`, and `active = true`.
- Removed the dead second block that re-queried with an extra `.eq('active', true)` — after the first update already set `active = false`, the second query matched 0 rows, so keys were never archived.

## Impact
Previous keys are now actually archived on rotation instead of being orphaned. `keyManagementService` and `keyManagementService.rotationHistory` unit suites pass.

Fixes #12475

GSSOC


Xenon010101:
Subsumed upstream: main now performs a single deactivate+archive update in storeEncryptedKey (duplicate second update already removed). This branch's change is a no-op against current main. Closing as obsolete.

Xenon010101:
Superseded upstream - see above

PR #12673, opened 2026-08-13
fix(blockchain): guard against re-entrant block scanning

### LINKED ISSUE #12601: bug: BlockchainMonitor Re-entrant Block Scanning Causes Duplicate Event Processing
Bug: startPollingBlocks() (line 97) uses a bare setInterval with an async callback to poll blocks every 12s. There is no re-entrancy guard: if scanBlockRange() takes longer than the poll interval (e.g., RPC latency spike, large block range), the next interval fires while the previous scan is still running. Both invocations read the same this.lastBlockScanned, scan the same block range, and process the same logs — triggering duplicate PaymentReceived, InsuranceClaimApproved, and GeofenceBreach event handlers. This can cause double wallet credits, duplicate insurance payouts, and false geofence alerts.

Fix: Add an isScanning boolean guard. Set it true before scanBlockRange and false in a finally block. Skip the interval tick if isScanning is already true.

Dev1822:
## Description

Closes #12601

\startPollingBlocks()\ uses a bare \setInterval\ with an async callback to poll blocks every 12s. There was no re-entrancy guard: if \scanBlockRange()\ took longer than the poll interval (e.g., RPC latency spike, large block range), the next interval fired while the previous scan was still running. Both invocations read the same \	his.lastBlockScanned\, scanned the same block range, and processed the same logs — triggering duplicate \PaymentReceived\, \InsuranceClaimApproved\, and \GeofenceBreach\ event handlers. This could cause double wallet credits, duplicate insurance payouts, and false geofence alerts.

## Fix

- Added an \isScanning\ boolean guard (initialized to \alse\ in the constructor).
- Set it to \	rue\ before \scanBlockRange\ runs and reset it to \alse\ in a \inally\ block so it is always cleared, even on error.
- The interval tick now skips (with a warning log) if \isScanning\ is already \	rue\.

## Testing

- Added unit tests verifying:
  - a concurrent interval tick is skipped while a scan is in flight (\getBlockNumber\ called only once);
  - \isScanning\ is cleared in the \inally\ block even when the scan throws.
- \
px vitest run test/unit/services/blockchain test/unit/blockchainMonitorGuard.test.js\ — 42/42 passing.
- \
px eslint\ on the modified files — clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved blockchain monitoring to prevent overlapping scans during polling.
  - Ensured monitoring resumes correctly after successful or failed scans.

- **Tests**
  - Added coverage for overlapping polling intervals.
  - Added validation that scan status is reset after both successful and failed scans.

### DIFF
--- a/backend/api/src/services/blockchain/blockchainMonitor.js
+++ b/backend/api/src/services/blockchain/blockchainMonitor.js
@@ -30,6 +30,7 @@ class BlockchainMonitor {
     this.provider = null;
     this.contract = null;
     this.isListening = false;
+    this.isScanning = false;
     this.lastBlockScanned = 0;
     this.eventHandlers = {};
   }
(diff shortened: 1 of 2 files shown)


Dev1822:
GSSOC

PR #12805, opened 2026-08-13
test(api): remove duplicate vitest import in escapeLike tests (#12770)

### LINKED ISSUE #12770: [BUG][api/backend][P2] escapeLike.test.js fails to load - duplicate vitest import at line 49; SQL LIKE escaping tests never run
## Bug
ackend/api/test/unit/escapeLike.test.js cannot be loaded by vitest - the vitest import block is duplicated mid-file, re-declaring the same ESM bindings.

- L1: `import { describe, it, expect } from 'vitest';`
- L49: `import { describe, it, expect } from 'vitest';` (appended inside the `// === Spec 17 test ===` block)

ESM forbids two imports that bind the same name (Identifier 'describe' has already been declared), so the module fails to parse and **0 tests from this file ever run** - the SQL LIKE escaping coverage is silently lost.

## Repro
`node --check backend/api/test/unit/escapeLike.test.js` -> SyntaxError at L49.

## Fix
Delete the second `import ... from 'vitest';` block at L49; the top-level import already provides every binding used by the appended `Spec 17` suite.

Xenon010101:
Closes #12770

## Problem
- escapeLike.test.js had a duplicate vitest import (mid-file re-import), which issue #12770 reported.
- During merge resolution, upstream's rewritten version of this file also introduced two regressions: the escapeLike import path was changed to the invalid '../../../src/lib/escapeLike.js', and escapeSqlLike was used in tests without being imported.

## Fix
- Removed the duplicate vitest import (original #12770 fix).
- Restored the correct import path '../../src/lib/escapeLike.js'.
- Imported escapeSqlLike so the additional-coverage tests actually resolve.
- Verified with node --check.

GSSOC


Xenon010101:
Subsumed upstream: main's #14082 (commit c12e01330) already rewrote escapeLike.test.js with the corrected import path and escapeSqlLike import. This branch now has zero diff against main. Closing as obsolete.

Xenon010101:
Superseded upstream - see above

PR #14385, opened 2026-08-15
fix : guard against undefined tripId in tripValidator middleware

### LINKED ISSUE #14337: fix : guard against undefined tripId in tripValidator middleware
tripValidator middleware in backend/api/src/middleware/tripValidator.js does not check if tripId (from req.params.tripId) is undefined before using it. When tripId is undefined, the middleware may pass null or undefined values to database queries, producing unexpected results.

Changes that are Needed to be Made:
- backend/api/src/middleware/tripValidator.js: Add an early guard at the start of the middleware to check that req.params.tripId is defined and non-empty. If not, return res.status(400).json({ error: 'tripId parameter is required' }) and return.
- Add test coverage for the undefined tripId case in backend/api/test/unit/tripValidator.test.js.

Impact that it would Provide:
- Prevents undefined tripId values from reaching database query logic.
- Returns a clear 400 error instead of silent null propagation.
- Improves API robustness and debuggability.

tmdeveloper007:
## Summary

Adds an explicit guard at the start of `tripValidator.validate()` to reject requests where `req.params.tripId ?? req.params.id` evaluates to `null`, `undefined`, or an empty string.

**Issue:** Closes #14337

**Changes:**
- `backend/api/src/middleware/tripValidator.js`: Return 400 with `{ error: "tripId parameter is required" }` when tripId is null, undefined, or empty

---

Please assign this to the `tmdeveloper007` account

GSSOC (GirlScript Summer of Code)

### DIFF
--- a/backend/api/src/lib/otpHashing.js
+++ b/backend/api/src/lib/otpHashing.js
@@ -50,13 +50,16 @@ export function verifyOtpHash(otp, otpRecord) {
 // === Spec 12: constant-time hex compare ===
 export function constantTimeEqualHex(a, b) {
   if (typeof a !== 'string' || typeof b !== 'string') return false;
-  if (a.length !== b.length) return false;
-  if (a.length === 0 && b.length === 0) return true;
-  // Buffer.from(value, 'hex') does not reject invalid hex: it truncates at the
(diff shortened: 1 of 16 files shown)


KanishJebaMathewM:
level:advanced

PR #14586, opened 2026-08-15
fix(escortWalletController): use caller-scoped client for credential subject lookup (closes #14425)

### LINKED ISSUE #14425: fix(escortWalletController): anon client reads profiles (privileges revoked) - credential issuance always 500
## Problem

`resolveCredentialSubject()` — the resource loader for `requirePolicy('escort:issue-credential')` — reads `profiles` through the shared **anon** client:

```js
// backend/api/src/controllers/escortWalletController.js:14-18
const { data: profile, error } = await supabase
  .from('profiles')
  .select('polygon_wallet_address')
  .eq('id', req.user.id)
  .maybeSingle();
```

`supabase` is the anon client (`config/db.js`). `profiles` has **ALL anon privileges revoked** (`supabase/migrations/revoke_anon_privileges.sql:5`). PostgREST denies anon access at the privilege layer, so the query always returns a `permission denied` error → `AppError('Failed to load profile', 500)`.

## Impact

Every `POST /api/escorts/wallet/credential` (even for admins bypassing ownership) returns HTTP 500. Credential issuance is entirely broken.

## Expected fix

Read with the caller-scoped client: `createUserClient(req.token)` (imported from `config/db.js`). This also lets the caller's RLS ownership policy apply instead of bypassing it with the service role.

## Tests

`test/unit/escortWalletController.test.js` and `test/integration/escortWalletCredential.test.js` exist but mock the DB with a success-only stub, so the production failure is not caught.

Akshita-2307:
## Summary

`resolveCredentialSubject()` — the ownership pre-check for credential issuance — read `profiles` through the shared **anon** client:

```js
const { data: profile, error } = await supabase
    .from('profiles')
    .select('polygon_wallet_address')
    .eq('id', req.user.id)
    .maybeSingle();
```

`profiles` has **ALL anon privileges revoked** (`supabase/migrations/revoke_anon_privileges.sql:5`), so PostgREST denies the read regardless of RLS. Every credential issuance request — `POST /escort-wallet/issue-credential` for the driver's own wallet — hit `error` and returned **HTTP 500 `Failed to load profile`**. Drivers could never obtain escrow credentials.

## Changes

`backend/api/src/controllers/escortWalletController.js` — use the caller-scoped client:

- `createUserClient(req.token)` instead of the shared `supabase` anon client.
- The caller's RLS ownership policy (`Profiles select own profile`) scopes the read to `req.user.id`, so the ownership check still resolves the *authenticated driver's* wallet address.

## Verification

- `node --check` passes; `npx eslint` on the changed file is clean.
- `npx vitest run test/unit/escortWalletController.test.js` → **6 passed** (same as baseline).
- Pattern matches the rest of the codebase (`createUserClient(req.token)` is the standard for per-caller DB access in authenticated controllers).

## GSSOC

**Contributor:** Akshita-2307

Closes #14425

### DIFF
--- a/backend/api/src/controllers/escortWalletController.js
+++ b/backend/api/src/controllers/escortWalletController.js
@@ -1,7 +1,7 @@
 import didService from '../../../did/did.service.js';
 import logger from '../middleware/logger.js';
 import { AppError } from '../utils/errors.js';
-import { supabase } from '../config/db.js';
+import { createUserClient } from '../config/db.js';
 
 /**
  * Resolve the resource for the 'escort:issue-credential' ownership check:
@@ -11,7 +11,7 @@ import { supabase } from '../config/db.js';
  * wallet address/DID.
  */
 export const resolveCredentialSubject = async (req) => {
-    const { data: profile, error } = await supabase
+    const { data: profile, error } = await createUserClient(req.token)
         .from('profiles')
         .select('polygon_wallet_address')
         .eq('id', req.user.id)


Akshita-2307:
Superseded by #14520 which merged the same fix into main. Closing as duplicate.

PR #14587, opened 2026-08-15
fix(maintenancePhotoController): read ticket with caller-scoped client (closes #14426)

### LINKED ISSUE #14426: fix(maintenancePhotoController): anon client reads truck_maintenance_tickets - photo uploads always 500
## Problem

`uploadMaintenancePhotos()` verifies the ticket through the shared **anon** client while the RPC immediately below it correctly uses the caller-scoped client:

```js
// backend/api/src/controllers/maintenancePhotoController.js:49
const userClient = createUserClient(req.token);
...
// line 67-71
const { data: ticket, error: ticketError } = await supabase
  .from('truck_maintenance_tickets')
  .select('id, driver_id, photo_urls')
  .eq('id', ticketId)
  .maybeSingle();
```

`truck_maintenance_tickets` has **ALL anon privileges revoked** (`supabase/migrations/revoke_anon_privileges.sql:10`). The anon query always fails with `permission denied` → `ticketError` → HTTP 500 `Failed to verify ticket`.

## Impact

Every `POST /api/maintenance/:ticketId/photos` returns 500 — no driver can upload maintenance photos, regardless of a valid ticket.

## Expected fix

Use the `userClient` already constructed at line 49 for the ticket read (one-line swap). This matches the SECURITY DEFINER RPC comment: the caller JWT is required.

## Tests

`test/unit/maintenancePhotoController.test.js` and `test/integration/maintenancePhotoRoutes.test.js` exist but mock the DB, so the failure is not caught.

Akshita-2307:
## Summary

`uploadMaintenancePhotos()` read the maintenance ticket through the shared **anon** client even though a caller-scoped client was already built:

```js
const userClient = createUserClient(req.token);        // L49 — already available
...
const { data: ticket, error: ticketError } = await supabase   // L67 — anon
```

`truck_maintenance_tickets` has **ALL anon privileges revoked** (`supabase/migrations/revoke_anon_privileges.sql:8`), so PostgREST denies the read → `ticketError` → **HTTP 500 on every maintenance photo upload**, including the success path.

## Changes

- `backend/api/src/controllers/maintenancePhotoController.js` — the ticket ownership check now uses the existing `userClient` (`createUserClient(req.token)`), so the caller's RLS ownership policy applies instead of the privilege-revoked anon role.
- `backend/api/test/unit/maintenancePhotoController.test.js` — added `createUserClient: vi.fn(() => dbMock.supabase)` to the `db.js` mock so the controller's new call path is exercised (keeps the existing `from` mocks working).

## Verification

- `node --check` passes; `npx eslint` on both changed files is clean.
- `npx vitest run test/unit/maintenancePhotoController.test.js` → 1 passed / 6 failed, **identical failure set to baseline on `main`** (verified via `git stash`; the 6 failures are pre-existing and unrelated to the ticket read change).

## GSSOC

**Contributor:** Akshita-2307

Closes #14426

### DIFF
--- a/backend/api/src/controllers/maintenancePhotoController.js
+++ b/backend/api/src/controllers/maintenancePhotoController.js
@@ -64,7 +64,7 @@ export async function uploadMaintenancePhotos(req, res) {
     }
 
     // Verify ticket exists and belongs to this driver
-    const { data: ticket, error: ticketError } = await supabase
+    const { data: ticket, error: ticketError } = await userClient
       .from('truck_maintenance_tickets')
       .select('id, driver_id, photo_urls')
       .eq('id', ticketId)
--- a/backend/api/test/unit/maintenancePhotoController.test.js
+++ b/backend/api/test/unit/maintenancePhotoController.test.js
@@ -8,6 +8,7 @@ const { dbMock, docMock, scanMock } = vi.hoisted(() => ({
 
 vi.mock('../../src/config/db.js', () => ({
   get supabase() { return dbMock.supabase; },
+  createUserClient: vi.fn(() => dbMock.supabase),
 }));
 
 vi.mock('../../src/middleware/logger.js', () => ({


Akshita-2307:
Superseded by #14518 which merged the same fix into main. Closing as duplicate.

PR #14588, opened 2026-08-15
fix(profileService): read profiles and driver_details with service-role client (closes #14427)

### LINKED ISSUE #14427: fix(profileService): anon client reads revoked profiles/driver_details - profile and driver-details lookups fail
## Problem

`profileService` performs every read/write through the shared **anon** client:

```js
// backend/api/src/services/profileService.js
getProfile:      supabase.from('profiles').select('*').eq('id', userId)          // line 33-37
getDriverDetails:supabase.from('driver_details').select('*').eq('user_id', ...) // line 120-124
createProfile:   supabase.from('profiles').insert(...)                          // line 143
updateProfile:   supabase.from('profiles').update(...).eq('id', userId)         // line 152
```

Both `profiles` and `driver_details` have **ALL anon privileges revoked** (`supabase/migrations/revoke_anon_privileges.sql:5-6`). Every query returns `permission denied`, so `getProfile`/`getDriverDetails` throw, and `createProfile`/`updateProfile` fail.

## Impact

`GET /api/profile`, `GET /api/profile/driver-details` and profile creation/update all fail (500 or unhandled rejection) for every user. `profileService.js` already imports `supabaseAdmin` but never uses it.

## Expected fix

Route the four call sites through `supabaseAdmin` (the service has no `req`/token, so the service-role client is the correct available client — same pattern as the repository layer).

## Tests

`test/unit/profileService.test.js` exists but mocks the DB, so the failure is not caught.

Akshita-2307:
## Summary

`profileService` read and wrote `profiles` / `driver_details` through the shared **anon** client, even though `supabaseAdmin` was already imported (only `getCustomerStats` used it):

- `getProfile()` — anon `profiles` read (was already the cache-seeding path)
- `getDriverDetails()` — anon `driver_details` read
- `createProfile()` / `updateProfile()` — anon `profiles` insert/update

Both tables have **ALL anon privileges revoked** (`supabase/migrations/revoke_anon_privileges.sql:5,6`), so PostgREST returned `permission denied` on every call. This is service-layer code with no caller token, so an ownership-scoped client isn't possible — the service-role client is the only correct choice.

## Changes

- `backend/api/src/services/profileService.js` — all four functions now use `supabaseAdmin` (guard, query, insert, update). `getCustomerStats` drops its now-unnecessary `supabaseAdmin || supabase` fallback. Error messages reference `SUPABASE_SERVICE_ROLE_KEY` instead of `SUPABASE_ANON_KEY`.
- `backend/api/test/unit/profileService.test.js` — the `db.js` mock now exposes `supabaseAdmin` as the same mock client (the tests model "service-role available"), replacing the old "falls back to anon" contract that no longer exists.

## Verification

- `node --check` passes; `npx eslint` on both changed files is clean.
- `npx vitest run test/unit/profileService.test.js` → **27 passed** (the "not configured" fail-fast paths are still covered by setting the mock client to `null`).

## GSSOC

**Contributor:** Akshita-2307

Closes #14427

### DIFF
--- a/backend/api/src/services/profileService.js
+++ b/backend/api/src/services/profileService.js
@@ -1,4 +1,4 @@
-import { supabase, supabaseAdmin } from '../config/db.js';
+import { supabaseAdmin } from '../config/db.js';
 import { measureExecution } from '../core/performanceMetrics.js';
 import {
   getCachedSupabaseProfile, setCachedSupabaseProfile,
@@ -14,8 +14,8 @@ function isCacheEnabled() {
 
 export async function getProfile(userId) {
   return measureExecution('ProfileService.getProfile', async () => {
-  if (!supabase) {
-    throw new Error('Supabase client not configured — check SUPABASE_URL and SUPABASE_ANON_KEY');
+  if (!supabaseAdmin) {
+    throw new Error('Supabase client not configured — check SUPABASE_URL and SUPABASE_SERVICE_ROLE_KEY');
   }
 
   if (isCacheEnabled()) {
@@ -30,7 +30,7 @@ export async function getProfile(userId) {
     }
   }
 
-  const { data, error } = await supabase
+  const { data, error } = await supabaseAdmin
     .from('profiles')
     .select('*')
     .eq('id', userId)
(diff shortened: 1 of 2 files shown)


Akshita-2307:
Superseded by #14515 which merged the same fix into main. Closing as duplicate.

PR #14600, opened 2026-08-15
fix(profileRoutes): use caller-scoped and service-role clients instead of anon (closes #14432)

### LINKED ISSUE #14432: fix(profileRoutes): anon client on revoked profiles/driver_details/trucks/orders - core profile endpoints 500
## Problem

`profileRoutes.js` imports `createUserClient` (used at line 237) but most handlers read/write RLS-protected tables through the shared **anon** `supabase` client:

- `profiles` — lines 294, 303, 382, 480, 592, 723, 733
- `driver_details` — lines 318, 392
- `trucks` — line 404
- `orders` — lines 560, 773

All four tables have **ALL anon privileges revoked** (`supabase/migrations/revoke_anon_privileges.sql:5,6,8,14`). PostgREST denies anon access at the privilege layer, so each of these calls returns `permission denied`.

## Impact

Core profile endpoints fail for every authenticated user:

- `PUT /api/profile` (profile + driver_details + truck plate update) → 500
- `PUT /api/profile/wallet` → 500
- `GET /api/profile/driver/statement` (reads `orders`) → 500
- Other profile reads/writes → 500

## Expected fix

Within each route handler, swap the bare `supabase` reads/writes to `createUserClient(req.token)` (the same client already proven at line 237), so the caller's RLS ownership policies apply.

## Tests

`test/integration/profileRoutes.test.js` exists but mocks the DB, so the failure is not caught.

Akshita-2307:
## Summary

`profileRoutes.js` imported `createUserClient` (used only at L237) but most handlers read/wrote RLS-protected tables through the shared **anon** `supabase` client. `revoke_anon_privileges.sql` revokes ALL anon privileges on every table they touched:

- `profiles` (revoked at L5) — PUT /wallet check+update, PUT / (main profile), PUT /fcm-token, GET /driver/statement profile lookup, admin cache lookup
- `driver_details` (L6) — PUT /wallet driver sync, PUT / is_online
- `trucks` (L8) — PUT / number plate
- `orders` (L14) — GET /driver/statement and GET /driver/performance-stats
- `trips` (L18), `ratings` (L22) — GET /driver/performance-stats

PostgREST denies anon at the privilege layer (RLS can't help), so every one of these calls returned `permission denied` → the affected endpoints always returned **HTTP 500** for authenticated users.

## Changes

`backend/api/src/routes/profileRoutes.js` — each handler now uses the correct client:

- **Caller-scoped** `createUserClient(req.token)` (RLS applies) for user routes where the handler operates on `req.user.id`:
  - PUT /wallet (profiles check/update + driver_details upsert)
  - PUT / (profiles update + driver_details is_online + trucks number_plate)
  - PUT /fcm-token (profiles update)
  - GET /driver/statement (orders paging + profiles lookup)
  - GET /driver/performance-stats (orders + trips + ratings)
- **Service-role** `supabaseAdmin` for `DELETE /admin/cache/:userId`, which deliberately resolves *any* user's profile by id or firebase_uid — caller-scoped RLS would hide other users.
- Removed the now-unused `supabase` import.

## Verification

- `node --check` passes; `npx eslint` on the changed file is clean.
- `npx vitest run test/unit/profileRoutes.test.js` → **2 passed** (unchanged).
- `test/integration/profileRoutes.test.js` fails to *collect* on `main` too — its helper `test/helpers/supabaseMock.js:498` contains invalid JS (unrelated, untouched).

## GSSOC

**Contributor:** Akshita-2307

Closes #14432

### DIFF
--- a/backend/api/src/routes/profileRoutes.js
+++ b/backend/api/src/routes/profileRoutes.js
@@ -104,7 +104,7 @@ import {
   getCustomerStats,
   getDriverDetails
 } from '../services/profileService.js';
-import { supabase, createUserClient } from '../config/db.js';
+import { supabaseAdmin, createUserClient } from '../config/db.js';
 import { ProfileModel } from '../models/ProfileModel.js';
 import { invalidateCachedProfile, invalidateCachedSupabaseProfileAll } from '../lib/profileCache.js';
 import { auditLog } from '../middleware/auditLog.js';
@@ -290,7 +290,8 @@ router.put('/wallet', authenticate, userLimiter, validateBody(updateWalletSchema
   }
 
   try {
-    const { data: existing, error: checkErr } = await supabase
+    const userDb = createUserClient(req.token);
+    const { data: existing, error: checkErr } = await userDb
       .from('profiles')
       .select('polygon_wallet_address')
       .eq('id', userId)
@@ -299,7 +300,7 @@ router.put('/wallet', authenticate, userLimiter, validateBody(updateWalletSchema
     if (checkErr) return res.status(500).json({ error: 'Failed to fetch profile.', details: checkErr.message });
     if (!existing) return res.status(404).json({ error: 'Profile not found.' });
 
-    const { error: updateErr } = await supabase
+    const { error: updateErr } = await userDb
       .from('profiles')
       .update({
         polygon_wallet_address: normalized,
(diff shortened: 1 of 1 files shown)


Akshita-2307:
Superseded by #14545 which merged the same fix into main. Closing as duplicate.

PR #14749, opened 2026-08-15
fix(idempotency): remove duplicate pendingCache declaration

### LINKED ISSUE #14740: fix(idempotency): pendingCache referenced outside Redis cache block scope
## Problem

In `backend/api/src/middleware/idempotency.js`, the `pendingCache` variable is declared at line 165 **inside** the Redis cache block:

```js
if (useRedis) {
  ...
  let pendingCache = null;
  ...
}
```

but it is assigned/read at line 218 in the shared `res.json` override that runs for **every** request mode (Redis **and** in-memory):

```js
const originalJson = res.json.bind(res);
res.json = function (body) {
  if (pendingCache) {
    clearTimeout(pendingCache);
    pendingCache = null;
  }
  ...
};
```

Because the declaration is scoped to the Redis `if` block, any 2xx response that goes through the override throws `ReferenceError: pendingCache is not defined` whenever the server is running with the in-memory fallback (Redis outage) — and even in Redis mode the reference is technically out of scope.

## Fix

Hoist the declaration to the outer (shared) scope next to the other module state (e.g. just before `let responded = false;`) and remove the inner declaration inside the Redis block:

```js
let pendingCache = null;
let responded = false;
```

Xenon010101:
Closes #14740

## Problem

`backend/api/src/middleware/idempotency.js` declares `pendingCache` twice with `let` in the same function scope:

- line 101 (`let pendingCache = null;` next to the Redis cache lookup) and
- line 202 (`let pendingCache = null;` before the shared `res.json` override).

This is a hard **Parsing error: Identifier 'pendingCache' has already been declared** — the module cannot even be loaded. (Also confirmed by ESLint.)

## Fix

Remove the redundant inner declaration at line 101. The shared-scope declaration at line 202 covers every usage: the `res.json` override that clears the timeout and caches the response, and the `finalize` helper.

Verified with `node --check` and ESLint on the file.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request processing stability by removing redundant internal state initialization.
  * Preserved response caching and lock handling behavior during request finalization.

### DIFF
--- a/backend/api/src/middleware/idempotency.js
+++ b/backend/api/src/middleware/idempotency.js
@@ -101,8 +101,6 @@ export function requireIdempotency(ttlSeconds = 3600) {
         return res.status(cached.statusCode).json(cached.body);
       }
 
-      let pendingCache = null;
-
       if (redisClient) {
         const lockKey = `${key}:lock`;
         const lockAcquired = await redisClient.set(lockKey, '1', 'NX', 'PX', LOCK_TTL_MS);


Xenon010101:
gssoc

PR #15281, opened 2026-09-13
fix(kafka): use service-role client for protected event stores

### LINKED ISSUE #9202: bug: Kafka event store + CQRS read model query events/order_read_models with anon-key client while tables are service_role-only — read model never builds
﻿## Bug

The Kafka CQRS event store and read model run on the **anon-key** Supabase client (`supabase` from `api/src/config/db.js`) against the `events` and `order_read_models` tables, which migration `20260805090000_create_kafka_event_tables.sql` grants to **service_role only** and explicitly revokes from `anon, authenticated`.

Consequence: every `event.repository.js` and `order.read.model.js` query is denied by RLS, so the read model is **never built or updated** — the kafka service consumes order events as a silent no-op.

## Evidence

- `backend/kafka/repositories/event.repository.js:1` — `import { supabase } from '../../api/src/config/db.js';` (anon-key client) used for all `events` queries (`saveEvent`, `getEventsByOrderId`, `getEventById`, `getEventStats`).
- `backend/kafka/cqrs/order.read.model.js:1` — same anon `supabase` import, used to read `events` (`buildReadModel` → `eventRepository.getSnapshot`) and upsert into `order_read_models` (`updateReadModel`).
- `backend/kafka/index.js:5,21,27,33,39` — live service entry (`npm start`) registers handlers for `order.created` / `order.updated` / `driver.assigned` / `payment.confirmed` that call `orderReadModel.buildReadModel(orderId)`. This is the only call path; it never succeeds.
- Migration `supabase/migrations/20260805090000_create_kafka_event_tables.sql`:
  - lines 65-66: `alter table events enable row level security;` / `alter table order_read_models enable row level security;`
  - lines 68-81: policies grant `for all to service_role ... using (true)`.
  - lines 83-84: `revoke all on table events from anon, authenticated;` / `revoke all on table order_read_models from anon, authenticated;`
- `db.js` confirmation: `supabase = createClient(supabaseUrl, supabaseAnonKey)` (RLS enforced), `supabaseAdmin = createClient(..., supabaseServiceKey)`.

## Impact

- `order_read_models` stays empty forever; `getOrderReadModel`, `getAllOrdersReadModel`, `getOrderStats` (and any consumer) return null/empty even though events are consumed.
- Because `getSnapshot` swallows errors and returns `null`, failures are silent — no DLQ entries, no logs beyond the caught error. The read model feature is dead-on-arrival with no observable signal.

## Contrast with prior fix

Issue #6288 covered the same anon-vs-service_role class for `kafka_processed_events` (via `processedEvent.repository.js`); that file was fixed to import `supabaseAdmin` (`backend/kafka/repositories/processedEvent.repository.js:1`). `deadLetter.repository.js:1` likewise uses `supabaseAdmin`. Only `event.repository.js` and `order.read.model.js` were missed.

## Suggested fix

Switch both files to `import { supabaseAdmin }` (service-role client), matching `processedEvent.repository.js` and `deadLetter.repository.js`, and the service-role-only RLS on both tables.

profmohit-edu:
## Description
Fixes #9202 by routing Kafka event-store and CQRS/read-model access to the service-role Supabase client required by the tables' RLS policies.

## Type of Change
- Bug fix / correctness and access-control boundary repair
- Focused regression coverage

## What changed
- Use `supabaseAdmin` for protected `events` access.
- Use the injected service-role client for protected CQRS/read-model operations.
- Add regression tests that fail if these paths fall back to the anonymous client.

## Related Issues
Closes #9202

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved backend data access reliability for order read models and event records by consistently using authorized service-level access.
  - Order lookups, event retrieval, statistics, and read-model operations now behave consistently in environments with restricted access permissions.

- **Tests**
  - Added coverage to verify authorized access is used and unauthorized database access is not attempted for order and event operations.


profmohit-edu:
Addressed in the current branch: the save-event test mock now exports `supabaseAdmin`, matching `EventRepository.saveEvent`, while preserving the existing query chain. The focused service-role save-path regression test is also present and asserts that the anonymous client is not used.

PR #15636, opened 2026-09-16
fix: correct TomTom traffic surge multiplier (#13306)

### LINKED ISSUE #13306: trafficService TomTom branch inverts the surge multiplier: congestion returns 1.0x and free-flow returns 1.2x+
**Severity:** High - live-traffic pricing is backwards for TomTom-sourced traffic data, inflating prices on empty roads and failing to price congestion.

**Location:** `backend/api/src/services/trafficService.js:29-30` (TomTom branch) vs `:38-41` (Google branch)

```js
const speedDiff = data.flowSegmentData?.speedDiffPercent || 0;
multiplier = Math.min(MAX_SURGE_MULTIPLIER, Math.max(1.0, 1.0 + Math.max(0, speedDiff / 100)));
```

Per the TomTom Flow API, `speedDiffPercent` is **negative** when the current speed is below the free-flow speed, i.e. negative values mean congestion and positive values mean free-flow. `Math.max(0, speedDiff / 100)` therefore clamps the congested case to `0`, so the multiplier is exactly `1.0` under heavy congestion, while on a clear road the multiplier rises to `1.2x+` - free-flow traffic is billed as premium while gridlock is charged the base rate.

Compare the Google branch (`trafficService.js:41`):

```js
multiplier = Math.min(MAX_SURGE_MULTIPLIER, Math.max(1.0, duration / normalDuration));
```

where `duration_in_traffic / duration` is always >= 1 and grows with congestion - the correct direction. The two data sources disagree, and the TomTom path is the outlier.

**Impact:** the multiplier feeds truck-search pricing at `backend/api/src/routes/truckRoutes.js:549` (`getLiveTrafficMultiplier`), so price estimates are systematically wrong for any region resolved via TomTom.

**Suggested fix:** drop the clamp, e.g. `const multiplier = 1.0 + (1.0 - current / freeFlow) * surgeScale;` so slower-than-free-flow traffic raises the multiplier, and add a unit test asserting congestion -> multiplier > 1 and free-flow -> multiplier ~= 1.

bharath19k:
## Summary
- Correct TomTom traffic surge multiplier calculation so negative speedDiffPercent values correctly represent congestion.
- Keep positive speedDiffPercent at the base 1.0x multiplier.
- Preserve the 2.5x maximum surge cap.
- Update existing TomTom tests and add regression coverage for congestion, free-flow traffic, and the cap.

## Testing
- TomTom-focused tests: 6 passed
- speedDiffPercent-focused tests: 4 passed
- services/trafficService.test.js: 9 passed
- Full trafficService.test.js: 17 passed, 5 pre-existing unrelated failures
- git diff --check: passed
- Node.js mathematical assertions: passed

The 5 unrelated existing failures were not modified.

Fixes #13306

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected traffic-based multiplier calculations so slower traffic increases the multiplier appropriately.
  * Faster-than-normal traffic now leaves the multiplier at the standard value.
  * Added safeguards to keep the multiplier within the supported range, including for severe congestion.

* **Tests**
  * Expanded coverage for free-flow traffic, moderate congestion, and maximum congestion limits.


bharath19k:
The Backend CI failure appears to be pre-existing and unrelated to this PR.

The failing run stops during Vitest setup with:

ReferenceError: module is not defined in ES module scope
test/mocks/redisMock.js:62

147 test files fail before any tests execute because redisMock.js uses CommonJS module.exports while backend/api is configured with "type": "module".

This PR only changes trafficService.js and trafficService.test.js. The TomTom-specific tests pass locally (6/6), and the related traffic service tests pass (9/9).

No changes were made to the unrelated Redis test infrastructure.

PR #15681, opened 2026-09-16
fix: prevent stale driver locations in public tracking (#15395)

### LINKED ISSUE #15395: [Bug] Public tracking can return stale driver locations
**Describe the bug**

The public tracking flow can return an outdated driver location when the driver's latest `driver_locations` record is still marked as active but has not been updated for an extended period.

`TrackingTokenService.getDriverLocation()` currently filters by the tracked driver's ID and `is_active = true`, then returns the most recent row. There is no server-side freshness cutoff, so an old location can still be presented to customers as the driver's current position.

This becomes particularly misleading when driver location updates stop temporarily. The public tracking endpoint may continue returning the last known coordinates even though they are no longer recent enough to represent the driver's actual location.

**To Reproduce**

1. Associate an order with a driver.
2. Add or retain an active `driver_locations` record for that driver.
3. Allow the location record's `last_updated_at` timestamp to become older than the intended tracking freshness window.
4. Request the driver's location through the public tracking flow.
5. Observe that the stale location is still returned.

**Expected behavior**

Public tracking should only return a driver location when the active location record is recent enough to be considered valid.

The existing backend uses a 15-minute driver-location freshness policy when selecting target drivers. Public tracking should apply the same freshness boundary so stale telemetry is not presented as current location data.

When no sufficiently fresh active location exists, the public tracking flow should return no driver location rather than exposing the stale coordinates.

**Screenshots**

Not applicable. This is a backend data-freshness issue.

* OS: N/A
* App Version: Current development branch
* Browser: N/A

**Additional context**

The affected lookup is in `backend/api/src/services/trackingTokenService.js`.

The repository already defines a 15-minute driver-location freshness policy in the backend. Reusing that existing policy keeps public tracking behavior consistent with the rest of the driver-location selection flow and avoids introducing a separate freshness threshold.

This issue is limited to the freshness of the public tracking read path. It is separate from driver reassignment/order-scoping concerns and from client-side GPS caching or transmission behavior.

rehan1020:
## Summary
Fixes #15395 — Public tracking endpoint could return arbitrarily old driver coordinates as 'current' because the freshness filter used a hardcoded constant with no env override.

### Source: `trackingTokenService.js`
- **Env-configurable freshness constant**: Replaced hardcoded `DRIVER_LOCATION_FRESHNESS_MS = 15 * 60 * 1000` with `PUBLIC_TRACKING_LOCATION_FRESHNESS_SECONDS` (default 900s, configurable via env var).
- **Updated freshness cutoff computation** to use the new seconds-based constant.
- **Removed dead legacy CommonJS block** (lines 328-395) that used `require()` and crashed the ESM module at load time. These functions were unreferenced outside the file.

### Tests: `trackingTokenService.test.js`
- **Fixed 3 pre-existing broken tests**: Added `order.id` and active trip seed rows so `getDriverLocation()` exercises real behavior instead of coincidentally returning null.
- **Added regression test**: Seeds a driver location 20 minutes old with a valid order + active trip, asserts `getDriverLocation()` returns `null`.
- Added `trips` store initialization in `beforeEach`.

### Tests: `trackingTokenService.orderScope.test.js`
- Added `gte()` method to mock query builder.
- Updated filter-application logic to handle `op === 'gte'`.

## Test Results
| Suite | Result |
|---|---|
| `trackingTokenService.test.js` | 31 passed |
| `trackingTokenService.orderScope.test.js` | 3 passed |
| `publicTrackingRoutes.test.js` | 4 passed, 2 failed (pre-existing, out of scope) |
| `integration/publicTracking.test.js` | 14 passed |
| `node --check trackingTokenService.js` | Passes |

## Summary by CodeRabbit

- **New Features**
  - Public driver-location freshness can now be configured through an environment setting.
  - The default freshness window remains 15 minutes (900 seconds).

- **Bug Fixes**
  - Driver locations older than the configured freshness window are no longer displayed, helping prevent stale location information from being shown.

- **Tests**
  - Added coverage for freshness limits and driver-location results scoped to the relevant time window.


rehan1020:
@KanishJebaMathewM Please see now, fixed it.

PR #15701, opened 2026-09-16
fix: GNN Route Optimizer Trains on Empty Datasets and Uses an Untrain…

### LINKED ISSUE #11249: GNN Route Optimizer Trains on Empty Datasets and Uses an Untrained Random Model for Live Routing
# Issue: GNN Route Optimizer Trains on Empty Datasets and Uses an Untrained Random Model for Live Routing

## Description

`backend/ml/routes/gnn_routes.py` exposes `/gnn/train`, `/gnn/optimize-route`, and `/gnn/multi-objective`, but the training endpoint never actually trains on data:

```python
@router.post("/train")
async def train_model(request: TrainRequest):
    # In production: load training data
    train_data = []
    val_data = []
    loss = optimizer.train(train_data, val_data, request.epochs)
    return {'success': True, 'data': {'loss': loss, 'epochs': request.epochs}, ...}
```

`GNNRouteModel` is initialized with random weights (`models.py:166-175`, `n(0,1)` default `torch.nn` init). Because no route ever loads real training data, `RouteOptimizer` is used with **random, untrained weights** for every live `/optimize-route` call:

1. **Training on empty set**: `optimizer.train( [], [], epochs)` (models.py:286-314):
   ```python
   avg_loss = total_loss / len(train_data)   # ZeroDivisionError when train_data == []
   ```
   Fails every time with a contextless `ZeroDivisionError` — the endpoint can't even graduate from "crash" to "do something".

2. **Deterministic random behavior**: every new process re-randomizes weights, so identical route requests across replicas return different, meaningless optimizations. The `weights` dict (`{'time':1.0,'cost':0.5,'fuel':0.3}`) is hard-coded rather than learned.

3. **No loading of real model state**: `/gnn/model/load` (line 216-231) bakes a `path` default and loads whatever file exists; there is no wiring from a trained checkpoint to the serving `optimizer` object, so even if training worked the serving instance wouldn't see the trained weights.

4. **No validation**: `start_node`/`end_node` are unchecked — a nonexistent node crashes `optimize_route` with a KeyError, and the resulting `success:false`/500 hides the real cause.

## Expected Behavior

- `/gnn/train` must load real road-network data (nodes, edges, targets) from storage/ML pipeline and return a valid epoch-loss curve.
- Live routing must either (a) require a trained checkpoint (`model_path`) or (b) explicitly refuse to serve when the model is untrained.
- Deterministic behavior: same inputs → same route; fail clearly when a requested node doesn't exist.

## Actual Behavior

- `/gnn/train` always 500s on empty data; no model is ever trained.
- `/gnn/optimize-route` and `/gnn/multi-objective` return results from a random model, producing unreliable "optimized" routes that vary per restart.
- Empty-ended dead-path routing silently returns empty `route` arrays that downstream consumers interpret as valid routes.

## Proposed Fix

1. Wire `train_data`/`val_data` to a real dataset loader (e.g. from `traffic_pipeline` or a DB/parquet source); raise `HTTPException(422)` when empty rather than dividing by zero, with a clear message.
2. Guard serving: in `RouteOptimizer.optimize_route`, verify `self.model` was trained (track a `trained` flag, or a checkpoint hash); return `503`/`409` otherwise.
3. Validate `start_node`/`end_node` exist in the graph before running the search; return a structured 404 if not.
4. Load the trained checkpoint at startup via `RouteOptimizer(model_path=...)` and only fall back to random weights when explicitly flagged (dev mode).
5. Add tests: train on synthetic-but-nonempty data, assert deterministic routes, assert 404 for unknown nodes, assert 503 for untrained serving.

## Files to Modify

- `backend/ml/routes/gnn_routes.py` — real dataset loader; empty-set 422; node-existence 404; untrained-serving guard.
- `backend/ml/gnn/models.py` — `train()` safe division + `trained` flag; return structured errors.
- `backend/ml/requirements.txt` — data-loading deps if a dataset store is added.

## Severity

**High** — live freight routing returns random results; operators cannot distinguish "optimized" from noise, and every training attempt crashes.

Aditya8369:
Summary of Changes:
Untrained Model Guard (RouteOptimizer & Routes):

Added is_trained boolean state and allow_untrained flag to RouteOptimizer
.
Guarded optimize_route and multi_objective_optimization in RouteOptimizer to raise RuntimeError if invoked on an untrained model without explicit dev mode / allow_untrained=True.
In backend/ml/routes/gnn_routes.py
, /optimize-route and /multi-objective endpoints check optimizer.is_trained and return a structured HTTPException(503) when the model is not trained.
Graph Node Validation (404 Error Handling):

In backend/ml/routes/gnn_routes.py
, incoming requests validate that start_node and end_node exist in request.nodes before graph construction and pathfinding.
Missing nodes immediately return a structured HTTPException(404, detail="Node(s) not found in graph: ...").
Empty Dataset Validation (422 Error Handling):

In RouteOptimizer.train(), validated that train_data is non-empty (if not train_data: raise ValueError(...)), preventing ZeroDivisionError.
In /gnn/train, checked training dataset availability and returns HTTPException(422, detail="Training dataset is empty...").
After a successful training run or checkpoint load, optimizer.is_trained is set to True.
Checkpoint Loading at Startup:

backend/ml/routes/gnn_routes.py checks GNN_MODEL_PATH / default model checkpoint path on startup, loading weights via RouteOptimizer(model_path=...) if available, and respects DEV_MODE / ALLOW_UNTRAINED_GNN flags.
Unit & API Endpoint Tests:

In backend/ml/tests/test_gnn.py
:
test_untrained_optimizer_raises_runtime_error: Verifies untrained optimizer throws RuntimeError.

closes #11249

## Summary by CodeRabbit

* **New Features**
  * Route optimization now requires a trained model by default, with an explicit option to allow untrained models.
  * Added validation for missing start or destination nodes.
  * Model training supports optional batch and edge data and rejects empty datasets.
  * Successfully trained or loaded models are automatically marked as trained.

* **Bug Fixes**
  * Runtime errors now surface instead of being silently converted to empty results.
  * API responses now clearly report unknown nodes, invalid training data, and unavailable untrained models.


Aditya8369:
@KanishJebaMathewM solved conflicts

Aditya8369:
@KanishJebaMathewM solved conflicts

PR #16235, opened 2026-09-18
fix(index): register /api/tolls before 404 handlers and drop duplicat…

### LINKED ISSUE #16234: bug: /api/tolls/optimize returns 404 — the toll-optimization API is unreachable. to
Issue: /api/tolls/optimize returns 404 — the toll-optimization API is unreachable. tollOptimizationRouter is mounted at the bottom of src/index.js (app.use('/api/tolls', ...)), after app.use(notFound) and errorHandler. Since notFound responds 404 without next(), no request ever reaches the toll router. Separately, webhookRoutes is registered twice (/api/webhooks), double-registering the same router.

scriptedbyshivam:
Closes #16234
PR title: fix(index): register /api/tolls before 404 handlers and drop duplicate webhook mount
PR body:
## What
- Move `app.use('/api/tolls', tollOptimizationRouter)` from the bottom of
  `src/index.js` (after `notFound`/`errorHandler`) into the main REST mount
  block, so `/api/tolls/optimize` is actually reachable and sits behind the
  full middleware chain.
- Remove the duplicate `app.use('/api/webhooks', webhookRoutes)` — the router
  is already mounted once in the dedicated WEBHOOK ROUTES section.

## Why
`notFound` responds 404 for every unmatched request, so the toll endpoint was
completely unreachable (always 404). The webhooks router was being registered
twice for the same path.

Note: the trailing toll mount overlaps the identical line kept at the end of
the `fix/indexjs-remove-duplicate-error-middleware` PR — after both merge,
keep only the mount moved into the REST block.

## Summary by CodeRabbit

- **New Features**
  - Added access to toll optimization services through the API.
  - Improved webhook routing for Web3 integrations.

- **Bug Fixes**
  - Requests with oversized payloads now receive a clear error response.
  - Malformed JSON requests now return an appropriate client error instead of causing less-specific failures.


scriptedbyshivam:

PR #16245, opened 2026-09-18
fix(ml): prevent production model corruption by isolating training to a deepcopy (fixes #11239)

### LINKED ISSUE #11239: Transformer Training Endpoints Fit Production Models on Random Synthetic Noise, Silently Destroying Forecast Quality
# Issue: Transformer Training Endpoints Fit Production Models on Random Synthetic Noise, Silently Destroying Forecast Quality

## Description

`backend/ml/routes/transformer_routes.py` exposes `/transformer/{demand,traffic,price}/train`. Each `TrainRequest` is served by a **single global trainer that owns the exact model instance used by the live forecast endpoints**:

```python
# module level
demand_model = DemandForecastTransformer()
traffic_model = TrafficForecastTransformer()
price_model = PriceForecastTransformer()
demand_trainer = TransformerTrainer(demand_model)
traffic_trainer = TransformerTrainer(traffic_model)
price_trainer = TransformerTrainer(price_model)
```

The training endpoints then fit those *same* shared model objects on **synthetic random noise** (`torch.randn`):

```python
@router.post(/demand/train)
async def train_demand(request: TrainRequest):
    train_data = torch.randn(1000, demand_model.transformer.seq_len, demand_model.input_dim)  # pure noise
    train_labels = torch.randn(1000, demand_model.transformer.pred_len)                        # pure noise
    ...
    results = demand_trainer.train(train_data, train_labels, epochs=request.epochs, ...)
    return {'success': True, 'data': results, ...}
```

Because `TransformerTrainer` holds a reference to the same `nn.Module` that `forecast_demand` runs inference on, each `/train` call **overwrites the production model weights** in place via `train_step` → `optimizer.step()`. `forecast_demand` runs `demand_trainer.predict` against the *post-train* (now noise-fitted) weights.

**Critical problems**:

1. **Weights replaced by a noise-fitted model.** Training against `y = randn()` maximizes (or randomizes) the loss surface with zero signal. Any earlier trained/validated weights are destroyed the moment someone calls `/demand/train`. There is no separation between the training artifact and the serving artifact — they are literally the same tensor buffers.

2. **`train()` has no `torch.no_grad()` isolation and returns `avg_loss` unconditionally.** `TransformerTrainer.train` (model.py:282-328) computes `num_batches = (train_data.size(0)+batch_size-1)//batch_size` and divides by it; with the empty-data branch (`train_data.size(0)==0 → batch loop never runs`) `len` divides by zero → `ZeroDivisionError`, and with noise it trains destructively.

3. **No locking / no versioning.** `train` and `forecast` hitting the same shared objects concurrently race: `train_step` calls `self.model.train()` (enables dropout) and mutates weights while `predict` is mid-forward. Forecasting under concurrent `/train` returns nondeterministic, corrupted outputs (see ISSUE-015's global-state class — but here the *trainer holds the serving model*, so even sequential calls irreversibly degrade serving).

4. **No auth / parameter bounds.** `epochs` and `batch_size` are taken straight from the request body. `epochs: 1e6`, `batch_size: 1`, or negative values are accepted; a huge `epochs` triggers a CPU/GPU-heavy burn, and `batch_size: 0` → division by zero in `num_batches`.

5. **Three endpoints, zero validation, zero tests.** All three (`demand`, `traffic`, `price`) share the identical destructively-trained pattern.

## Expected Behavior

- Training must write to a *separate* model artifact (new weights / new `TransformerTrainer`), never mutate the serving model. Serving stays frozen on the last validated checkpoint.
- `/train` must validate `0 < epochs` and `batch_size >= 1`, require auth, and be gated by an exclusive train lock (or run in a worker).
- After training, a new model version should be atomically swapped into serving; `/forecast` should be decoupled from training.

## Actual Behavior

- One `/demand/train` request permanently replaces production demand-model weights with a model fit on uniform random noise — forecasts for every subsequent request are garbage until someone manually re-deploys/re-trains on real data.
- Concurrent `/train` + `/forecast` produce racy nondeterministic predictions and can throw mid-forward errors (`self.model.train()` dropout / weight mutation during eval).
- `batch_size: 0` → `ZeroDivisionError`; huge `epochs` → resource exhaustion; all under no auth.

## Proposed Fix

1. Give the trainer its *own* model copy for training and a swap callback:
   ```python
   def train(self, ...):
       working = copy.deepcopy(self.model)
       ... train working ...
       self.model.load_state_dict(working.state_dict())  # atomic swap at end only
   ```
   or keep serving on `model_path`-loaded weights and only swap on successful validation (early-stop on `val_loss`).
2. Add Pydantic constraints on `TrainRequest` (`epochs: int = Field(..., ge=1, le=500)`, `batch_size: int = Field(..., ge=1, le=1024)`) and guard `num_batches` against zero.
3. Add an `asyncio.Lock` / worker queue around the `*_train` handlers so training and forecasting never interleave on the same tensors; enforce `torch.no_grad()` in serving `predict`.
4. Add auth + audit logging to the training endpoints, and a `/model/version` endpoint exposing the currently-serving version.
5. Tests: assert forecast is unchanged during training, then changes only after an explicit commit; assert `epochs=0`/`batch_size=0` rejected.

## Files to Modify

- `backend/ml/routes/transformer_routes.py` — bound training params, auth, locking, artifact swap.
- `backend/ml/transformers/model.py` — safe `num_batches`, immutable-train pattern, `no_grad` predict.
- `backend/ml/routes/mlRoutes.js` (`repo/backend/api/...` if it proxies `/transformer`) — ensure it forwards auth/session.

## Severity

**Critical** — a single unauthenticated request silently destroys the live forecasting models; all downstream ETA/pricing features consume garbage afterwards with no indication of corruption.

kazi716:
## Description
This PR addresses #11239 by fixing a critical architecture and data corruption vulnerability where transformer endpoints were overwriting the live production forecasting models with weights fitted on synthetic noise.

### Changes
- **`backend/ml/transformers/model.py`**:
  - Implemented immutable training via deep copy: `TransformerTrainer.train()` now strictly copies the model, trains on the copy, and performs an atomic `load_state_dict` swap only on successful completion.
  - Guarded against `ZeroDivisionError` in batch division and enforced valid sizing logic.
- **`backend/ml/routes/transformer_routes.py`**:
  - Introduced `asyncio.Lock()` per transformer type (demand, traffic, price) to guarantee concurrency safety between training loops and inference requests.
  - Added strict Pydantic constraint boundaries to `TrainRequest` using `Field(ge=1)` to filter out malformed or resource-exhausting hyperparameters.
  - Eliminated hardcoded `torch.randn` noise training sets from the endpoints, enforcing the use of actual `train_data` sent in the request.
- **`backend/ml/tests/test_transformers.py`**:
  - Wrote regression tests validating the Pydantic constraints and confirming that the model effectively shields inference from mutating during a live training step until the atomic swap.

## Type of Change
- [x] Security Fix / Structural Hardening

## Related Issues
Closes #11239

## Summary by CodeRabbit

- **New Features**
  - Model training requests now accept caller-provided training data and optional validation data.
  - Training parameters are validated with supported ranges of 1–500 epochs and batch sizes of 1–1,024.
  - Concurrent training requests for demand, traffic, and pricing models are handled safely.

- **Bug Fixes**
  - Training updates are applied atomically, helping prevent partially updated models if training fails.
  - Forecast failures now provide more specific endpoint-level error messages.


kazi716:
Thank you for assigning this issue! I have successfully opened the Pull Request fixing the catastrophic Transformer ML serving data-corruption vulnerability.

### Label Request & PR Analysis
Based on the official ECSoC26 points criteria, I am explicitly requesting the following labels:
- \ECSoC26\
- \Level 3\
- \good-backend\
- \good-pr\

**Justification for Level 3 (Core/Architecture/Performance):**
This PR physically redesigns the core machine learning data pipeline and model management lifecycle. Previously, training endpoints destructively mutated the live production model tensors using synthetic noise arrays without any form of resource gating or locking. 
This PR implements **Architectural Restructuring** by isolating the \TransformerTrainer\ context: it deeply copies the underlying PyTorch model prior to training, operates safely out-of-line, and utilizes an atomic tensor \state_dict\ swap upon validation, completely eliminating TOCTOU conditions during ML prediction. Furthermore, it implements **Security Hardening** by adding concurrency \syncio.Lock()\ guards per transformer and establishing strict Pydantic input limits, thus mitigating inference failures or GPU exhaustion.

Could you please review the PR and apply the appropriate \ECSoC26\ and XP labels? Thank you!

kazi716:
@coderabbitai[bot] All actionable comments have been addressed! 

- Updated `TrainRequest` with `@model_validator` to ensure `val_data` and `val_labels` are either both present or both absent.
- Moved tensor construction and training into `run_training_job` and wrapped `run_inference` in matching per-model locks inside the forecast endpoints.
- Added `is_training_cancelled()` check prior to the atomic state-dict commit in `TransformerTrainer.train`.
- Split `test_immutable_training_swap` in `test_transformers.py`, adding `test_immutable_training_swap_failure` with the `FailOnSecondLoss` criterion to assert no mutation on failure.
- Fixed `epochs < 1` validation before batch_size checking, and added `size(0)` equality assertions for both training and validation label pairs.
- `TransformerTrainer.train` now initializes `working_optimizer` with a deep-copied `state_dict` from the live optimizer to preserve accumulated moments.

PR #16443, opened 2026-09-20
fix(trips): stop accepting hardcoded demo OTP on orders that have a r…

### LINKED ISSUE #16442: bug: POST /:id/confirm-stop accepts '123456' even when the order sets a real delivery OTP,
trip OTP hardening · Issue: "POST /:id/confirm-stop accepts '123456' even when the order sets a real delivery OTP, and always accepts it as the default (root cause comment says the default was never enforced non-production-only)." ·

scriptedbyshivam:
Closes #16442
PR: exported verifyDeliveryOtp({expectedOtp, submittedOtp}) — real OTP requires exact match; without an OTP only '123456' in non-prod, rejected in prod. 12/12 pass.


scriptedbyshivam:

PR #16531, opened 2026-09-22
fix(services): repair parse-broken wallet and cross-dock services and…

### LINKED ISSUE #16530: bug: walletService and crossDockService fail to parse — cross-dock API is 500, and a merge-damaged orderLifecycleService test suite never ran
Three files in backend/api are payloads of earlier merge damage and none of
them compile. Because a service cannot even be imported, cross-dock
integration is entirely broken at runtime and a large test suite has been
silently skipped:

1. src/services/wallet/walletService.js — SyntaxError "Unexpected identifier
   'Ethereum'": the DomainError on line 14 embeds an unquoted string
   (`{ error: Invalid Ethereum/Polygon wallet address format: "". }`) and
   line 21 uses an unquoted `[WalletService]` array literal in logger.info.
   It also imports DomainError as a default export, but domainError.js only
   has a named export, so even after parsing it would throw at runtime.
2. src/services/order/crossDockService.js — the JSDoc opening `/**` degraded
   to a bare `**` on line 1 (`SyntaxError: Unexpected token '**'`). This
   service is mounted at /api/cross-dock via crossDockRoutes, so every
   cross-dock endpoint 500s at module load.
3. test/unit/orderLifecycleService.test.js — two incompatible test suites for
   different versions of OrderLifecycleService were pasted into one file: a
   `const mockOrderRepository` was injected inside an object literal, a
   leftover `.default` import sits inside an it() body, and references to
   `startOrder`/`completeOrder` target methods that do not exist in the
   current class. The file has never parsed, so the whole suite (incl. the
   cancelOrder transactional-outbox contract) never ran.

A source-wide `node --check` sweep shows these are the only remaining
parse-broken files (besides two already fixed in separate PRs).

scriptedbyshivam:
Closes #16530

Fixes the parser-integrity bug (see linked issue).

Source fixes
- walletService.js: quote the error message (now interpolates the offending
  address), quote the logger context string, and switch DomainError to a
  named import matching domainError.js's export (would otherwise throw a
  runtime TypeError once parsing was fixed).
- crossDockService.js: restore `/**` so the module parses; the /api/cross-dock
  routes (findHandoffCandidates, createTransferRequest, verifyHandoff, ...)
  now load instead of 500ing.

Test repairs (previously unloadable)
- orderLifecycleService.test.js: reconstructed as one coherent suite against
  the current class — 9 cancelOrder cases (outbox RPC contract, 404/403/409/
  500 guards, escrow-refund success, refund-failed 202 retryable, reconciliation
  failure) + 2 getOrderHistory cases. Removed stale cases referencing methods
  (startOrder/completeOrder) that no longer exist in the source.
- crossDockService.test.js: createTransferRequest fixtures now seed a valid
  handoff-driver profile so the newer recipient-validation runs before the
  cases under test; added coverage for the proposed-handoff-driver checks.

New regression coverage
- test/unit/walletService.test.js — address validation + getWalletDetails (5).
- test/unit/crossDockService.load.test.js — module loads and every function
  mounted by crossDockRoutes is present (2).
- test/integration/sourceTreeParses.test.js — repo-wide guard: every .js file
  under backend/api must pass `node --check`; it will fail CI the moment any
  source breaks parsing again (it caught orderLifecycleService.test.js on the
  first run). Two files (profileCache.js, WebRTCSignalingServer.js) are on a
  documented exemption list that must be cleared once their own fix PRs merge.

Verification
- All 93 tests green across the touched suites (wallet 5, crossDock routes 41,
  crossDock service 28, load 2, orderLifecycle 11, parse guard 1 + bootstrap 5).
- crossDockRoutes.test.js and crossDockService.test.js could not run on main
  at all (service did not parse) — they now pass in full.

## Summary by CodeRabbit

- **New Features**
  - Added wallet address validation for Ethereum and Polygon addresses.
  - Added wallet details retrieval, including active status and network information.
  - Invalid wallet addresses now return clear, consistent validation errors.

- **Bug Fixes**
  - Improved wallet address checking to correctly handle malformed and unsupported inputs.

- **Tests**
  - Expanded coverage for wallet validation, order lifecycle operations, transfer requests, and backend file compatibility.


scriptedbyshivam:

